PR #18, opened 2016-04-13
Add a copy of Apache 2.0 license

Calinou:
It is good practice to add a copy of the license to the repository.


PalmerAL:
Thanks!

PR #24, opened 2016-04-13
Add hosts file autocompletion

b0elter:
This proposed change would match hosts file entries in the search bar.


PalmerAL:
Looks nice! I'm wondering if this really needs a separate node module, though. Hostile adds about 400kb to the file size, which kind of large for a feature that probably won't be used very frequently (and the Min file size is really large already). Would it be possible to just use fs.readFile for this?

js/searchbar/hostsSuggestionsPlugin.js
```diff
@@ -0,0 +1,45 @@
...
+        container.appendChild(item);
+
+    })
+
+}
+
+registerSearchbarPlugin("hostsSuggestions", {
+    index: 1,
+    trigger: function (text) {
+        return !!text;
```
PalmerAL: There should probably be a minimum length to the text here, to avoid showing irrelevant entries. Once you visit a site a few times, the history search will start showing it immediately, so it probably isn't necessary to show these until a few characters have been typed.

js/searchbar/hostsSuggestionsPlugin.js
```diff
@@ -0,0 +1,45 @@
...
+
+    var results = hosts.filter(function (host) {
+        return host.indexOf(text) > -1;
+    });
+
+    results.slice(0, 4).forEach(function (result) {
+
+        var item = createSearchbarItem({
+            title: result,
+            secondaryText: 'via hosts file',
```
PalmerAL: This should probably be a noun, to match the other results (like "suggested site"). "Hosts file entry" maybe?

b0elter:
Thanks for the good feedback. I'll take a look when I get a chance.

b0elter:
@PalmerAL I've changed my approach on this and moved the hosts file parsing to `util/urlParser` so it can check arbitrary paths like *localhost:8080/path/file*. I'm not sure if that clobbers anything namespace-wise, so let me know if you'd like me to move things around. This should also provide a fix for #79. Thanks!

PalmerAL:
This looks great, thanks!

PR #30, opened 2016-04-13
Fix a few wrong names

PaulBGD:
I'm guessing you're building on Windows, where the capitalization doesn't matter. However on other systems, these files aren't found (and somehow the build succeeds without them) then their contents don't show up in the final build.


index.html
```diff
@@ -43,7 +43,7 @@
 
 	<!-- add scripts in Gruntfile.js -->
 
-	<script src="dist/build.min.js"></script>
+	<script src="dist/build.js"></script>
```
PalmerAL: Not sure this change was intentional?

PaulBGD:
It's not, good catch.

PalmerAL:
Thanks!

PR #49, opened 2016-04-17
[WIP] Add a tasks overlay

### LINKED ISSUE #5: Add tab tree structure
I would love to see support for a tree tab structure, similar to that of tree-style-tab for Firefox. Preliminarily, I could see a horizontal tab tree working as follows:
- the tab bar shows the current tab as well as all of its siblings, side-by-side
- a key combination can be used to move up and down the tree, e.g. ctrl-up & ctrl-down
- when cmd-shift-e is pressed, the current tab and its siblings stay in their positions, but an additional tab bar containing all the children of the current tab appears below. These children should update based on which tab is currently selected. Moving up in the tab tree selects the current tab's parent, and the tab bar now displays the parent and all of its siblings.

A vertical tab bar would also be useful, but is somewhat less so on normal 16:9 screens, especially considering that many sites are optimized for this aspect ratio. I'm quite busy with classes this semester, but I would love to contribute to min in some way when time allows. Maybe I'll start trying to tackle tree tabs.

I didn't want to clutter up the text of this issue, but I want to say how impressed I am with this project. I am astounded by the sheer number of commits you have made, especially considering that you are working alone. I use DuckDuckGo on all of my machines, and the tight integration between DDG and min is ridiculously awesome. This project deserves so many more stars, and I'm definitely going to tell my friends about it. Keep up the great work, and I'll definitely keep an eye on how min is progressing!

PalmerAL:
Fixes #5.


PalmerAL:
This branch has merge conflicts, and I can't figure out how to get ```rebase``` to work correctly.

PR #68, opened 2016-04-18
Hide menu bar by default on Windows and Linux. Alt toggles it.

idle-zealot:


PalmerAL:
Looks great! I didn't know electron had a preference for this.

I'm wondering, though, how many people know that alt shows the menu bar? Chrome and Firefox have a side menu, so you don't need the menu bar to access features, but in Min a lot of features are only available in the menu bar. Is this somethig a lot of apps do on those platforms?

idle-zealot:
Alt or Ctrl+M to toggle menu bar is the standard, though most applications have the menu bar shown by default. That said, other electron applications usually have autohide off by default, and allow one to toggle it in the View menu.

PalmerAL:
OK, I'll merge this and file an issue about adding an option: #74.

TheInitializer:
Doesn't seem to be working on Windows 10 x64

PR #73, opened 2016-04-18
fixed width problem for .hero

burnoutberni:
100vw does not account for eventual horizontal scrollbars, but renders to the full width, which results in an unnecessary vertical scrollbar.

This change should most likely also be made in the master branch at ```/pages/settings/settings.css``` but as this page also has overflow:hidden, I thought I'd leave it like that for now.

### DIFF
--- a/shared.css
+++ b/shared.css
@@ -195,7 +195,7 @@ ul {
 
 .hero {
 	height: 85%;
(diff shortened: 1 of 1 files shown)


PalmerAL:
Looks good, thanks!

Issue #84, opened 2016-04-20
Add support for setting Min as the default browser on Windows and Linux


PalmerAL:

emersion:
On Linux: set `MimeType` in desktop file (see https://wiki.archlinux.org/index.php/xdg-open#Set_the_default_browser).

Coming soon: https://github.com/unindented/electron-installer-debian/pull/15

Firefox one's:
```ini
Exec=firefox %u
Terminal=false
X-MultipleArgs=false
Type=Application
Icon=firefox
Categories=GNOME;GTK;Network;WebBrowser;
MimeType=text/html;text/xml;application/xhtml+xml;application/xml;application/rss+xml;application/rdf+xml;image/gif;image/jpeg;image/png;x-scheme-handler/http;x-scheme-handler/https;x-scheme-handler/ftp;x-scheme-handler/chrome;video/webm;application/x-xpinstall;
StartupNotify=true
Actions=NewWindow;NewPrivateWindow;
```

PalmerAL:
Thanks for the info! Do you know how Linux handles opening links in the default browser? It looks like Electron doesn't have an event for that, but I could be wrong.

emersion:
The `%u` in the `Exec` property is replaced by the URL.

emersion:
In the latest version, I'm able to set Min as the default browser :)

![capture du 2016-05-11 13 19 02](https://cloud.githubusercontent.com/assets/506932/15179289/11fc536a-177b-11e6-8797-98071f6a2981.png)

PalmerAL:
The second part of this (actually opening URL's) is fixed in facf72152968a709b484892b217dcbcb398be81f.

PalmerAL:
Still not fixed for Windows, though.

moigagoo:
Bumping this issue just to let you know it's impossible to set Min as the default browser on Windows.

PalmerAL:
I'm not really sure how to do this; Brave does it by manually editing a bunch of registry values: https://github.com/brave/browser-laptop/commit/f26e16020252c867adcc7c5793f4ff596bc8c154, but that looks like it would be fairly difficult to do. Does anyone know a different way?

AwesomeTech10:
I _might_ be able to add some code using ```windows-registry``` to add the values.

AwesomeTech10:
@PalmerAL What I would do is add an option in settings that looks something like this:

# Make Min your default browser
Min is currently _not_ set as your default browser
╭───────────────────╮
[...]

PR #92, opened 2016-04-21
Adds debian build categories and mime types

emersion:
These are the same as Firefox.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -139,6 +139,7 @@ module.exports = function (grunt) {
 				section: "web",
 				homepage: "https://palmeral.github.io/min/",
 				icon: "icons/icon256.png",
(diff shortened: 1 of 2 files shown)


emersion:
Fixes #92 on Linux

PalmerAL:
I think Electron needs to add an event for when a link is opened on Linux for this to work, right?

PalmerAL:
Nevermind, the comment in #84 answers this.

emersion:
`grunt-electron-installer-debian` adds `%U` after the command name, so it will spawn something like `min http://example.org`.

PalmerAL:
OK, I guess I'll go ahead and merge this. I was hoping to test it out, but Ubuntu isn't working in virtualbox, even after I go back to a snapshot that worked before.

PR #96, opened 2016-04-22
Add RPM builds, update main.build.js

shirshendu:
Require grunt-electron-installer-redhat
Create tasks for i686 and x86_64 RPM builds

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /node_modules
 .npm
 dist
+*.swp
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -156,6 +156,49 @@ module.exports = function (grunt) {
 				src: 'dist/app/Min-linux-x64',
 				dest: 'dist/app/linux'
 			}
+		},
+		'electron-installer-redhat': {
+			options: {
+				name: "min",
+				description: "Min Web Browser",
+				productName: "min",
+				productDescription: "A fast and efficient browser with built-in ad-blocking",
+				genericName: "Web Browser",
+				version: version,
+				homepage: "https://palmeral.github.io/min/",
+				icon: "icons/icon256.png",
+				categories: [
+					"Network"
+				]
+			},
+			linux32: {
+				options: {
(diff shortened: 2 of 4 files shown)


PalmerAL:
I'm having trouble getting this to build on OS X. ```rpmbuild``` requires a ```groups``` option, so adding this fixes the first problem:

```
group: "Applications/Internet"
```

but then I get this error:
```
Warning: Error executing command (1): 
rpmbuild -bb /var/folders/l4/q7gr5nt154lbfrsyngfx9bcc0000gn/T/electron-116322-67630-juii9f/min_1.1.2_i686/SPECS/min.spec --target i686
error: Unable to open temp file.
    Unable to open temp file.
 Use --force to continue.
```

I'm not sure what to do about this. I guess this is a bug in ```electron-installer-debian```.

shirshendu:
Thanks for your input. It was building and installing on Fedora 23, and I didn't think to check properly on other platforms. Let me do this:
[ ] find an OSX machine and try and get to the bottom of this (gut feeling is that its an rpmbuild configuration issue)
[ ] Try building RPMs from Fedora, RHEL, OSX and windows
[ ] install should work on Fedora as well as RHEL (start with fresh boxes so as to figure out any other library dependencies)

shirshendu:
The documentation for `electron-installer-redhat` does not mention OSX at all. Could you please let me know where you are installing rpmbuild from? (Homebrew/macports and which recipe)

PalmerAL:
Homebrew - ```brew install rpm```.

kajuberdut:
Looks like this pull request has been dead about two years and the referenced issue #219 hasn't moved in a year. Any chance somebody has working rpm build instructions using Alien or grunt-electron-installer-redhat?

PalmerAL:
I tried this again with a newer version of `electron-installer-redhat`, and everything seems to be working now, so I've added an RPM build script in 31384f19f8966aa23b2c4f89a3d1a4ce2d2452fa. A prebuilt RPM will be available starting with the next release.

PR #102, opened 2016-04-23
Reload only webview of current tab

### LINKED ISSUE #50: "Reload" reloads all tabs, not just the current tab
This is true for both command-R and the pull-down menu.

b0elter:
This would resolve #50.


PalmerAL:
Thanks, but I believe this was already fixed yesterday in a50ffdd8fc55b2d812421b31d75eaa7d4739d906.

b0elter:
Roger. Wasn't up to date!

Issue #86, opened 2016-04-20
Add icon for Windows


TheInitializer:
Currently it just has the default Electron icon.

chrissbendel:
Fixed, tested, and working on Windows. #107

DylanC:
+1 Also have this issue.

dbfreem:
@PalmerAL I was able to see the min icon on windows. Can this be closed?

PalmerAL:
Yup, this was fixed by #107 (thanks for going through these issues BTW!)

PR #107, opened 2016-04-25
windows icon

chrissbendel:
Added min icon instead of default electron icon on Windows (Issue #86)


PalmerAL:
Thanks! Can you please make these changes to the main.js file, instead of main.build.js? Main.build.js is automatically generated by Grunt, so changes to it will be overwritten every time Min is rebuilt.

PR #108, opened 2016-04-26
Windows Icon Fix

chrissbendel:
Windows Icon fixed in main.js as per request


chrissbendel:
Final fix for #86 ? Let me know if there's anything else you need @PalmerAL

PalmerAL:
Looks good!

PandawanFr:
Doesn't work for me.....
The small icon next to the title does change, but not the one in the menu bar or in File Explorer...

PR #118, opened 2016-05-03
Close tabs with mouse middle click

### LINKED ISSUE #40: Close tab on middle click
Just about every browser I've used recently has supported closing a tab by middle-clicking on it in the tab bar; could this behavior be added to Min?

burnoutberni:
Fixes #40 

@PalmerAL Did ```tabId``` exist on purpose and is there a difference to ```data.id```?


PalmerAL:
(Sorry for the delay in reviewing this; I was really busy yesterday and didn't have time to work on Min)

```tabId``` is the same as ```data.id```. I think the reason for ```tabId``` was that using that would allow the tab object to be garbage collected earlier, but it isn't near large enough of an object to make a difference in memory usage. (Also, ```tabId``` isn't actually used anywhere).

PR #121, opened 2016-05-04
Fix OS X bug when all tabs are closed

valtlfelipe:
When all tabs are closed in OS X and min is not quitted (by hitting the X button). If you click on the application dock icon, nothing happens.

With this fix, min opens a new tab if you click on the application dock icon in OS X, by listening to the electron `activate` event.

### DIFF
--- a/main.build.js
+++ b/main.build.js
@@ -119,6 +119,17 @@ app.on("open-url", function (e, url) {
 	}
 });
 
+/**
+ * Emitted when the application is activated, which usually happens when clicks on the applications's dock icon
+ * https://github.com/electron/electron/blob/master/docs/api/app.md#event-activate-os-x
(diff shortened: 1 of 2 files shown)


PalmerAL:
I'm not sure calling ```sendIPCToWindow``` is actually necessary. Would something like this work?

```
if (!mainWindow) {
    createWindow();
}
```

valtlfelipe:
I just followed the `open-url` action. 
Worked as expected 😄! Just committed the change

PalmerAL:
Can you please squash this into one commit? Otherwise, looks good!

valtlfelipe:
Done! Thanks!

PalmerAL:
Looks good!

PR #154, opened 2016-05-14
Add basic KeyMap customization support

plrenaudin:
This PR is adding a basic KeyMap customization support.
Changed the "Privacy Settings" to "Settings" and added the KeyMap Settings section
The key combinations are stored with the user settings.

This is a great project, thank you @PalmerAL !

### DIFF
--- a/pages/settings/settings.js
+++ b/pages/settings/settings.js
@@ -6,6 +6,55 @@ var contentTypes = {
 	//humanReadableName: contentType
 	"scripts": "script",
 	"images": "image",
+};
+
+settings.get('keyMap', function (keyMapSettings) {
+	keyMap = userKeyMap(keyMapSettings);
+
+	Object.keys(keyMap).forEach(createKeyMapInput);
+});
+
+function createKeyMapInput(action) {
+	var keyMapList = document.getElementById('key-map-list');
+	var key = keyMap[action];
+	var li = document.createElement('li');
+	var label = document.createElement('label');
+	var input = document.createElement('input');
+	label.innerText = formatCamelCase(action);
+	label.htmlFor = action;
+
+	input.type = "text";
+	input.id = input.name = action;
+	input.value = key;
+	input.addEventListener('change', onKeyMapChange);
+
+	li.appendChild(label);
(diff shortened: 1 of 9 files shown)


PalmerAL:
This looks nice!

settings.js isn't really meant to hold default settings - the ```list``` item is used as a cache so the database doesn't have to be queried. This causes a race condition where the keymap will sometimes fail to load, since the value in the database will be used if the settings haven't loaded yet.

I think a better solution would be to create a file in ```/util/defaultKeyMap.js```, which does something like this: 

```
var defaultKeyMap = {
			"addPrivateTab": "shift+mod+p",
```

Then, in ```keybindings.js```, the code could look like this:

```
settings.get('keyMap', function (keyMap) {
  if(!keyMap) {
    keyMap = defaultKeyMap;
  }
```

PalmerAL:
The heading in the settings page should probably have the same style as the "content blocking" heading. Also, I think some changes to the spacing would be good, as well as adding a subtitle underneath the heading:

screen shot 2016-05-14

pages/settings/settings.js
```diff
@@ -6,6 +6,49 @@ var contentTypes = {
...
+	Object.keys(keyMap).forEach(createKeyMapInput);
+});
+
+function createKeyMapInput(action) {
+	var keyMapList = document.getElementById('key-map-list');
+	var key = keyMap[action];
+	var li = document.createElement('li');
+	var label = document.createElement('label');
+	var input = document.createElement('input');
+	label.innerText =	label.htmlFor = action;
```
PalmerAL: Would it be possible to have the formatted name of the action ("Open Private Tab" instead of "openPrivateTab") here?

plrenaudin:
Thanks, it makes more sense this way.

plrenaudin:
@PalmerAL Let me know if you want something else for this to be merged.

PalmerAL:
Looks good to me!

Can you please squash this into one commit? After that, I think this is ready to be merged.

plrenaudin:
Done!

PalmerAL:
Thanks!

PR #158, opened 2016-05-15
Add save option on image right click

plrenaudin:
This allows the user to download the image.

**Warning**: it does not keep the image **name** and **extension**.
Any idea to improve this is welcome.

### DIFF
--- a/js/webviewMenu.js
+++ b/js/webviewMenu.js
@@ -112,6 +112,12 @@ var webviewMenu = {
 					navigate(webviewMenu.cache.tab, IPCdata.image);
 				}
 			}));
+			menu.append(new MenuItem({
+				label: 'Save image',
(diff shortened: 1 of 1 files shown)


PalmerAL:
Interesting, the image name and extension are saved for me (on OS X).

I'm guessing this menu item won't be used very often, so it probably makes more sense to get the web contents each time the menu item is clicked rather than on startup. That way, it won't take up memory and increase the startup time.

Other than that, looks good!

plrenaudin:
I tested with a tricky situation where image filename is followed by a query string.
It may work correctly on some occasion.

PalmerAL:
Can you please squash this into one commit?

PalmerAL:
Looks good!

PR #164, opened 2016-05-18
Added pinch-zoom support for Mac

ChristiaanLock:
#### Added pinch-zoom support for Mac
On pinch, the browser-engine sends a ctrl keypress & a mousewheel event. However, platforms other than Mac use ctrl+mouswheel to zoom. Pinch-zooming should be smoother than normal zooming, so the threshold is lowered to 10. Non-mac devices are not affected and cmd+mousewheel is still working.

### DIFF
--- a/js/webview/swipeEvents.js
+++ b/js/webview/swipeEvents.js
@@ -9,10 +9,10 @@ window.addEventListener("mousewheel", function (e) {
 
 	verticalMouseMove += e.deltaY;
 	eventsCaptured++;
-	platformZoomKey = e.ctrlKey;
+	var platformZoomKey = e.ctrlKey;
 
 	if (navigator.platform == "MacIntel") {
-		if(e.ctrlKey) {
+		if(e.ctrlKey && !e.defaultPrevented) {
 			if( verticalMouseMove > 10 ) {
 				return zoomOut();
 			}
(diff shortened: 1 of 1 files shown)


js/webview/swipeEvents.js
```diff
@@ -9,15 +9,27 @@ window.addEventListener("mousewheel", function (e) {
 
 	verticalMouseMove += e.deltaY;
 	eventsCaptured++;
+	platformZoomKey = e.ctrlKey;
```
PalmerAL: This should probably be ```var platformZoomKey```, since it isn't defined globally.

js/webview/swipeEvents.js
```diff
@@ -9,15 +9,27 @@ window.addEventListener("mousewheel", function (e) {
...
 
+	if (navigator.platform == "MacIntel") {
+		if(e.ctrlKey) {
```
PalmerAL: This should also check that ```e.defaultPrevented``` is ```false```. Otherwise, things like the pdfjs viewer break, since they have their own pinch-to-zoom code.

PalmerAL:
Thanks for the PR! Looks good once comments are addressed.

js/webview/swipeEvents.js
ChristiaanLock: Done!

js/webview/swipeEvents.js
ChristiaanLock: Done!

PalmerAL:
Can you squash this into one commit?

ChristiaanLock:
Sure, I'll close this PR and create a new one

PR #171, opened 2016-05-20
Added pinch-zoom support for Mac

ChristiaanLock:
On pinch, the browser-engine sends a ctrl keypress & a mousewheel event. However, platforms other than Mac use ctrl+mouswheel to zoom. Pinch-zooming should be smoother than normal zooming, so the threshold is lowered to 10. Non-mac devices are not affected and cmd+mousewheel is still working.


PalmerAL:
Looks good, thanks!

PR #172, opened 2016-05-20
Added Save Link As to context menu

ChristiaanLock:
Download files & html pages on links from the context menu. Not enables on images because 'Save image' is already implemented


PalmerAL:
Thanks!

(You can actually squash commits without creating a new PR - http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html has instructions on how to do this).

PR #173, opened 2016-05-20
Remove wrong keyMap setting

plrenaudin:
"Add Task" keybinding is part of the application menu and cannot be customised for now


PalmerAL:
Thanks!

PR #178, opened 2016-05-25
Changed code style to Standard

wb9688:
I've changed the code style to [Standard](https://github.com/feross/standard), btw `npm test` doesn't pass now because some variables aren't defined in the same file and idk how to let Standard ignore that...

PS: when I try to run it (with and without this commit) I get the following error:
Screenshot of error

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Min
 
+[![js-standard-style](https://cdn.rawgit.com/feross/standard/master/badge.svg)](https://github.com/feross/standard)
+
 Min is a smarter, faster web browser. It includes features such as:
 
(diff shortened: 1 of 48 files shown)


PalmerAL:
Looks good!

> npm test doesn't pass now because some variables aren't defined in the same file and idk how to let Standard ignore that

I don't think there is a way... standard expects ```require``` to be used for everything, but Min doesn't do that. <a href="https://github.com/feross/standard#i-use-a-library-that-pollutes-the-global-namespace-how-do-i-prevent-variable-is-not-defined-errors">You can whitelist specific variables, but each one has to be listed individually. </a> For now I guess we can just ignore it.

> PS: when I try to run it (with and without this commit) I get the following error:

Can you file a separate issue about that?

PR #202, opened 2016-06-16
Add support for custom !bang commands

PalmerAL:


PalmerAL:
Git won't rebase this for some reason, so I guess I'll merge it without squashing.

PR #220, opened 2016-07-27
Fix inspector shortcut

City-busz:
This makes the shortcut consistent with other browsers.


PalmerAL:
On OS X the shortcut is option+command+i in Safari/Chrome/Firefox (and in Chrome, shift+command+I is "email page location"), so this should probably check the platform and use the correct shortcut based on that.

PalmerAL:
Looks good, thanks!

PR #221, opened 2016-07-27
Add option to switch full screen

City-busz:
This allows users to toggle full screen with F11.

### DIFF
--- a/main.build.js
+++ b/main.build.js
@@ -335,6 +335,16 @@ function createAppMenu () {
           type: 'separator'
         },
         {
+          label: 'Full Screen',
+          accelerator: (function() {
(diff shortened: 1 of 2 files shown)


PalmerAL:
Thanks for the PR!

On OS X and Ubuntu, the full screen item usually goes under the View menu, so it probably makes more sense to put it there.

Also, instead of the ```click``` event handler, you can use <a href="http://electron.atom.io/docs/api/menu-item/#menuitem">```role: 'fullscreen'```</a>, which should have the same effect.

City-busz:
Okay, I updated my commit.

PalmerAL:
Thanks!

PR #247, opened 2016-09-04
Dark Mode?

PandawanFr:
Hey,

So I tried using the basic Dark Mode I showed you and add it to the browser, I think most (if not all) have it, the only problem is that I don't know how you made the preferences system work (never used Electron before), so I just added a checkbox in the preferences for you to link it with the stylesheets...

Basically you just have to add/remove the css/darkMode.css from the head.

Hope that helps!


pages/error/index.html
```diff
@@ -3,11 +3,13 @@
 
 <head>
 	<link rel="stylesheet" href="../pagebase.css" />
+	<link rel="stylesheet" href="/error.css" />
+	<link rel="stylesheet" href="../../css/darkMode.css"/> <!-- This is how you add Dark Mode! Please do it -->
 	<link rel="stylesheet" href="../../ext/font-awesome-4.4.0/css/font-awesome.min.css" />
 	<script defer src="error.js"></script>
 </head>
 
-<body style="background-color: #eee">
```
PandawanFr: Removed Background-color from Error and Crash to put them in separate files because they were conflicting with the darkMode.css

pages/settings/settings.js
```diff
@@ -135,6 +136,12 @@ trackerCheckbox.addEventListener("change", function (e) {
 	});
 });
 
+// Dark Mode Settings
```
PandawanFr: I tried adding the JS/Preferences part, but failed...... Dunno if u need it

reader/reader.js
```diff
@@ -28,7 +28,7 @@ function startReaderView(article) {
 
 	document.body.removeChild(parserframe);
 
-	var readerContent = "<link rel='stylesheet' href='readerView.css'>";
+	var readerContent = "<link rel='stylesheet' href='readerView.css'>\n<link rel='stylesheet' href='../css/darkMode.css'>";
```
PandawanFr: This is to add Dark Mode to the Reader

PalmerAL:
This is a good start!

Rather than having a darkMode.css file that has all the CSS for the dark mode, I think it would be better to add a ```dark-mode``` class to the page when the dark mode is enabled. Then, each individual page (settings, the error page) could have separate styling for dark mode in that page's stylesheet.

The settings part you added works, and it will set the setting correctly. To read value of the setting, you can do something like this;

```
settings.get('darkMode', function(value) {
    //"value" will be equal to undefined if the user has never changed the setting, or the value if it has been changed
})
```

So to complete this, you would add something like this to each of the pages where the dark mode should be enabled:

```
settings.get('darkMode', function(value) {
    if (value === true) {
        document.body.classList.add('dark-mode')
    }
})
```

Since that code would be shared between all of the pages, it should probably go in a separate file, ```js/util/theme.js```.

Also, you need to include the JS for the database, the library used to access the database, and the settings API itself in each page where you access a setting. They are linked like this from the settings page:

```
	<script src="../../ext/Dexie.min.js"></script>
	<script src="../../js/util/database.js"></script>
	<script src="settings.js"></script>
```

but you may have to change the paths of those scripts depending on the location of the page.

Also, the reader view already has a dark theme, that turns on based on the time of day. The code for that is <a href="https://github.com/minbrowser/min/blob/master/reader/reader.js#L5"> in reader/reader.js on line 5</a>. The only change you need to make there is to also enable the dark theme when the dark theme setting is true. The best way to do that would probably be to put the ```if``` statement linked above inside the settings callback:

```
settings.get('darkTheme', function(value) {
    if (value === true || (hours > 21 || hours < 6) {
        ...
```

Don't hesitate to ask if you have any other questions!

PandawanFr:
Hey ok, followed your advice, and it definitely is better!

Now 2 problems:  
First, I don't know why, but the setting isn't saved... Meaning that if you click on the checkbox (in preferences) and restart/reload the page, it goes back to empty/false...

Second one, the Reader.js doesn't allow me to add css/js to the iframe, I think it has to do with the sandbox but anyways, this is a problem because the text is black on somewhat black.........

PalmerAL:
The cause of the first problem is that you have to get the value of the setting when the page loads, and then set the value of the checkbox based on that:

```
settings.get('darkMode', function (value) {
    darkModeCheckbox.checked = value
})
```

For the second problem, you can modify the iframe using ```rframe.contentDocument```; that will give you a reference to the document inside the iframe that you can modify. However, you should just be able to modify the existing code. In ```theme.js```, if the dark theme is enabled, you can set ```window.isDarkMode``` to be ```true```. Then, you can change <a href="https://github.com/minbrowser/min/blob/master/reader/reader.js#L57">this code</a> to add the ```dark-mode``` class to the iframe. Once you rename <a href="https://github.com/minbrowser/min/blob/master/reader/readerView.css#L153">the style in reader.css, line 153</a> to "body.dark-mode", the dark mode in the reader view should work.

After that, you should be able to delete <a href="https://github.com/minbrowser/min/blob/master/reader/reader.js#L1-L9">this code</a>, since it is being replaced by what is in theme.js.

pages/settings/index.html
```diff
@@ -36,6 +36,20 @@
 	<div class="spacer">
 	</div>
 
+	<div class="spacer"></div>
+	<div class="container padding" id="appearance-settings-container">
+		<h3>Appearance</h3>
+
+		<div class="setting-section">
+			<input type="checkbox" id="checkbox-dark-mode" />
+			<label for="checkbox-dark-mode">Enable dark mode.</label>
```
PalmerAL: Can you remove the period from the end of this description, to match the other settings?

pages/settings/settings.js
```diff
@@ -135,6 +136,12 @@ trackerCheckbox.addEventListener("change", function (e) {
...
+darkModeCheckbox.addEventListener("change", function (e) {
+	settings.set("darkMode", this.value)
```
PalmerAL: The ```value``` of a checkbox is always "on", whether the checkbox is checked or not, so this should be ```this.checked``` instead.

PandawanFr:
Wow! Thanks a lot! Now it works completely!
U are so nice! Instead of just telling me not to worry about it, and that you would do it, you taught me how to do it! 

**Thank you!**

PandawanFr:
Lemme see if I can also make the main index.html page (browser itself) also have the dark mode...

PalmerAL:
Thanks, I'm glad that was helpful.

PalmerAL:
Also, you should be able to simplify the code for the task overlay dark mode. In ```taskOverlay.css```,  you can write something like this:

```
body.dark-mode #task-overlay {
    background-color: ...
```

That will select an element with the ID "task-overlay" that is inside a ```<body>``` tag with the ```dark-mode``` class. Then, you don't need the extra code in theme.js and base.css, and can remove the ID from the ```<body>``` element.

The browser page already includes Dexie.js, database.js, and settings.js, so you don't need to add those. Also, all of the scripts in that page are combined into one using Grunt. To add the ```theme.js``` script to that page, you would add it on a new line after <a href="https://github.com/minbrowser/min/blob/master/Gruntfile.js#L50">this line</a>. Then, you would rebuild the ```build.min.js``` file. Open the terminal, navigate to the ```min``` folder (By typing ```cd ~/<path to min>```. Then, type ```grunt``` to rebuild the script.

PandawanFr:
For the first part, I am currently changing some parts of the code, but changing it to this 
`body.dark-mode #task-overlay`  
doesn't actually work... The Add Task button doesn't become dark anymore, so I changed it a bit. 

Anyways, most of your fixes worked so yeah!

index.html
```diff
@@ -17,14 +17,14 @@
 	<link rel="stylesheet" href="node_modules/dragula/dist/dragula.min.css" />
 </head>
 
-<body>
+<body id="browser-view">
```
PalmerAL: Can you remove this change, and the other spacing changes to this file?

reader/readerView.css
```diff
@@ -193,3 +196,16 @@ dd {
 .off-screen {
 	display: none;
 }
+
```
PalmerAL: This CSS (lines 199-211) is duplicated and can be removed.

css/base.css
```diff
@@ -14,3 +14,12 @@ html {
 	overflow: hidden;
 	-webkit-user-select: none;
 }
+
```
PalmerAL: This code doesn't seem to be necessary, and it also affects the background color of ```<webviews>``` on pages that don't define a background color (which is a problem, since the default background color is assumed to be white). It's probably best to remove this code, and move the ```color: lightgrey``` part to ```taskOverlay.css```.

PalmerAL:
I added a few comments on this. After those are fixed, I think this is ready to be merged!

PandawanFr:
Ok Done!

PandawanFr:
Ok, now we good!

PalmerAL:
Looks great!

PandawanFr:
Yay! Thanks!

PalmerAL:
This has been released in 1.5.0.

PR #252, opened 2016-09-17
Use app.getPath('userData') for configuration files

tensor5:
Same as path.join(app.getPath('appData'), packageJson.productName) by
default.

### DIFF
--- a/main.build.js
+++ b/main.build.js
@@ -1,9 +1,10 @@
 const electron = require('electron')
 const fs = require('fs')
+const path = require('path')
 const app = electron.app // Module to control application life.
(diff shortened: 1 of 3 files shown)


PalmerAL:
On OS X, this is saving the window coordinates to ```~/Libary/Application Support/minwindowBounds.json```, so I'm not sure Electron is calculating this path correctly.

tensor5:
I forgot a `/` at the end of the path. It should save to `~/Libary/Application Support/Min/windowBounds.json`.

PalmerAL:
Looks good!

PR #298, opened 2017-01-12
add metager

freddii:


PalmerAL:
I'm not sure it makes sense to add Metager to the search engine list, since it doesn't seem to have many users. An "other" option, with the ability to enter a custom search engine, would probably be better.

freddii:
yes metager is not yet one of the big players. 
But it is the only searchengine i know that has published its sourcecode.
https://gitlab.metager3.de/open-source/MetaGer/
Its also run by a non-profit association.

I think as long as min browser is not payed by one of the big players for a top place for its search engine it would be nice to give also smaller searchengines a chance.

of course adding an "other" option with the ability to enter a custom search engine would work too.

PalmerAL:
I'm definitely not opposed to providing a choice of other search engines. However, there's a lot of smaller search engines, and adding them all would result in a long list of options that makes it difficult to choose the search engine you want. I think it makes more sense to have the most popular search engines in a list, and an "other" option to choose a less popular search engine like Metager.

I'm going to close this for now, and I've opened an issue about changing the settings: #303.

PalmerAL:
Support for custom search engines has been added in c90549c270a22a2ca2537935c9756f3a56327c79.

PR #317, opened 2017-02-01
make Standard 😃, bookmarksHistory

Or3stis:
Linter errors have been fixed in `js/bookmarksHistory` 🎉.

The only complaining file is the `placesWorker.js` with some `(no-unused-vars)`. Those warnings will be fixed once we implement the modular approach.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .npm
 dist
 main.build.js
+.DS_Store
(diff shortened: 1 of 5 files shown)


PalmerAL:
Looks good!

My only question is about the "global" statements at the top. Some of those (like Worker) should definitely be included, but others (spacesRegex, searchPlaces...) will probably unnecessary once these files are switched to using modules. Maybe just include the browser-provided globals (```Worker```, ```performance```, ```importScripts```) and leave the rest out for now?

Or3stis:
I would prefer to leave them as `global` statements for now as a mental reminder of which scripts need to change into modules. We can remove the unnecessary ones once we switch to modules.

PalmerAL:
OK, sounds fine!

PR #318, opened 2017-02-02
Fixes #311

### LINKED ISSUE #311: Dragging the window while the mouse is over a tab causes the tab to be clicked

abejfehr:
Replaces each tab element's `click` event listener with separate `mousedown`, `mousemove` and `mouseup` event listeners which determine whether or not the user dragged the mouse during the click.

If a drag was detected on the tab element, a flag is toggled and the action for the click won't happen.

This is the same solution mentioned in [this](http://stackoverflow.com/questions/6042202/how-to-distinguish-mouse-click-and-drag) StackOverflow post.

### DIFF
--- a/js/navbar/navbarTabs.js
+++ b/js/navbar/navbarTabs.js
@@ -238,13 +238,22 @@ function createTabElement (data) {
   })
 
   // click to enter edit mode or switch to a tab
-  tabEl.addEventListener('click', function (e) {
-    if (e.which === 2) { // if mouse middle click -> close tab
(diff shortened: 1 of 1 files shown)


PalmerAL:
Thanks for doing this! I've added a fix based on this PR in 757e2a45d2b8fbca032652fc1d1c65cbc41120b1, which also fixes some additional bugs with window dragging (such as dragging the window while hovering over a tab close button, or dragging while the task overlay is open).

Issue #169, opened 2016-05-19
Open multiple windows / instances?


timkinnane:
I think min is a great tool for running focused web apps and dev tools. To that end, I'd like to have multiple instances open, so I can test apps with different concurrent browser sessions side-by-side, without having to tab through to see updates in each.

Is this possible, planned or not possible by design?

CalObbur:
I also would like multiple instances to work, although maybe I should open a separate bug/issue for this?

With one Min instance, everything works fine. If I open another min instance on another desktop, autocomplete is broken, my settings to minimize the menu are reverted, and the tabs from my other instance are brought up instead of the blank start page. So technically I guess multiple instances do work aside from those bugs above.

PalmerAL:
@CalObbur Multiple instances are probably never going to work, because two instances can't open the same IndexedDB database at the same time, which is why settings and autocomplete don't work. Multiple windows per instance is definitely doable however, and would solve most of the reasons that having multiple instances is necessary.

ibash:
@PalmerAL was there a reason that multiple windows wasn't added? If not I could help out here, too. I'd live to see an awesome split view though (like #349).

PalmerAL:
Yeah, I agree multiple windows would be a good idea. It's going to be fairly difficult to implement this, because a bunch of stuff (particularly the task system and session restore) was designed with the idea that there's only one open window, but I think it's definitely doable if you're interested in working on it.

will-ca:
Out of all the features I've habitually tried to use so far, `CTRL+N`/`File > New Window` is the only one that hasn't worked. If it were possible to view multiple pages at the same time, then I'd frankly have a hard time thinking of any real functionality where Min is lacking when compared to the big browsers (Chrome, Firefox, etc).

Kekker-git:
Multiple windows is the one thing min is missing for me. With other browsers, I'm often reading documentation or wiki pages in one window while I have a video playing in a window on my other monitor. For now, I have to keep a second browser installed to manage this.

dancojocaru2000:
Are multiple windows considered/worked on?

PalmerAL:
No ETA for now.

DrSensor:
In Linux, does the min browser always run on single instance mode?
I mean when I spawn multiple min browser it will share the same instance (similar on how `firefox -new-window` or `kitty --single-instance` works)

---
<sup>from what I know electron app in macOS always run in single instance mode</sup>

PalmerAL:
[...]

PR #338, opened 2017-04-10
Added macOS dock menu

israelroldan:
I have added a dock menu for macOS. This should also be useful for whenever #169 is implemented as it is standard practice to open new windows from the dock menu.

Preview:
screen shot 2017-04-10 at 12 17 11

### DIFF
--- a/main/main.js
+++ b/main/main.js
@@ -176,6 +176,7 @@ app.on('ready', function () {
   // mainWindow.openDevTools()
 
   createAppMenu()
+  createDockMenu()
 })
 
 app.on('open-url', function (e, url) {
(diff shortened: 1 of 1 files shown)


main/main.js
```diff
@@ -539,3 +540,36 @@ function createAppMenu () {
...
+      },
+      {
+        label: 'New Private Tab',
+        click: function (item, window) {
+          sendIPCToWindow(window, 'addPrivateTab')
+        }
+      },
+      {
+        label: 'New Task',
+        click: function (item, window) {
```
z3t0: What is a task.

israelroldan:
Min calls groups of tabs "Tasks".

z3t0:
@israelroldan Ah thanks that makes sense

PalmerAL:
Thanks for doing this! I've merged this with a few minor style fixes in #344.

PR #352, opened 2017-05-06
Added close buttons for tabs in taskOverlay

code-hunger:
After using this browser for a while and enjoying the taskOverlay I found I missed this feature.
I implemented it and I hope it'll be useful for others, too.


PalmerAL:
The change to the key map looks good, I've merged that into master in c670d00e859c744a524a5aaf92fac3bbf675022e.

code-hunger:
I'm just doing some refactoring of the `createTaskContainer` (quite a big one, huh) to split it into manageable parts. I'll push it in a few minutes.

PalmerAL:
The refactoring is really good! A few things:

* It probably makes sense to combine ```TaskOverlayBuilder.create.tabElement``` with ```getTaskOverlayTabElement```, or rename ```getTaskOverlayTabElement``` to make it more clear what the separation between the two is.
* I think it would be simpler to have ```addTabCloseButton``` be a function that takes a tab and a task, and returns a button. Then, ```TaskOverlayBuilder.create.tabElement``` could do ```el.appendChild(getTabCloseButton(tab, task))```.
* ```closeTab``` doesn't work very well from ther task overlay (for example, closing the last tab in a task will close the overlay, which shouldn't happen). There's already code to handle this on <a href="https://github.com/code-hunger/min/blob/a2e92cebfaaf6a167665cec517fbe9010b324b90/js/taskOverlay.js#L33">line 33</a> that closes the tab when you swipe left on the tab element; this code could probably be extracted into a function that takes a task and a tab and closes the tab.

js/taskOverlay.js
```diff
@@ -54,62 +54,76 @@ function getTaskOverlayTabElement (tab, task) {
...
 
-  container.setAttribute('data-task', task.id)
+  closeTabButton.addEventListener('click', function (e) {
+    closeTab(taskTabElement.getAttribute('data-tab'))
+    tabContainer.removeChild(taskTabElement)
 
-  var taskActionContainer = document.createElement('div')
-  taskActionContainer.className = 'task-action-container'
+    // do not close taskOverlay
+    e.preventDefault()
```
PalmerAL: ```preventDefault``` doesn't do the correct thing here - it will stop the browser from doing anything based on this event, but it won't stop other listeners from running. ```stopImmediatePropogation``` should work instead.

code-hunger: Thanks for the clarification. I'll fix it in the next commit.

code-hunger:
* About the first point - a good separation would look like this:
  1. The first function builds the element
  2. The second adds the attributes, the event listeners and the close button.

However, it is hard to find a good name for the first function. What it essentially does is wrap a 20-line call to `createSearchbarItem` (if we move the setAttribute calls to the second function) so the name needs to be something more specific. And I get a name like `createSearchbarItemForTaskTabElement` which obviously is not a solution.

* About `addTabCloseButton` - what this function actually uses are the current parameter. If we used task and tab instead, we'd have to query for the tabContainer and the taskTabElement inside the function.
If there were an easy way to convert a tab id to a taskoverlay tab element - ok, but now for this we'd need two additional querySelector operations inside `addTabCloseButton`. It would become overcomplicated (I suppose?) so for now I think the two parameters are an acceptable solution.
I understand your concerns about this but I suggest that we keep it simpler untill a better approach is found.

* I do not really understand the relation between the `closeTab` function and the delete parameter in `createSearchBarItem`. The first is a function to close a tab, and the second is some callback for the element shown in the dropdown searchbar suggestions.
Now, looking at the source of `closeTab` I think it is not the right function to call here since it both *"closes a tab and switches to the next/creates a new one"*. We need only the first part - to *close* a tab - so I suppose `destroyTab` will solve the problems you talk about. 
Am I right? I.e. do this here:
```
  closeTabButton.addEventListener('click', function (e) {
    destroyTab(tabId)

```

PalmerAL:
What's happening is that the task overlay shares ```createSearchbarItem``` with the searchbar - each tab element shown in the task overlay is actually just a searchbar list item. So ```createSearchbarItem``` creates a ```<div class="searchbar-item"></div>```, which gets added to the task overlay. One of the built-in behaviors of the searchbar items that ```createSearchbarItem``` creates is that you can swipe left on the item using a trackpad to delete it (the code for that is <a href="https://github.com/minbrowser/min/blob/master/js/searchbar/searchbar.js#L224">in searchbar.js, line 224</a>). This also works in the task overlay - you can swipe left to delete tabs. When you swipe left, the function given to the ```delete``` parameter - which is what is on <a href="https://github.com/code-hunger/min/blob/8f7ae5d482239ea3a5fac4410926c46034f591d1/js/taskOverlay.js#L32">line 32</a> - will be called.

This code handles a few special cases that need to be considered:

* If the last tab in the task is removed, the task needs to be removed.
  * If that task is the only task open, the overlay needs to be closed and a new task needs to be opened.

Because this code already handles both of these cases, I think we can re-use it for the close buttons. Then, you could do something like this:

```
function closeTabFromOverlay(tab, task) {
      task.tabs.destroy(tab.id)
      destroyWebview(tab.id)

      // if there are no tabs left, remove the task

      if (task.tabs.count() === 0) {
        destroyTask(task.id)
        if (tasks.get().length === 0) {
          addTaskFromOverlay()
        } else {
          // re-render the overlay to remove the task element
          getTaskContainer(task.id).remove()
        }
      }
    }
```

Combined with the changes in code-hunger@ead379e1ac4b4d0593ca9930bfc3233111b557af, the code in createTaskOverlayTabElement would just become:
```
function createTaskOverlayTabElement (tab, task) {
  return createSearchbarItem({
    title: tab.title || 'New Tab',
    secondaryText: urlParser.removeProtocol(tab.url),
    classList: ['task-tab-item'],
    delete: function () {
      closeTabFromOverlay(tab, task);
    }
  })
}
```

This is small enough it could probably just be inlined into TaskOverlayBuilder.create.tabElement, solving the naming issue completely (although I'm fine with leaving them separate if you think that would be better).

For the third issue, addTabCloseButton could just use the ```closeTabFromOverlay``` function, since it would handle all of these cases.

For the second issue, I do see your point there. I guess it's fine to leave it that way for now (although you don't need ```tabContainer``` - you can just do ```taskTabElement.parentNode.removeChild(taskTabElement)```.

code-hunger:
Thanks for the idea! I've been actually pretty close to it but now the code's cleaner. I did also move the taskBuilder to a separate file (see the comit comment).

However I had to do some checking, namely here:
```
          if(tabId !== current_selected.tabs.getSelected() ) {
            removeTaskFromOverlay(tasks.get(taskId))
```
If I omit the check and close the currently active tab, then exitting the overlay gives exceptions due to the missing currently active tab.
I think a good solution would be to select the tab next to the closed one, and if the task is about to be deleted - instead of that, open a new blank tab in it and make it active.

PalmerAL:
This looks pretty good!

Fixing the closing behavior can probably wait, and this PR is getting pretty big already, so I'd like to go ahead and merge it.

For the actual buttons, it would be nice if they looked more like the buttons in the tabstrip. Maybe something like this:

```
        var closeTabButton = document.createElement('button')
        closeTabButton.innerHTML = '<i class="fa fa-close"/>'
        closeTabButton.className = 'closeTab'
```

```
.task-tab-item button.closeTab {
    position: absolute;
    width: 2em;
    height: 2em;
    left: -2.1em;
    top: calc(0.16em + 1px); /* searchbar styles padding-top */
    bottom:0;
    border: none;
    border-radius: 50%;
    outline: none;
    cursor: pointer;
    opacity: 0;
    background: transparent;
}
.task-tab-item:hover button.closeTab {
    opacity: 0.5;
}
.task-tab-item button.closeTab:hover {
    opacity: 1;
    background: rgba(0, 0, 0, 0.05);
}
.task-tab-item button.closeTab:active {
    background: rgba(0, 0, 0, 0.175);
}
.task-tab-item button.closeTab > i {
    margin-left: -1px;
    margin-right: 0;
    padding: 0;
    opacity: 1;
    vertical-align: baseline;
    font-size: 1em;
}
```

But I don't know. What do you think?

code-hunger:
Just tried your css replacing my changes. Currently using dark mode (forced automatically by time).
Can't tell how it looks like in light mode but for me the close button is black over a very dark background. I would personally prefer a more visible button but that again is really subjective.
In fact I'd really prefer no close buttons at all and only keyboard control but it's not the time for that now :)
I'd suggest looking at how such buttons are made in other UIs, or leaving the choice to the community, but I personally don't really care whether it has 0.5 or 0.6 opacity.

PalmerAL:
Dark mode would be fixed with this CSS:

```
.task-tab-item button.closeTab {
    position: absolute;
    width: 2em;
    height: 2em;
    left: -2.1em;
    top: calc(0.16em + 1px); /* searchbar styles padding-top */
    bottom:0;
    border: none;
    border-radius: 50%;
    outline: none;
    cursor: pointer;
    opacity: 0;
    background: transparent;
	transition: 0.05s;
	color: currentColor;
}
.task-tab-item:hover button.closeTab {
    opacity: 0.5;
}
.task-tab-item button.closeTab:hover {
    opacity: 1;
    background: rgba(0, 0, 0, 0.05);
}
.dark-mode .task-tab-item button.closeTab:hover {
    opacity: 1;
    background: rgba(0, 0, 0, 0.15);
}
.task-tab-item button.closeTab:active {
    background: rgba(0, 0, 0, 0.175);
}
.dark-mode .task-tab-item button.closeTab:active {
    background: rgba(0, 0, 0, 0.35);
}
.task-tab-item button.closeTab > i {
    margin-left: -1px;
    margin-right: 0;
    padding: 0;
    opacity: 1;
    vertical-align: baseline;
    font-size: 1em;
}
```

Firefox, Safari, and Chrome all seem to use some version of "X icon with a shape behind it on hover":

screen shot 2017-05-19 3
screen shot 2017-05-19 2
screen shot 2017-05-19 1

It probably is worth revisiting how the close button in the tabs look. For now, I think it would be best to just merge this with what I suggested above, and revisit it later.

PalmerAL:
I've filed an issue about the keyboard shortcut to close a tab not working in the overlay: #357.

PalmerAL:
I'd like to go ahead and merge the first part of this PR before adding more stuff to it. Would it be possible for you to move the last set of commits (b9bb43f to 9bb4156) to a new branch (I think this will work: https://stackoverflow.com/a/1628584)? That way, I can go ahead and merge the first part of the changes and review the keyboard shortcut changes later.

code-hunger:
Never like --forced pushes but fortunately there was no problem here. Commits moved to [FixTaskOverlayKeyboardShortcuts](https://github.com/code-hunger/min/commits/FixTaskOverlayKeyboardShortcuts).
I suppose I could also move the refactoring commits to a different branch so that we have both things logically separated: the addition of close buttons and the refactoring. Just for clarity - if you don't mind. Otherwise there would be only one big commit doing everything inside.

PalmerAL:
OK, I've gone ahead and merged this part. Thanks for doing this!

Putting the refactoring in a separate branch probably would have been better, but its not worth trying to undo that.

PalmerAL:
I've updated the button styling to the above in 64fd595b6d43c030800642ef3b2482a48b2b8ea6. I agree that it doesn't look perfect, but I think it's good enough for now.

Issue #322, opened 2017-02-15
Can't change preferences of swyping two fingers to get previous page or go next page.


antondexter:
I can't navigate through large landscape pages with two fingers because MIN recognizes it as swipe gesture to get the previous page or go next page. Changing of system preferences of touchpad gestures on MacOSX to three fingers swipe doesn't work. MIN ignores system preferences of a touchpad.

vinayh:
I would also like to request this preference. On a ThinkPad, it can be easy to trigger horizontal scrolling with a TrackPoint (red pointing stick).

PalmerAL:
#284 should fix most of the accidental swipe gestures (A preference is still a good idea though).

PalmerAL:
This has been fixed in #358.

PR #358, opened 2017-05-21
Add preference to remove swipe for navigating back and forth between pages

mikopits:
Currently, the default behaviour is that swiping right and left will trigger going back and forth between pages in browser history. This can be problematic for those with sensitive hardware (see #322). This adds the option, as a checkbox in the preferences page, to disable this feature.


PalmerAL:
Thanks for doing this!

I'm not sure moving the swipe code to run in the window process will work, because the swipe code needs to have information about the content inside the webview. In particular, it needs to get the scroll position of the webview on line 47. Also, (although I haven't tested this extensively), I'm guessing this would increase the latency of swiping, since the events would need to be sent to the parent process before running the listeners.

I think a better approach would be to leave the existing swipe and zoom code in the webview like it is now. Then, instead of directly navgating (<a href="https://github.com/minbrowser/min/blob/master/js/webview/swipeEvents.js#L23">here</a>), the webview code could send an IPC to the parent with ```ipc.sendToHost('goBack')```. The parent process could then check the setting value, and either navigate or do nothing depending on the value of the setting.

mikopits:
Thanks for the feedback and the pointer in the right direction. I've made the changes you've suggested.

PalmerAL:
This looks much better!

Could you change the setting to be whether or not swipe navigation is enabled, with a default value of true? That would fit better with the other settings, which all enable features when checked.

Otherwise, this looks fine to merge.

mikopits:
The default behaviour is now a checked box, where checked means enabled. Green light?

PalmerAL:
Yup, looks fine to me! I've added a commit to simplify this a bit by checking if ```value === undefined``` rather than having a separate ```swipePreferenceSet``` pref in af6965487b5f9da24e7b8dc35f34d4d3880db4ae.

Thanks for adding this!

Issue #290, opened 2017-01-03
Restore Last Tab


jominions:
Since it's easy to click on the "close" button by mistake, I wonder whether there is a way to reopen the last tab? Like using the "cmd+shift+t" in firefox or chrome.

PalmerAL:
This has been added in 1.6.0.

PR #360, opened 2017-05-26
Add ability to restore closed tabs from the current session

mikopits:
Not sure how you would want this implemented, but here's a start at least.

The tab history is not persistent. Each session can store up to 20 closed tabs on a stack, older tabs are pushed out. No real reason for 20 other than it seemed reasonable. This replaces the shift+mod+t map for opening private tabs that was defined from the menu, now it's just on the custom keybinding. It's shift+mod+t because this is how most major browsers map it, afaik.

Fixes issue #290

### DIFF
--- a/js/tabRestore.js
+++ b/js/tabRestore.js
@@ -1,0 +1,33 @@
+const MAX_TAB_HISTORY_DEPTH = 20
+var tabHistoryStack = Array()
+
+var tabHistory = {
+  push: function (tabId) {
+    var tab = tabs.get(tabId)
+    // Do not store private tabs or blank tabs
+    if (tab.private || tab.url === 'about:blank') {
+      return
+    }
+
+    if (tabHistoryStack.length < MAX_TAB_HISTORY_DEPTH) {
+      tabHistoryStack.push(tab)
+    } else {
+      tabHistoryStack.shift()
+      tabHistoryStack.push(tab)
+    }
+  },
+  restore: function () {
+    if (tabHistoryStack.length === 0) {
+      return
+    }
+
+    var newIndex = tabs.getIndex(tabs.getSelected()) + 1
+    var newTab = tabs.add(tabHistoryStack.pop(), newIndex)
+
+    addTab(newTab, {
+      focus: false,
+      leaveEditMode: true,
+      enterEditMode: false,
+    })
+  }
+}
--- a/js/tabRestore.js
+++ b/js/tabRestore.js
@@ -11,11 +11,10 @@ function TabStack(tabStack) {
 TabStack.prototype.push = function (tabId) {
   var closedTab = tabs.get(tabId)
 
-  // Do not store private tabs, blank tabs, or preferences
+  // Do not store private tabs or blank tabs
   if (closedTab.private
     || closedTab.url === 'about:blank'
-    || closedTab.url === ''
-    || closedTab.url.match(/^file:\/\/\/.*\/min\/pages\/settings\/index.html$/)) {
+    || closedTab.url === '') {
     return
   }
 
--- a/js/tabRestore.js
+++ b/js/tabRestore.js
@@ -1,15 +1,21 @@
-function TabStack() {
-  this.depth = 20
-  this.stack = []
+function TabStack(tabStack) {
+  if (tabStack) {
+    this.depth = tabStack.depth
+    this.stack = tabStack.stack
+  } else {
+    this.depth = 20
+    this.stack = []
+  }
 }
 
 TabStack.prototype.push = function (tabId) {
   var closedTab = tabs.get(tabId)
 
-  // Do not store private tabs or blank tabs
+  // Do not store private tabs, blank tabs, or preferences
   if (closedTab.private
     || closedTab.url === 'about:blank'
-    || closedTab.url === '') {
+    || closedTab.url === ''
+    || closedTab.url.match(/^file:\/\/\/.*\/min\/pages\/settings\/index.html$/)) {
     return
   }
 
@@ -20,29 +26,6 @@ TabStack.prototype.push = function (tabId) {
   this.stack.push(closedTab)
 }
 
-TabStack.prototype.restore = function () {
-  if (this.stack.length === 0) {
-    return
-  }
-
-  lastTab = window.currentTask.tabs[window.tabs.length - 1]
-
-  // Open the tab in the last slot
-  if (lastTab === undefined)  {
-    newIndex = 0
-  } else {
-    newIndex = tabs.getIndex(lastTab.id) + 1
-  }
-
-  if (isEmpty(tabs.get())) {
-    destroyTab(tabs.getAtIndex(0).id)
-  }
-
-  var newTab = tabs.add(this.stack.pop(), newIndex)
-
-  addTab(newTab, {
-    focus: false,
-    leaveEditMode: true,
-    enterEditMode: false,
-  })
+TabStack.prototype.pop = function () {
+  return this.stack.pop()
 }
--- a/js/tabRestore.js
+++ b/js/tabRestore.js
@@ -5,7 +5,7 @@ var tabHistory = {
   push: function (tabId) {
     var tab = tabs.get(tabId)
     // Do not store private tabs or blank tabs
-    if (tab.private || tab.url === 'about:blank') {
+    if (tab.private || tab.url === 'about:blank' || tab.url === '') {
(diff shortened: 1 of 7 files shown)


js/tabRestore.js
```diff
@@ -0,0 +1,33 @@
+const MAX_TAB_HISTORY_DEPTH = 20
+var tabHistoryStack = Array()
```
code-hunger: Great feature! Just a note: why don't you put the two tabHistory-related variables inside the `tabHistory` object?
Probably it's a matter of personal preference but this way you'll connect tightly the things that belong together and won't "expose" any variables irrelevant for the rest of the browser.

mikopits: You make a good point! The reason I don't do that is because I don't think the namespace pattern, as it is used here, allows it. Quite a bit of the source uses this same pattern, actually (take js/tabState.js for example). There are more careful approaches we could use to avoid naming collisions but I don't think it's an issue here.

js/tabRestore.js
```diff
@@ -0,0 +1,33 @@
...
+  },
+  restore: function () {
+    if (tabHistoryStack.length === 0) {
+      return
+    }
+
+    var newIndex = tabs.getIndex(tabs.getSelected()) + 1
+    var newTab = tabs.add(tabHistoryStack.pop(), newIndex)
+
+    addTab(newTab, {
```
code-hunger: I think you have to consider which task was the tab in when it was closed.
For example, if the tab was closed in task A but you're in task B now you should switch to task A before restoring.
And if task A has been closed meanwhile you'd ask the user wheter they want the tab in task B or reopen task A with that tab.

js/tabRestore.js
code-hunger: I was thinking about the `tabHistory` and the `tabHistoryStack` variables - one could easily missuse the one instead of the other, and taking in mind that both have a `.push()` method, the distinction between the two could be confusing in another context.
Either way, there are some proposals for module-based system (instead of concatenation) which would solve such problems.

mikopits: Oh, I see what you mean now. You're right, `tabHistory` should be private as there is room for misuse here.

js/tabRestore.js
mikopits: Good catch.

> For example, if the tab was closed in task A but you're in task B now you should switch to task A before restoring.
Agreed 100%

> And if task A has been closed meanwhile you'd ask the user wheter they want the tab in task B or reopen task A with that tab.

The problem I see with this is that electron offers no prompt dialog in its API. It might be more trouble than it's worth to implement that nicely on all platforms. I think we should just assume that the user meant to restore the task they destroyed.

mikopits: This is addressed in https://github.com/minbrowser/min/pull/360/commits/9d46b6d54428964ccdec282a30e4813d06f16b7d

PalmerAL:
Switching to a task to restore a tab doesn't seem like the right behavior to me. For example:

1. open task A.
2. Open a tab.
3. close that tab.
4. switch to task B.
5. Open a tab.
6. Close that tab.
7. Switch back to task A.
8. Press shift+cmd+T.

I would expect the tab from step 2 to be restored, not the one from step 5. It almost seems like there should be a separate history stack for each task, rather than a global one.

mikopits:
We could store a map of task ids to stacks:
```javascript
var stackMap = {}
// When a task is added do:
stackMap[task.id] = []
// When a task is destroyed do:
delete stackMap[task.id]
```

js/tabRestore.js
```diff
@@ -29,8 +47,30 @@ TabHistory.prototype.restore = function () {
...
+  var parentTask = stackTop.task
+
+  if (tasks.get(parentTask.id) === null) {
+    // The task has since been destroyed. Create a new one.
+    parentTask.tabs = []
+    tasks.add(parentTask)
+  }
+
+  switchToTask(parentTask.id)
+  lastTab = getSelectedTask().tabs[parentTask.tabs.length - 1]
```
code-hunger: BTW, `window.tabs` already stores the tabs in the current task (tabState.js, lines 175-176).

js/tabRestore.js
```diff
@@ -10,17 +10,35 @@ function TabHistory() {
...
+    || closedTab.url === 'about:blank'
+    || closedTab.url === '') {
     return
   }
 
   if (this.stack.length < this.depth) {
-    this.stack.push(tab)
+    this.stack.push({
+      tab: closedTab,
+      task: parentTask || getSelectedTask(),
```
code-hunger: I wrote about this below, too, but you can use directly the `window.currentTask` variable. I'm not really sure why the `getSelectedTask()` function exists at all - perhaps in some edge cases we cannot rely on the global variable? I think there're no such cases as both `switchToTab` and `switchToTask` set it correctly; it's probably left from some old code.
Maybe @PalmerAL has a better explanation?

code-hunger:
> We could store a map of task ids to stacks:
> 
> var stackMap = {}
> // When a task is added do:
> stackMap[task.id] = []
> // When a task is destroyed do:
> delete stackMap[task.id]

What about storing the stackMap inside tabState itself?
```
var tabState = {
  tasks: [], // each task is {id, name, tabs: []} // <--- add closedTabs here
  selectedTask: null
}

```

js/tabRestore.js
PalmerAL: I don't remember what the reason was; it looks like it didn't even work when it was added in 71870a32d2c93b2eb5a63c56daba1bcc3137f119. I think I probably just added it while working on adding tasks support and then forgot about it. We can probably just get rid of it.

mikopits:
I believe https://github.com/minbrowser/min/pull/360/commits/0b49d9c8f76d21e49423106c6a4ea42eff130eec covers everything mentioned here. Each task now has its own history stored in the tabState.

js/tabRestore.js
```diff
@@ -10,17 +10,35 @@ function TabHistory() {
...
 
 TabHistory.prototype.push = function (tabId) {
-  var tab = tabs.get(tabId)
+  var closedTab = tabs.get(tabId)
+  var parentTask
+
+  tasks.get().forEach(function (task) {
+    task.tabs.forEach(function (tab) {
+      if (tab.id === tabId) {
+        parentTask = task
```
mikopits: It's not annoying, I appreciate the review.

PalmerAL:
This looks pretty good.

I think it makes sense to split the storing of tabs from actually restoring the last tab. Maybe tabStack could have a ```getLastTab``` method that removes the last tab from the stack and returns it, and the actual opening of the tab could me moved into a separate function (or even inside the keyboard shortcut listener, this should just basically be an ```addTab``` call).

js/tabState.js
```diff
@@ -130,9 +131,17 @@ var tasks = {
       task = {}
     }
 
+    var tabStack
+    if (task.history) {
+      tabStack = Object.setPrototypeOf(task.history, TabStack.prototype)
```
PalmerAL: I think it would be better if the tabStack constructor could take an object to set the stack to. This could look like this:

```
new TabStack(task.history)
```

js/tabRestore.js
```diff
@@ -0,0 +1,48 @@
...
+}
+
+TabStack.prototype.restore = function () {
+  if (this.stack.length === 0) {
+    return
+  }
+
+  lastTab = window.currentTask.tabs[window.tabs.length - 1]
+
+  // Open the tab in the last slot
```
PalmerAL: Opening tabs next to the current one seems better, since it matches what happens when you open a new tab. Is there a benefit to opening it at the end?

mikopits:
Cleaned up a bit with the above suggestions.

Something that has come up from testing this feature is that Min's local pages (ie. preference pages) will be pushed into the tab history. I'm not certain they should be. Or, at least certain ones.

I added the following check in `TabStack.push` to not add the preference page to the history.

```javascript
tabRestore.js:18
----------------

closedTab.url.match(/^file:\/\/\/.*\/min\/pages\/settings\/index.html$/)
```

This is very hacky. Does it make more sense to add, say, a boolean flag to a tab indicating if it is a local Min page? Or does it make more sense to just add it to the history and forego this altogether?

PalmerAL:
I think it's fine to add internal pages to the stack (and that's what I would expect anyway, since the history page/reader mode etc. act just like a normal tab in every other way).

PalmerAL:
LGTM!

PR #369, opened 2017-06-18
improved: Now looks beautiful in windows and linux, too

ProPuke:
This removes the system titlebar and menubars in windows and linux, for a more "native" min experience.

temp

I've added a menu button in replace of the menubar, and an autosizing drag region to the right of the tabs so the window can be dragged, as well as interacted with.
I've also made a few local improvements for windows and linux.
I've also added an option to preferences for using the standard system titlebar if a user prefers it (or you're rolling a wm in linux where you need the standard os chrome).
I've used windows 10 themed caption buttons in windows, and a similar but slightly more generic style in linux. This may not match the layout/style in linux, depending on user configuration. If this is the case system titlebars can be used instead (which still look a lot cleaner without the menubar)

Bit of a big pr, sorry. Tried to keep it fairly trim and in-style.

Full list of changes:

- added: min-themed titlebar in windows and linux, with draggable caption area
- changed: menubar removed and replaced with menu button to the left of tabs
- added: option in preferences for using system titlebars if the new min-style is undesirable
- added: css classes for styling based on os (mac, windows or linux)
- improved: button icons on the titlebar were not pixel-precise and featured some aa artifacts on windows and linux
- fixed: shortcuts mapped to the `alt` key would interrupt system shortcuts in windows
- changed: default keyboard shortcuts adjusted for windows and linux (alt-left/right is now back/forward, f5 is now refresh, etc)
- improved: preferences now shows localised key mappings (no longer says option/mod when not on a mac); Mappings now in titlecase for easier reading, and no longer malfunction if specified by the user in non-lowercase
- added: "settingsUpdated" event, to allow immediate application of settings without restart (currently only applied this to darkmode)
- improved: darkmode can now be toggled in preferences live, without restarting min
- changed: startup config changed from windowBounds.json to windowSettings.json and now also stores whether to use the system titlebar
- added: main.js is now messaged when the renderer is ready, so no need for nasty setTimeout hacks
- fixed: error display page would throw an error, when attempting to display an unknown error
- added: context menu to tabs


PalmerAL:
This is awesome!

I'll start reviewing this over the next few days; it may take me a bit to review since there's so many changes.

PalmerAL:
I'm not sure how I feel about the tabs not taking up the full width of the navbar. It seems like it would be better and more consistent to have them fill the full width, and make the whole navbar draggable like it was before:

screen shot 2017-06-18

However, it does look a bit odd to have the two styles of buttons next to each other.

ProPuke:
Yeah, I would have preferred that. Sadly you can't have client interactions inside the caption area in windows and linux, so wouldn't be able to detect clicks on the tabs if it was also a drag region.

I did play about with have artificial dragging instead. (literally moving the window upon mousemove events) and it worked okay; But it didn't support proper os drag events (like being able to drag a window to the top of the screen to maximise, or sides to dock, in windows) so felt weird.

The only real solution seemed to be to have a separate non-interactive area just for dragging (like opera and edge in windows) :S

ProPuke:
Whoops, just noticed I've broken keyboard shortcuts somehow, will chuck a fix in in a mo

ProPuke:
okay, fixed that now. Ended up using the [https://www.npmjs.com/package/electron-localshortcut](electron-localshortcut) library to install the window shortcuts, as it uses electron and is compat with the same menu shortcut format. We should probably be using that instead of mousetrap.

Also, added context menus to tabs, as that's expected on windows and linux.

PalmerAL:
> Yeah, I would have preferred that. Sadly you can't have client interactions inside the caption area in windows and linux, so wouldn't be able to detect clicks on the tabs if it was also a drag region.

Ugh, that's annoying. 

Some other things we could do:

Have a separate titlebar above:

screen shot 2017-06-18 2

Have a few pixels of space to drag the window:

screen shot 2017-06-18 3

(In this example, everything above the top of the tab highlight would be draggable, and everything below it would be clickable).

I think it's important to have the tabs fill the whole width of the window. This makes it consistent with other platforms, fixes the issue of the task overlay button looking weird when the overlay is open, and makes it possible to see multiple tabs even when the window is shrunk to a narrow width.

Like you mentioned above, having a separate title bar takes up a lot of space. Right now, I'm leaning towards option 3 (having a draggable area at the top of the window), but I havn't tested how easy dragging will be with this option (and I don't use windows enough to know how natural this seems compared to other applications).

js/navbar/titlebarButtons.js
```diff
@@ -0,0 +1,61 @@
...
+    case 'win32':
+      if (browserWindow.isMaximized()||browserWindow.isFullScreen()) {
+        titlebarWindowsMaximise.hidden = true
+        titlebarWindowsRestore.hidden = false
+      } else {
+        titlebarWindowsMaximise.hidden = false
+        titlebarWindowsRestore.hidden = true
+      }
+
+      browserWindow.on('maximize', function (e) {
```
PalmerAL: Putting callbacks in the renderer process breaks whenever the browser is reloaded, causing this error:

screen shot 2017-06-18 4

These probably need to be moved to the main process, and send IPC messages to the window.

ProPuke: Ahh whoops, sorted (curiously I don't get popup errors from that in my version of electron, just some notices in the terminal)

PalmerAL: That build was with an older version of electron, I think I've seen the terminal notices as well in newer versions.

index.html
```diff
@@ -19,12 +20,62 @@
 
 <body>
 
-	<i id="switch-task-button" class="fa fa-bars navbar-action-button theme-text-color windowDragHandle" title="View Tasks"></i>
+	<i id="switch-task-exit-button" class="fa fa-bars navbar-action-button active theme-text-color windowDragHandle" title="View Tasks" hidden></i>
+
+	<div class="titlebar-windows">
+		<div class="element caption-minimise">
```
PalmerAL: Did you make the icons yourself, or did you get them from somewhere else? If so, it might be good to add a comment with a link to the original source.

js/navbar/navbarTabs.js
```diff
@@ -239,15 +271,26 @@ function createTabElement (data) {
 
   // click to enter edit mode or switch to a tab
   tabEl.addEventListener('click', function (e) {
-    if (e.which === 2) { // if mouse middle click -> close tab
-      closeTab(data.id)
-    } else if (tabs.getSelected() !== data.id) { // else switch to tab if it isn't focused
+    if (tabs.getSelected() !== data.id) { // else switch to tab if it isn't focused
```
PalmerAL: remove "else" from the comment

js/navbar/navbarTabs.js
```diff
@@ -1,6 +1,36 @@
 var navbar = document.getElementById('navbar')
 var tabContainer = document.getElementById('tabs')
 
+var navbarTabClicked = null;
```
PalmerAL: maybe rename this to ```contextMenuSelectedTab```?

js/util/defaultKeyMap.js
```diff
@@ -1,21 +1,32 @@
...
+    defaultKeyMap.goBack = 'mod+left'
+    defaultKeyMap.goForward = 'mod+right'
+    defaultKeyMap.reload = 'mod+r'
+    break;
+
+  default:
+    defaultKeyMap.showMenu = ['option+f', 'option+e']
+    defaultKeyMap.goBack = 'option+left'
+    defaultKeyMap.goForward = 'option+right'
+    defaultKeyMap.reload = ['mod+r', 'f5']
```
PalmerAL: I think F5 should be one of the defaults on mac as well.

js/util/defaultKeyMap.js
```diff
@@ -1,21 +1,32 @@
...
+switch(navigator.platform){
+  case 'MacIntel':
+    defaultKeyMap.goBack = 'mod+left'
+    defaultKeyMap.goForward = 'mod+right'
+    defaultKeyMap.reload = 'mod+r'
+    break;
+
+  default:
+    defaultKeyMap.showMenu = ['option+f', 'option+e']
+    defaultKeyMap.goBack = 'option+left'
```
PalmerAL: should mod+left also go back on windows?

js/util/settings.js
```diff
@@ -58,11 +60,12 @@ var settings = {
     }).then(function () {
       settings.loaded = true
 
-      settings.onLoadCallbacks.forEach(function (item) {
```
PalmerAL: I think I like the old way better here. Is there a reason for changing this?

js/webviews.js
```diff
@@ -87,6 +87,20 @@ bindWebviewIPC('goForward', function () {
   })
 })
 
+bindWebviewIPC('updatedSettings', function () {
+  settings.clearCache()
+
+  darkModeUpdate()
```
PalmerAL: I don't think this is necessary here - clearCache will call load(), which will dispatch a settingsUpdated event, which will cause darkModeUpdate to happen anyway.

js/webviews.js
```diff
@@ -87,6 +87,20 @@ bindWebviewIPC('goForward', function () {
...
+
+  settings.get('systemTitlebar', function (value) {
```
PalmerAL: It might be good to move this somewhere else (not sure where, though).

js/webview/default.js
```diff
@@ -3,3 +3,7 @@
 var electron = require('electron')
 var ipc = electron.ipcRenderer
 var webFrame
+
+window.addEventListener('minUpdatedSettings', function () {
+	ipc.sendToHost('updatedSettings')
```
PalmerAL: We should check if the current page is a file:/// origin before sending an IPC message.

PalmerAL:
@ProPuke Is it OK if I add some commits to this?

doreminiature:
@ProPuke hello, nice job you did here. Will you be interested to help us with [Miniature ](https://github.com/doreminiature/miniature)project?

razzintown:
My two cents: Tabs can be reordered in the tasks view, so perhaps it's not a problem if they are not draggable in the webpage view. 

I consider it a part of the minimal paradigm to not reorder tabs while viewing pages. It's easier to manage them in the tasks view, because more title text is visible and more tabs are visible. Each view is best at its own job.

PalmerAL:
@ProPuke I'd like to start merging parts of this into Min. Are you still interested in working on this?

PalmerAL:
I'm going to start adding parts of this into the windows-support branch so it can be merged. @ProPuke Let me know if you'd like to continue working on this.

PalmerAL:
I've started merging parts of this in. Right now, I'm planning on doing the second mockup above, with a few pixels of drag space above the window. @razzintown I tried making the tabs part of the window drag area, which would allow us to fit everything on one line, but like @ProPuke discovered, this makes click events not work, meaning you can't click on the tabs to switch to them.

Also, it turns out that as long as you set the application menu, keyboard shortcuts will still work even if it isn't visible, so we can just create a hidden menu rather than using electron-localshortcut.

PalmerAL:
I've created a PR with the first part of this in #472.

PalmerAL:
I've merged the windows UI changes in #472. @ProPuke Thanks for all the work you put into this!

If anyone has the opportunity, it would be great to have some people testing this before the 1.7.0 release.

I believe the other things from this PR that are left to merge are:

 - [ ] Linux UI changes
 - [ ] Tab context menu
 - [ ] System titlebar option (we might leave this out if the frameless version works reliably for everyone)
 - [ ] "button icons on the titlebar were not pixel-precise and featured some aa artifacts on windows and linux" - not sure what code changes exactly went with this, I'll need to look
 - [ ] "shortcuts mapped to the `alt` key would interrupt system shortcuts in windows"
 - [ ] "default keyboard shortcuts adjusted for windows and linux (alt-left/right is now back/forward, f5 is now refresh, etc)"
 - [ ] "preferences now shows localised key mappings (no longer says option/mod when not on a mac); Mappings now in titlecase for easier reading, and no longer malfunction if specified by the user in non-lowercase"
 - [ ] settings that don't require restart (I think this might be better to include with the overall settings rewrite that I should get around to soon)
 - [x] "error display page would throw an error, when attempting to display an unknown error" (this was already fixed in a separate commit).

Let me know if I missed anything.

antfu:
I believe this is already been implemented and merged by other PRs. As this PR has been idle for a long while and has a lot of conflicts to be solved, I am closing for now. If there are any uncovered enhancements, please feel free to open new PRs. Thanks for your contribution.

PR #370, opened 2017-06-18
fixed: middleclicking on tabs to close does not work in modern electron

ProPuke:
The `auxclick` event is now used instead for non-primary clicks


PalmerAL:
I think we can just merge this as part of #369.

PalmerAL:
Since #369 is being merged in parts, I'm going to merge this separately.

PR #379, opened 2017-07-07
Linux builds fails due to typo in application source directory

shinenelson:
Linux builds fails due to a typographical error in the `electron-installer-debian` `grunt` task. The task looks for `Min-linux-*` whereas `electron:linuxBuild` writes `__dirname` as `min-linux-*`. This change-set fixes that issue.


PalmerAL:
Thanks!

PR #386, opened 2017-08-20
remove setTimeout when opening (default) browser with link

will-stone:
Uses a custom event emitter to trigger when the browser should be sent the URL (from opening the browser when clicking a link and Min isn't running). This removes the `setTimeout` workaround.

### DIFF
--- a/main/main.js
+++ b/main/main.js
@@ -3,6 +3,8 @@ const fs = require('fs')
 const path = require('path')
 const app = electron.app // Module to control application life.
 const BrowserWindow = electron.BrowserWindow // Module to create native browser window.
+const CustomEventEmitter = require('events')
+const customEventEmitter = new CustomEventEmitter()
 
 var userDataPath = app.getPath('userData')
 
@@ -140,6 +142,10 @@ function createWindowWithBounds (bounds, shouldMaximize) {
       e.preventDefault()
     }
   })
+  
+  mainWindow.once('ready-to-show', function() {
+    customEventEmitter.emit('mainWindowReadyToShow')
+  })
 
   registerFiltering() // register filtering for the default session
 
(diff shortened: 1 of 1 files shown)


PalmerAL:
Thanks for doing this! I think we can make it a bit simpler by combining the logic for opening a url on macOS with the existing did-finish-load listener on <a href="https://github.com/minbrowser/min/blob/master/main/main.js#L167">line 167</a>.

In the ```app.ready``` listener, we could do something like this:

```
if(appIsReady) {
sendIPCToWindow(...
} else {
    global.URLToOpen = url;
}
```

Then on line 167, we could do this:

```
mainWindow.webContents.on('did-finish-load', function () {
    //the process.argv... statement that was outside this call would be moved here
    if(process.argv...) {
        ...
    } else if (global.URLToOpen) {
        sendIPCToWindow(...
    }
});
```

An event listener would be better if we plan on adding other messages to send in the future, but I don't think we plan on doing this, and it's best to avoid loading extra modules on startup.

will-stone:
No problem, I found the _solution_ whilst trying to fix the problem with one of my apps. You know this better than I, can you make the changes in the code so I can see them?

PalmerAL:
Sure, the fix is here: e17e5d27acdbb1b6bf8b4c7c93e93ad5a4c276fb.

will-stone:
Ah, that's a much nicer way of doing it. Thanks; that helped with my app too.

will-stone:
@PalmerAL Thanks for the pointers with the code above; as I previously mentioned, I needed it in an Electron app I was building. I just thought you might like to see it: http://wstone.io/browserosaurus/ It's a browser picker (when you click a link from an app) for macOS, and, of course, it supports Min :)

PalmerAL:
Interesting, thanks for sharing! I think you may want to add the file:/// protocol <a href="https://github.com/will-stone/browserosaurus/blob/master/package.json#L66">here</a> (example: https://github.com/minbrowser/min/blob/master/Gruntfile.js#L113), so that opening HTML files stored on the computer works. Also, you probably want to prompt to set browserosaurus as the default browser on startup, since I'm guessing a lot of people won't figure out how to change that from preferences.

will-stone:
Thanks @PalmerAL. I've made those changes https://github.com/will-stone/browserosaurus/releases/tag/v1.0.1 Feel free to log any issues on the repo if you have any other improvement ideas.

PR #387, opened 2017-08-20
add nvmrc file

will-stone:
build tool doesn't work with latest versions of node/npm. Add this file so users know which version of node works.


PalmerAL:
LGTM. It looks like the breakage in the latest version is caused by npm/npm#17781; once the next version of npm is released we should be able to update this to the latest version.

PR #401, opened 2017-09-15
Update de.json

fethomm:
Translation finished, checked twice for correctness and typos. One string is untranslated because I do not understand what it relates to. The string is: "hostsFileEntry": "Hosts file entry",
//this is a label used to indicate that a result is coming from the user's host file

I don't know what 'host file' means here. Maybe you can explain per mail and I will try to translate it.

### DIFF
--- a/localization/languages/de.json
+++ b/localization/languages/de.json
@@ -5,148 +5,148 @@
         /* Context menu items
         these are the items displayed in the menu when you right-click on a page
         */
-        "openInNewTab": "Open in New Tab",
-        "openInNewPrivateTab": "Open in New Private Tab",
-        "viewImage": "View Image",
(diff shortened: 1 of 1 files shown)


PalmerAL:
Thank you for doing this! I'll release a 1.6.2 this weekend with this included.

The "host file" string refers to the <a href="https://en.wikipedia.org/wiki/Hosts_(file)">hosts file</a> (which is /etc/hosts on Linux and macOS, or C:/Windows/System32/drivers/etc/hosts on Windows). It lets you set the IP addresses that the computer should use for certain domains, basically letting you override the DNS server you are using. When you add hosts to the hosts file, Min will autocomplete them for you in the searchbar:

screen shot 2017-09-15-2

So the "hosts file entry" string means that the suggestion is based on the contents of your host file.

PR #402, opened 2017-09-16
Update de.json

fethomm:
Translation finished, checked twice for correctness and typos. One string is untranslated because I do not understand what it relates to. The string is: "hostsFileEntry": "Hosts file entry",
//this is a label used to indicate that a result is coming from the user's host file

I don't know what 'host file' means here. Maybe you can explain per mail and I will try to translate it.

edit: translated the 'hosts entry' string.


PalmerAL:
Thanks!

PR #405, opened 2017-09-18
Added Portuguese language

smarquespt:


PalmerAL:
Thank you for translating this!

I think the file name and the identifier field should be ```pt-PT``` instead of just ```pt```? Both macOS and Ubuntu seem to offer only a "Portugese (Portugal)" and a "Portugese (Brazil)" option, and I see that you listed "Portugese (Portugal)" as the name. Other than that, looks good!

smarquespt:
@PalmerAL Ok, please check.

PalmerAL:
Looks good, thank you!

PR #408, opened 2017-09-23
Added Czech translation

BlueManCZ:


localization/languages/cs-CZ.json
```diff
@@ -0,0 +1,152 @@
+{
+    "name": "Česky",
+    "identifier": "cs-CZ",
```
PalmerAL: Based on the language list (https://electron.atom.io/docs/api/locales/) I think the correct language code is just "cs". Are you seeing something different on your OS?

localization/languages/cs-CZ.json
```diff
@@ -0,0 +1,152 @@
...
+        "defaultTaskName": "Dimenze %n", //this is the name used for newly-created tasks; %n is replaced with a number ("task 1", "task 2", etc)
+        /* find in page toolbar */
+        "searchInPage": "Vyhledávat na stránce", //this is used as the placeholder text for the textbox in the find in page toolbar
+        "findMatchesSingular": "%i z %t výsledku", //this and the next label are used to indicate which match is currently highlighted
+        "findMatchesPlural": "%i z %t výsledků",
+        /* Focus mode */
+        "isFocusMode": "Jste v režimu zaměření.",
+        "closeDialog": "OK", //used as a label for the button that closes the dialog
+        "focusModeExplanation1": "V režimu zaměření jsou všechny karty kromě aktuální skryté a nelze vytvářet nové karty.",
+        "focusModeExplanation2": "Můžete opustit režim zaměření odškrtnutím \"focus mode\" v nabídce zobrazení.",
```
PalmerAL: Should "focus mode" be translated here as well, since the actual menu item (```appMenuFocusMode```) is translated?

localization/languages/cs-CZ.json
```diff
@@ -0,0 +1,152 @@
...
+        "settingsPrivacyHeading": "Blokování obsahu",
+        "settingsContentBlockingToggle": "Blokovat trackery a reklamy",
+        "settingsBlockScriptsToggle": "Blokovat skripty",
+        "settingsBlockImagesToggle": "Blokovat obrázky",
+        "settingsEasyListCredit": {
+            "unsafeHTML": "Na základě EasyList a EasyPrivacy bez pravidel skrývání prvků. Vytvořili <a target=\"_blank\" href=\"https://easylist.to/\">The EasyList authors</a> - <a target=\"_blank\" href=\"https://creativecommons.org/licenses/by-sa/3.0/legalcode\">s licencí</a>."
+        },
+        "settingsAppearanceHeading": "Vzhled",
+        "settingsDarkModeToggle": "Použít tmavý režim",
+        "settingsSwipeNavigationHeading": "Swipe navigation",
```
PalmerAL: Is is possible to translate this and the string below it?

PalmerAL:
Thanks for doing this! I've added a couple of comments, other than that the formatting of this looks fine.

localization/languages/cs-CZ.json
BlueManCZ: That's good point. I don't know, what is correct.
My `locale | grep LANG=` returns **LANG=cs_CZ.UTF-8**.

localization/languages/cs-CZ.json
BlueManCZ: Another good point.
This should be automatic replaced in program! ;)

localization/languages/cs-CZ.json
BlueManCZ: Can you please explain me, what is this feature for? I can't get this to work on Linux so I don't know, how to correctly translate this.

localization/languages/cs-CZ.json
BlueManCZ: But `locale | grep LANGUAGE=` returns **LANGUAGE=cs:en_GB**.
Ok then, let's use only **cs**.

BlueManCZ: Check [this](http://www.localeplanet.com/icu/cs-CZ/index.html).
> Java Locale | cs-CZ

PalmerAL: What do you get if you open the developer tools in Min and type ```remote.app.getLocale()```?

localization/languages/cs-CZ.json
PalmerAL: It lets you place two fingers on your trackpad, and use a swiping gesture to navigate back and forwards, like this: https://youtu.be/6NBMHEWLSoc?t=95.

localization/languages/cs-CZ.json
BlueManCZ: > What do you get if you open...

Only `cs`.

localization/languages/cs-CZ.json
BlueManCZ: Ahh, I see! Ok, I'll try translate it.

BlueManCZ:
Can I ask you @PalmerAL, will be locale setting available in Min browser preferences?

PalmerAL:
@BlueManCZ Probably eventually. The problem is that we're storing settings in indexedDB, which takes a few seconds to load from when you open the browser. We need the strings before we can create the menubar or display any UI, so if we waited for the settings to load, Min would start much more slowly. So probably we need to convert settings to use a regular file, and then add a special API in the preload script that allows internal pages to read from the settings file.

BlueManCZ:
I'm a bit confused.
> we're storing settings in indexedDB

Did you mean browser global settings or only locales?

> if we waited for the settings to load, Min would start much more slowly.

This feature is now disabled? And Min doesn't wait for settings when starting?

PalmerAL:
@BlueManCZ Sorry for the confusion! Right now, all of the browser settings (things like whether content blocking is enabled, whether to use the dark theme, ...) are stored in an indexedDB database. Because it takes a few seconds to load the database, we have to wait before we can actually read the saved settings. Right now this isn't an issue, because we don't need to use any of the saved values right at startup. What this means is that we can show the initial UI, and then a few seconds later when the database loads, we can switch to the dark theme, enable content-blocking, etc. However, we need to know what language to use before we can display any UI. Right now, this doesn't matter very much, because we're just using whatever your operating system language is to decide what languge to use in Min, and getting the operating system data is pretty much instant. However, if we added a setting in Min to change the language, we would have to wait for the database to load, since the language you select could be different than the language your operating system is using. As a result, you would just see a blank window for several seconds until the database finishes loading, making Min feel much slower.

> This feature is now disabled?

The translations will be used, we just use whatever your operating system's language is instead of having a separate entry on the preferences page. In most cases, this will be what people want anyway - if your operating system UI is set to show in a language, you probably want the browser UI to use the same language.

PalmerAL:
Here's a diagram to hopefully explain this better:

Just using the operating system's language setting, we can show the UI almost immediately:
loading_diagram_old

If we have to wait for the settings database to load first, it becomes much slower:
loading_diagram_new

BlueManCZ:
Now I understand everything! Thank you for your long answer and diagram! It's very illustrative. :heart_eyes: 

> you probably want the browser UI to use the same language

Glad to hear it, It's probably better than separate locale entry on the preferences page.

I think Czech translate is complete and you can merge it. I look forward to releasing the next version of Min!

PalmerAL:
OK, will merge!

PR #420, opened 2017-09-24
Added simple tab dividers to make tabs more defined.

SpecialBomb:
I thought min was very confusing because it was hard to see where a tab started and began. Using CSS, I added a simple border to the right of tab items. It works pretty well, and does not seem to get in the way of much. Thankfully, the color of the border matches the tab font color, making it contrast correctly.

Before:
tabdividers_before

After:
tabdividers

Hope ya like it.

### DIFF
--- a/css/navbarTabs.css
+++ b/css/navbarTabs.css
@@ -242,4 +242,5 @@ body.is-focus-mode .navbar-action-button {
 
 body.is-focus-mode .tab-item {
 	background: none !important;
+	border-right-style: none;
 }
--- a/css/navbarTabs.css
+++ b/css/navbarTabs.css
@@ -50,6 +50,13 @@ body.notMac #navbar {
 	/* https://github.com/w3c/csswg-drafts/issues/92 - don't make the scrollbar height affect the layout */
 	overflow: overlay;
 }
+
+/* for borders */
+.tab-item + .tab-item {
+	border-left-style: solid;
+	border-width: 1px;
+}
+
 .tab-item {
 	flex: 1;
 	min-width: 125px;
@@ -58,8 +65,6 @@ body.notMac #navbar {
 	font-size: 0.8em;
 	line-height: 36px;
 	height: 36px;
-	border-right-style: solid;
-	border-width: 1px;
 	overflow: hidden;
 	word-break: break-all;
 }
--- a/css/navbarTabs.css
+++ b/css/navbarTabs.css
@@ -58,6 +58,8 @@ body.notMac #navbar {
 	font-size: 0.8em;
 	line-height: 36px;
 	height: 36px;
+	border-right-style: solid;
+	border-width: 1px;
 	overflow: hidden;
 	word-break: break-all;
 }
@@ -119,6 +121,7 @@ body.notMac #navbar {
 	width: 0;
 	height: 0;
 	padding: 0;
+	border-style: none;
 	overflow: hidden;
 }
 
--- a/js/navbar/tabColor.js
+++ b/js/navbar/tabColor.js
@@ -286,12 +286,12 @@ function setColor (bg, fg) {
   if (fg === 'white') {
     document.body.classList.add('dark-theme')
     for (var i = 0; i < tabs.length; i++) {
-		tabs[i].style.borderRightColor = lightenDarkenHexColor(rgb2hex(bg), 20)
+		tabs[i].style.borderLeftColor = lightenDarkenHexColor(rgb2hex(bg), 20)
 	}
   } else {
     document.body.classList.remove('dark-theme')
     for (var i = 0; i < tabs.length; i++) {
-		tabs[i].style.borderRightColor = lightenDarkenHexColor(rgb2hex(bg), -20)
+		tabs[i].style.borderLeftColor = lightenDarkenHexColor(rgb2hex(bg), -20)
 	}
   }
 }
--- a/js/navbar/tabColor.js
+++ b/js/navbar/tabColor.js
@@ -224,9 +224,56 @@ function updateColorPalette () {
   }
 }
 
+// Below function from http://jsfiddle.net/Mottie/xcqpF/1/light/
+
+function rgb2hex(rgb){
+	
+ rgb = rgb.match(/^rgba?[\s+]?\([\s+]?(\d+)[\s+]?,[\s+]?(\d+)[\s+]?,[\s+]?(\d+)[\s+]?/i);
+ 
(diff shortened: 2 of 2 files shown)


SpecialBomb:
As you can see in the picture, I needed a bit of a cheatsheet \:P.

SpecialBomb:
Small problem I found, in focus mode, the divider is visible. Figuring out how to fix it now.

SpecialBomb:
Fixed it, will make another pull request if this one goes through.

PalmerAL:
I like the idea of separating tabs somehow, but I think these stand out too much (which can be a good thing when you're trying to pick tabs, but also can be distracting when you're not actively using the tabstrip). I'll try to think more about this tomorrow.

SpecialBomb:
Its pretty simple, I could easily add an opaque attribute to make it look
less intense.

SpecialBomb:
Never mind, it's harder than I thought. I can't simply apply opacity to borders without determining a color, making the automatic color handler useless with it. I'm gonna have to find the code that determines what color the text is depending on the color of the tabs and somehow get it to apply opacity specifically to the borders, which is possible.

Looks perfect to me to be completely honest.

SpecialBomb:
Another method is to use a straight half grey all of the time, but then automatic contrasting would be lost.

SpecialBomb:
Made an update that you will for sure like.

Browsing some js stuff, I managed to find a function that brightens and darkens colors through string manipulation. Using it, I set the color of the tab dividers to a brighter or darker version of the background used to set the tab color. This causes subtle dividers no matter what the color is. If a color is seen as dark, and darktheme is added to the page class, the color is lightened. Vice versa for adding lighttheme class. It looks nice, and I really hope you like it.

minreactivecolordividers

Note that the space under the navbar in that particular pic is because of the stackechange page, I didn't mess up the styling or anything.

More examples:

SpecialBomb:
Also to note, the darkening of inactive tabs also subtracts correctly, so the colors are still accurate when that happens. This can be seen in the third image the two right tabs.

PalmerAL:
Yeah, I like that much better! You actually don't need JS to set the color like that (interesting find though!), you can use ```rgba()``` to include opacity in the border color. To make the dividers white with opacity, you could do this:

```
.tab-item {
    border-color: rgba(255, 255, 255, 0.1);
}
```
Where the ```0.1``` is the opacity of the color.

I've noticed a couple of other things:

1. The border appears to the right of the last tab where it shouldn't.
2. The border overlaps with the border at the bottom of the navbar, resulting in a pixel that's darker:
screen shot 2017-09-25

 (there's an image there, GitHub's formatting it kind of weird so its hard to see).

To fix the first issue, I think you could use ```.tab-item + .tab-item``` and set the ```border-left``` property, which would make the border appear to the left of every tab except the first one. For the second issue, you could use a <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/::before">::before pseudo-element</a> to creat a 1px line, instead of using the border property. This would let you adjust the height, so it can stop before the bottom of the tab element.

Together, those would look something like this:

```
.tab-item {
    position: relative; /* this is needed to position the line at the edge of the tab */
} 

/* this is the element used to create the line */
.tab-item + .tab-item::before {
    content: "";
    position: absolute;
    top: 0;
    left: 0;
    width: 1px;
    height: calc(100% - 1px); /* the subtraction is so it doesn't overlap with the border at the bottom of the navbar */
    background-color: rgba(something); /* this would be the color of the line */
}
```

PalmerAL:
Also, I've been trying out using your branch of Min, and I think it makes sense to only show the borders when the tabstrip is hovered, since that's when they're most likely to be useful.

SpecialBomb:
Regarding color, the reason I couldn't just use that is because I am defining a set color. It also doesn't have active contrast, it can only be brighter and not darker where it needs to be to make visibility better.

With the border overlapping, it actually isn't. It may look like it, but it's actually because of that thin line at the bottom, because none of my tab dividers are opaque. I will try your method though.

I really like the idea of them only showing up once the mouse has entered the navbar, and that should be easy because of the mouseenter/exit events. I will definitely do that.

SpecialBomb:
Do you think getting rid of that thin opaque line at the bottom of the navbar will be okay? It doesn't exactly do much.

I changed `#navbar:before` to this:

```
#navbar:before {
    content: "";
    width: 100vw;
    height: 1px;
    position: absolute;
    bottom: 0;
    left: 0;
    background: rgba(0, 0, 0, 0);
    transition: 0.2s;
}
```
... and made the alpha in the `background` 0. This doesn't have any if at all of a visual impact, so would it be okay to just keep it at 0 opacity? I'm sure the before thing has an actual purpose, so I obviously just made it invisible.

SpecialBomb:
Did that fix to remove the rightmost divider next to the add new tab button.

PalmerAL:
Sorry, I've been pretty busy this week. I should have time to review this and the other PR tomorrow or Saturday.

SpecialBomb:
I realized what that little bar is for, so I won't remove it.

PalmerAL:
> Regarding color, the reason I couldn't just use that is because I am defining a set color. Your method also doesn't have active contrast, it can only be brighter and not darker where it needs to be to make visibility better.

I'm not sure it actually needs to be a fixed color though. It looks like the function right now always makes the border color 20 values (there's probably an actual name for these numbers, I'm not sure what it is though) darker than the background color. Assuming the tab background is pure-white, that would be equivalent to ```border-right-color: rgba(0, 0, 0, 0.078)```.  On a medium-gray background like ```rgb(150, 150, 150)```, the rgba approach would result in slightly lighter borders, but I wouldn't expect it to be very noticable - ```rgb(138, 138, 138)``` with the rgba approach vs ```rgb(130, 130, 130)``` with the JS-based one.

Regarding the contrast, I apologize for not mentioning it before, but there's a ```dark-theme``` class that gets added to the body when a dark color scheme (like with GitHub or Stack Overflow) is being used in the navbar. So to make the borders brighter when on top of a dark background, and darker when on top of a light background, you could do this:

```
.tab-item + .tab-item {
   border-left-color: rgba(0, 0, 0, 0.078);
}

.dark-theme .tab-item + .tab-item {
  border-left-color: rgba(255, 255, 255, 0.078);
}
```

The bar at the bottom is there to separate the navbar from the page content in cases like the Stack Overflow page, where the top of the page is the same color as what we use for the navbar. I agree it isn't very visible right now; I would actually support making it more visible instead of removing it, but that can be done in another PR.

PalmerAL:
Closing since this is being worked on in #1109.

PR #421, opened 2017-09-25
Added a back button in navbar

SpecialBomb:
Does what it says on the tin!

minhistory

It's a very basic implementation, but it works! I am hoping to add style editing for when you can't go forward or back anymore, so the arrow becomes less opaque or something.

Hope ya like it!


code-hunger:
I think this should be part of the preferences - whether the buttons are shown or not. Probably not all users want them on screen :)

SpecialBomb:
PLEASE FOR THE LOVE OF GOD ADD THESE IN.

... but I get your point. You have to bully me into making a menu item and everything? Blech. I would also argue not having essential controls from the UI, as history buttons are a sure and set standard. I will see what I can do, although I probably will end up asking you how to do the whole menu option thing. You could also argue the same logic with the task view and new tab buttons, as those technically don't need to be there and are accessable through shortcuts.

To be honest, if I ever heard someone saying while using firefox or chrome, "man, I sure wish these history buttons weren't in my way," I would die laughing.

SpecialBomb:
So, from what I can see, there is no "settings" file that needs to be annexed. Do you just assume it's there by using the settings.get(blah) and it just works? How are defaults set then, because swipe navigation is always on by default.

SpecialBomb:
Okay, I allowed it to be enabled and disabled in the settings. Problem is that now we need a translation for the data-string the checkbox uses.

SpecialBomb:
So, if you want me to add the graying out of the history buttons, then I need a method to get the history of the currently selected tab as an array. How would I do this?

code-hunger:
> PLEASE FOR THE LOVE OF GOD ADD THESE IN.
> 
> ... but I get your point. You have to bully me into making a menu item and everything? Blech. I would also argue not having essential controls from the UI, as history buttons are a sure and set standard. I will see what I can do, although I probably will end up asking you how to do the whole menu option thing. You could also argue the same logic with the task view and new tab buttons, as those technically don't need to be there and are accessable through shortcuts.
> 
> To be honest, if I ever heard someone saying while using firefox or chrome, "man, I sure wish these history buttons weren't in my way," I would die laughing.

Did I say something offensive/wrong/etc? Your tone makes me think I've done/said something seriously bad and I really don't see what it is. I never said anything about menu items or whatever, either.

I only meant this browser survived more than a year without these buttons and (almost) no one ever wanted it. I'm fine if you want them, but just don't force them on others - that's why I suggested making them optional.

Also, in my opinion, min is not a *standard* browser - it's a minimalistic browser. It's only visible UI is tha tabbar. There are no buttons, no menus (they're hidden, at least), the entire focus is on the webpages.  
That's what makes min unique.

SpecialBomb:
Nah, I was just trying to be funny, I didn't mean any harm. Obviously, it didn't translate very well through text. I will watch myself next time and be more formal here. There are some other comments I have made that are more aggressive for no good reason, and I will apologize for those as well.

When I was referring to "menu items," I was trying to ask how one would add an option, like "enable dark theme" to the menu. I figured it out after a little digging.

I see what you mean by Min being minimalistic, but I just think there are some things that are kinda necessary to browse the web efficiently. Going back a step in your history is almost essential in modern browsing due to the hub-style pages that dominate the net right now. You should be able to comfortably and efficiently browse using only your mouse to navigate. Going forward isn't too useful though, so I could happily get rid of it.

From how I see it, Min is perfect as soon as those history buttons are added, nothing else in the UI would be needed. back/forward, new tab, options, and that's it. I absolutely love this browser and how it looks, it's just that it was really hard to do stuff when I couldn't go back.

As a side note, I would still like to know how to access the current tab history as an array so I can gray out the arrows.

code-hunger:
https://github.com/minbrowser/min/blob/280bde25a227cf4fe38b3767e4fe3244c8a01b55/js/tabState.js#L159-L164

It's very confusing as this code creates a *task* but assigns it a *tab* history. I think *tabHistory* should be part of each *tab* in the *tabs* member of *newTask*.

SpecialBomb:
Okay, so then is there a way to compare the tabs current point in history to the main tabHistory stack?

Also, how would I even start to get the tabHistory? I'm assuming it starts with `getWebview(tabs.getSelected())`. I'm a bit of a JS newbie.

code-hunger:
Ok so I think `tabHistory` are the tabs in order they were last visited. Your question should be answered by these lines:

https://github.com/minbrowser/min/blob/15ff1d46b3d815c43051d6b8650e6fd6379c4489/js/keybindings.js#L109-L113

code-hunger:
> Also, how would I even start to get the tabHistory? I'm assuming it starts with getWebview(tabs.getSelected()). I'm a bit of a JS newbie.

Then do `console.log(getWebview(tabs.getSelected()))` and see what it has to offer.

SpecialBomb:
Okay, I found how to get the history array for the current tab:

`getWebview(tabs.getSelected()).getWebContents().history`

Now all I have to do is figure out the tabs position in the history array...

SpecialBomb:
So, there is another pair of functions that let me see if I am at my history limit already:

`getWebview(tabs.getSelected()).getWebContents().canGoBack()/.canGoForward()`

This is good, so now I can program the arrows to react to these statuses. Problem is that I don't know how to get them to constantly update, because I can only get them to check once.

PalmerAL:
> Nah, I was just trying to be funny, I didn't mean any harm. Obviously, it didn't translate very well through text. I will watch myself next time and be more formal here. There are some other comments I have made that are more aggressive for no good reason, and I will apologize for those as well.

Good to know! Its often difficult to tell what someone's thinking through text, so it's generally a good idea to indicate when you're being sarcastic/funny/etc. To add on to what @code-hunger said, it's difficult to find a balance between adding features and keeping Min minimalist and fast. The primary reason most people seem to have for using Min is that it's fast and the UI is unobtrusive, but almost every issue that gets opened here is a request to add more features. Individually, these requests seem fairly reasonable, but if we keep adding features over time, Min will get slower and slower, and the UI less and less minimalist, until you end up with something no better than any other browser today. We're never going to make everyone happy, and I don't think it's worth trying to do so by adding every possible feature, so the fact that a feature isn't implemented doesn't necessarily mean that it isn't a good idea, but just that it might not be a good idea to include in Min.

With a back button specifically, I think it's OK to add. The forward button gets used much less frequently in my experience (and also according to this Mozilla study from 2012: https://blog.mozilla.org/ux/2012/06/firefox-heatmap-study-2012-results-are-in/chrome_menus/ although I'm not sure if the results from that apply to Min as well), so I'm OK with just having a back button.

@code-hunger:

> It's very confusing as this code creates a task but assigns it a tab history. I think tabHistory should be part of each tab in the tabs member of newTask.

```tabHistory``` in that instance isn't referring to the previously-opened webpages within a tab, its referring to the previous tabs in a task that have been closed. The TabStack is used to keep track of these, so you can press shift+cmd+t to reopen a tab you closed by mistake.


@SpecialBomb:

> This is good, so now I can program the arrows to react to these statuses. Problem is that I don't know how to get them to constantly update, because I can only get them to check once.

Yeah, it doesn't look like Electron has a way to do that. You could listen for did-finish-load events, and update the buttons every time a page finishes loading, but I'm questioning whether we actually need to fade out the buttons at all. I haven't actually tried to use it yet, but seems like it would be more distracting than useful - you're just starting to look at a webpage, when something flashes in the navbar, making you move your eyes away from what you were looking at. Also, you have to be careful with webview methods, since they tend to be sort of slow, and if you're using them a lot that time adds up.

**Comment by SpecialBomb** [comment]
@PalmerAL Okay, I am fine with it not reacting anyway, I just wanted a good way to let the user know you can't go back or forward. Isn't nessisarily required.

>Also, you have to be careful with webview methods, since they tend to be sort of slow, and if you're using them a lot that time adds up.

As I stated above, I have changed my mind, so this won't matter anymore.

**Comment by SpecialBomb** [comment]
So, in that case, I think it's done. If there is anything else you want me to change, let me know.

It would also be nice to know if you guys are willing to merge this anyway.

**Comment by SpecialBomb** [comment]
Oh, you literally just said to get rid of the forward button, lol.

One sec.

**Comment by SpecialBomb** [comment]
How it looks now:

backbutton

**Comment by SpecialBomb** [comment]
uh oh, someone updated gruntfile. Well, I will fix that later I guess.

File: css/navbarTabs.css [hunk 1]
```diff
@@ -42,6 +42,16 @@ body.notMac #navbar {
 	background: none;
 }
 
+/* history buttons */
+
+#forward-button {
```
PalmerAL: This rule can be removed now.

js/navbar/goBackButton.js
```diff
@@ -0,0 +1,15 @@
...
+
+goBackButton.addEventListener('click', function (e) {
+  getWebview(tabs.getSelected()).goBack()
+})
+
+settings.get('historyButtons', function (value) {
+  if (value === false)
+  {
+	  goBackButton.style.display = 'none'
+	  goBackButton.removeAttribute('transition')
```
PalmerAL: I don't think this is necessary?

localization/languages/en-US.json
```diff
@@ -99,6 +99,7 @@
         },
         "settingsAppearanceHeading": "Appearance",
         "settingsDarkModeToggle": "Enable dark mode",
+        "settingsHistoryButtonsToggle": "Enable back button",
```
PalmerAL: should be "button" singular, since there's only one button

pages/settings/index.html
```diff
@@ -46,6 +46,9 @@ <h3 data-string="settingsAppearanceHeading"></h3>
 		<div class="setting-section">
 			<input type="checkbox" id="checkbox-dark-mode" />
 			<label for="checkbox-dark-mode" data-string="settingsDarkModeToggle"></label>
+			<div class="spacer"></div>
+			<input type="checkbox" id="checkbox-history-buttons" />
```
PalmerAL: "buttons" should be singular here as well

PalmerAL:
Looks good to me, withe the exception of a couple minor comments I've added.

SpecialBomb:
@PalmerAL https://github.com/minbrowser/min/pull/421/files/31a5d272fe08a79f12d2d73b3259e713add687fa#diff-2aeb67ec86e7eb1a4adb42aa8e8d0c68

This is very important, it's essentially what disables the back button when you have it disabled in the menu. The display none is obvious, by the transition part was to keep it from loading in and then disappearing with an animation. That part forces it to be instantly invisible.

PalmerAL:
I'm kind of confused now. ```transition``` isn't a standard HTML attribute, and we don't have any CSS rules for it, so I wouldn't expect it to do anything. Also, that attribute isn't added to the button in the first place, so I would expect that calling ```removeAttribute``` wouldn't have any effect.

Testing this, I can see the flash you mentioned with or without the ```remoteAttribute``` call. I think this is happening because the IndexedDB that contains the settings loads after the main window, and we still show the back button while it's loading. This should be fixed once we switch to using a regular file for the preferences; in the meantime, we could consider making the button hidden initially and then showing it when the preferences load.

SpecialBomb:
@PalmerAL Well, one of your main developers directly added that css attribute in to create the smooth animations in the navbar. It's mainly in the navbar elements, take a look at it in the code, or inspect it in min.

Transition is an animation attribute, its new to css3, dunno where you are getting your information on that. I'm assuming you are using a textbook or something? Here is W3school's thing on it [https://www.w3schools.com/css/css3_transitions.asp](https://www.w3schools.com/css/css3_transitions.asp).

by using `removeAttribute`, it removes the `transition` style from the backbutton element, which disables it's attached animation from the navbar parent since it inherits it from the navbar. This is the simplest way to remove the animation if the menu option is not enabled.

Your solution is correct though, so go ahead and make the change if you need to. The arrow is visible for a single frame before hiding itself. I'm really busy right now in college.

PalmerAL:
Ah, OK, I think I get what you're saying now. ```removeAttribute``` is for HTML attributes, not CSS properties. So calling ```removeAttribute``` would change this:

```
<div transition>
```
to this:

```
<div>
```

What you're trying to do is change this:
```
<div style="transition:something">
```
to this:

```
<div style="">
```

So for that, we would need to replace the ```removeAttribute``` call with this:
```
goBackButton.style.transition = "none";
```
or just add another rule to the CSS.

I'll go ahead and make that change and merge this.

PalmerAL:
Done. Thanks for working on this! I realize this took a bit longer than it should have; I'll try to improve my communication in the future.

Issue #414, opened 2017-09-24
Remove tab switching delay


SpecialBomb:
When I click on a tab, there is a short 1 second delay until I can click another one. This is annoying.

Edit:

I did some experimentation, and it turns out that there may not be a delay at all. It has to do something with moving your cursor while you click, which for some reason does not register. This means that your cursor has to be 100% still to click on a tab. So, if you are used to quickly clicking through your tabs, this is a really big UI problem.

SpecialBomb:
No matter how hard I look, I simply can't find why this happens. Blech. Where does Min handle mouse inputs?

PalmerAL:
I think this is because of these lines:

https://github.com/minbrowser/min/blob/280bde25a227cf4fe38b3767e4fe3244c8a01b55/js/default.js#L82-L113

The intention of this was to fix #311, which was a bug where dragging the window while the mouse is hovering over a tab (which is possible on macOS, since the tab bar and the window buttons are combined into one line), caused a tab to be clicked when you didn't want it to. We probably need to add something to this to allow a little bit of movement (maybe 5px?) before the click-canceling behavior kicks in.

SpecialBomb:
I see, will attempt tomorrow.

SpecialBomb:
So, I think the best solution is to do this is to put a big if statement around all of this that checks if you are using a mac or not, the difference being that mac users can't quickly click tabs. In my opinion, that is the sacrifice you have made by putting part of the UI in the grab bar, but that's just me.

If you want to fix all of this, I would give the mac version the standard border back and keep the tabs within the window.

You could also specifically for mac add a dedicated drag bar, but you would still drag from the tabs.

PalmerAL:
That would work on Linux and Windows, but it wouldn't fix the issue on Mac. I think (although I could be wrong) that when you're clicking quickly on a tab, the mouse is only moving a few pixels between the time when you press the mouse button down and when you release it. So when we update the ```isDragging``` variable, we could also keep track of how much the mouse is moving, and when the mouse is released, we can only cancel the click if the mouse has moved more than a certain amount. That should hopefully fix the issue on every platform.

PalmerAL:
This was fixed by #423.

PalmerAL:
The 1.7.0 version that includes this fix has been released, you can download it from here: https://github.com/minbrowser/min/releases/tag/v1.7.0.

PR #423, opened 2017-10-01
Fixed Issue #414: Added mouse movement threshold for clicking on tabs

ianfare:
Issue #414 was that if the mouse is moving at all, it is not possible to click on tabs. So sometimes when trying to click on a tab, if the mouse moves a few pixels between mouse press and release, the click is cancelled. I have added code that records the mouse position upon pressing the mouse on a tab, and again upon releasing it. The click is only cancelled if the distance between the location of the mouse press and the mouse release is above a small threshold. 

Hope this is to everyone's liking!

### DIFF
--- a/js/default.js
+++ b/js/default.js
@@ -81,13 +81,28 @@ function empty (node) {
 
 /* prevent a click event from firing after dragging the window */
 
+var mouseX = 0
+var mouseY = 0
+
+var mouseLocation = function(e) {
+  mouseX = e.screenX
+  mouseY = e.screenY
+  if (mouseX < 0){mouseX = 0}
+  if (mouseY < 0){mouseY = 0}
+  return true;
+}
+
 window.addEventListener('load', function () {
   var isMouseDown = false
   var isDragging = false
+  var recordX = 0
+  var recordY = 0
+  var pauseRecord = true
 
   document.body.addEventListener('mousedown', function () {
     isMouseDown = true
     isDragging = false
+    pauseRecord = false
   })
 
   document.body.addEventListener('mouseup', function () {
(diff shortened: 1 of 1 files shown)


PalmerAL:
Thanks for fixing this! The mousemove event has [movementX](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/movementX) and movementY properties, so I think we could simplify this a bit by just having a ```distance``` variable. On mousedown, we could reset distance to 0, and in the mousemove event, we could do this:
```
distance += Math.abs(e.movementX) + Math.abs(e.movementY);
```
This would be slightly less accurate, since we would be guessing at the distance rather than actually calculating it, but it should work just as well for this purpose.

ianfare:
I've changed it to use movementX and movementY as you suggested!

js/default.js
```diff
@@ -115,24 +102,15 @@ window.addEventListener('load', function () {
     dragHandles[i].addEventListener('mousemove', function (e) {
       if (isMouseDown) {
         isDragging = true
-        if (!pauseRecord){
-          mouseLocation(e)
-          recordX = mouseX
-          recordY = mouseY
-          pauseRecord = true
-        }
+        distance += Math.abs(e.movementX) + Math.abs(e.movementY)
```
code-hunger: Sorry to interrupt here but as soon I see this the deep in my mind explodes with the need of two squares and sqrt(). 

E.g.: a single movement of 5px by *x* and 5px by *y* has a distance of nearly 7, while you'd calculate it as 10.
For this case it won't probably matter much but I just wanted to point it out.

ianfare: That's the way I had it when I initiated the pull request. When @PalmerAL suggested the use of movementX and movementY, they gave an example for calculating distance:

`distance += Math.abs(e.movementX) + Math.abs(e.movementY);`

I figured that was deliberate, for simplicity or to avoid calculating a lot of square roots. But if we decide that the mathematically rigorous way of doing it is preferable, I can do that!

PalmerAL: Whatever way you want to do it is fine. I don't think we actually need the accuracy here, and I figured we could simplify this a bit by just estimating, but if you'd prefer to actually calculate the distance that'd be fine too.

ianfare: I also think the sum of absolute values is fine here. When a click should be canceled, the distance should be much larger than the threshold anyways. With the sum of absolute values, we avoid having to compute square roots on the mousemove event, which I think is worth it.

PalmerAL: OK, I'll go ahead and merge this.

PalmerAL:
Thanks!

Issue #411, opened 2017-09-24
MailTo links wont work


SpecialBomb:
When a mailto link is clicked, the system default email client does not respond to it.

PalmerAL:
This is fixed in #428.

incanus:
I am seeing this again — `mailto:` links don't do anything. Mac Big Sur, latest Min version.

PalmerAL:
This regressed in 10d312f97e803d0c7f5525a78e723ec0e3a359b4, because the protocol handler is being registered to the wrong session now (see [here](https://www.electronjs.org/docs/api/protocol#using-protocol-with-a-custom-partition-or-session): "If you don't specify a session, then your protocol will be applied to the default session that Electron uses").

PR #428, opened 2017-10-04
Fixes issue #411 - Adds mailto protocol handling

matheus208:
Fixing #411 

This commit adds a protocol handler for "mailto" links. It only delegates the url to the default system mail application.

As a disclaimer, I'm quite new to electron projects (and open source in general), so any input is appreciated.

I thought of creating a separate protocol handling file (so we could handle not only mailto, but magnet links, stream, etc), but wasn't too sure about changing a lot of things, so I just changed enought to fix the issue.

### DIFF
--- a/main/main.js
+++ b/main/main.js
@@ -2,6 +2,8 @@ const electron = require('electron')
 const fs = require('fs')
 const path = require('path')
 const app = electron.app // Module to control application life.
+const protocol = electron.protocol // Module to control protocol handling
+const shell = electron.shell // Module to call system functionality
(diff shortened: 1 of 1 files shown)


PalmerAL:
Looks good to me, thanks! I think it's fine to leave this in main.js for now, we can always move it if we decide to support more protocols in the future.

One minor thing: I think electron.shell should be used inline, since in most cases it will never be used, and not loading it on startup should reduce memory usage a bit. I'll add that in a separate commit.

Issue #192, opened 2016-06-01
Suggestion: A real kicker... Native support for Flash-based OnDemand streams


baldrick-nz:
Adobe ditching Linux support, and more ondemand options popping up, having a one-stop for VOD services would be a definite 'opt-in' advantage for Min.

Currently only Chrome offer in-built native support for Netflix, but sadly this doesn't extend to other services as well.

PalmerAL:
Min is unlikely to support Flash anytime soon, but in the long-term it probably won't be necessary. There are now HTML5 alternatives for flash, and a lot of sites are moving to those. <a href="http://www.tomshardware.com/news/chrome-to-block-flash-content,31807.html">Chrome is blocking all Flash content by the end of the year</a>, with only a few sites given a temporary exemption. Adobe has mostly stopped development on Flash Player, and it's becoming less and less common to see sites that use it.

Netflix actually doesn't use flash player - it uses a different plugin, widevine (which Min also doesn't support: see #161).

PalmerAL:
There is a JS-based flash player, <a href="https://github.com/mozilla/shumway">shumway</a>, that could be integrated into Min, but it looks like it's been abandoned, and I don't think it supports video anyway.

dbfreem:
Since Flash isn't really support be by alot of other browsers and this is from 2016, closing.

PR #432, opened 2017-10-10
Add zh-CN translation

caixw:


PalmerAL:
Thank you!

nwdxlgzs:
Fix lost Chinese.When will the next build be made?Can you add pepperflash.dll? Or do we modify the js content to implement it ourselves? It is recommended to add a method that references the system default pepperflash.dll.
[zh-CN.txt](https://github.com/minbrowser/min/files/3332989/zh-CN.txt)

PalmerAL:
@nwdxlgzs Can you make a pull request for updating the translations?

> When will the next build be made

I'm not sure yet, whenever we have enough changes to justify making a new release.

> Can you add pepperflash.dll?

We're not going to support it (#192), but there are instructions for how to enable it here: https://electronjs.org/docs/tutorial/using-pepper-flash-plugin.

PR #437, opened 2017-10-18
Find in page: Follow link with ctrl-return

klorenz:


PalmerAL:
Thanks for adding this! Is this a keyboard shortcut that already exists in another application? In Firefox, ctrl+return seems to toggle the "highlight all" option, and in Chrome and Safari it doesn't seem to do anything. I'm not opposed to adding a new shortcut, but we would need to make sure it doesn't conflict with anything else that already exists (if other applications are already using it, it's probably safe to use).

Since this is a multiple-key shortcut, it probably makes sense to use Mousetrap and put the shortcut in js/keybindings.js like this:

```
defineShortcut('followFoundLink', function() {
  findinpage.end({action: 'activateSelection'})
}
```

If we do it this way, selecting between the ctrl and cmd keys based on the OS will work automatically, and it would also be possible to set a custom shortcut in the preferences.

code-hunger:
> Chrome and Safari it doesn't seem to do anything
Use `Tab` to focus a link in Chrome, Ctrl+Enter opens it in a new tab then.

klorenz:
Hi @PalmerAL,

the behaviour, I implemented, is the same with Chrome and Chromium on Linux.
- Ctrl+F for finding something,
- Return for next result
- Ctrl+Return for follow link
Best regards, Kiwi

PalmerAL:
OK, sounds fine to me. As mentioned above, I think this would be better as a Mousetrap shortcut, since that lets people modify the shortcut from settings and handles the OS detection part automatically. If you don't have time to do that right now, I'd be happy to fix that as well.

PalmerAL:
Sorry for taking so long to get to this! I've updated this so the shortcut is customizable. Otherwise everything looks fine. Thanks for the PR!

PR #442, opened 2017-10-21
Redesign the PDF viewer

PalmerAL:
This PR replaes PDF.js' built-in PDF viewer with a custom one for Min (the actual PDF rendering is still based on PDF.js, we're just replacing the viewer UI). This has a bunch of advantages:

* The design better matches the rest of the browser.
* It works with the browser's built-in find-in-page function.
* It works with the browser's native zoom controls.
* It feels much faster, because we can hide pages that aren't currently visible, which improves frame rates when scrolling a lot.

Compared to the old viewer, I've removed a bunch of features that I'm guessing aren't used very much. These are:

* Document outline
* Presentation mode
* Ability to create bookmark URL's (which let you tell the viewer to open the PDF at a certain point)
* The ability to rotate the page
* The ability to drag the page to scroll it
* Document properties window

I don't want to remove anything that gets used frequently, so if you use these features, please comment on this PR so we can decide if they should be included or not.

Things to do before releasing:

 - [x] clean up the code
 - [x] fix printing not working on large documents
 - [x] fix saving the PDF from file > save as not working (pressing cmd+s works fine)
 - [x] add a migration for bookmarked PDF's to the new viewer URL
 - [ ] remove the old viewer code

Other things that can be fixed after release:

 - [ ] add support for annotations


PalmerAL:
I'm going to remove the old viewer after merging this.

Issue #440, opened 2017-10-20
Inform users that Electron-based browsers are insecure


anko:
From https://electron.atom.io/docs/tutorial/security/:

> When working with Electron, it is important to understand that Electron is not a web browser.
>
> [ ... ]
>
> With that in mind, displaying arbitrary content from untrusted sources poses a severe security risk that Electron is not intended to handle. In fact, the most popular Electron apps (Atom, Slack, Visual Studio Code, etc) display primarily local content (or trusted, secure remote content without Node integration) – if your application executes code from an online source, it is your responsibility to ensure that the code is not malicious.

It is clear that Electron is not intended to be a framework for building a web browser. Tracking Chromium releases (which contain security updates) is not a priority for its maintainers, so min may be arbitrarily out of date, and arbitrarily insecure.

I think the readme should **loudly** warn users about this.

doreminiature:
You are absolutely right! @PalmerAL , I suggest [Muon](https://github.com/brave/muon) instead.

PalmerAL:
Right, there are issues with this. I think it's worth pointing out that the paragraph quoted above is kind of misleading regarding Min. What the docs are saying is that, unlike a regular webpage you load in a browser, which can't do a lot to affect the local computer, the URL you give Electron to load in a window has complete access to the computer. This is completely true - any code run in the context of the Min UI (which is anything in this file: https://github.com/minbrowser/min/blob/master/index.html) has full node access. However, webpages in Min aren't loaded directly into this file - they're loaded using a <a href="https://electron.atom.io/docs/api/webview-tag/">webview</a> tag, which behaves much more like a regular webpage with the limited permissions you would expect.

The lag in Chromium releases is an issue. Part of the problem is that Electron lags behind Chromium, and part of the reason is that our releases lag behind Electron releases. For example, right now, we're using Electron 1.7.8, which is based on Chromium 58. The latest Electron release is based on Chromium 59; we should be able to release more frequently to help avoid this. Meanwhile, the latest Chromium release is 62, so this obviously doesn't solve the whole issue.

[...]

PR #445, opened 2017-10-30
Readme: Warn about late security updates

anko:
As explained in https://electron.atom.io/docs/tutorial/security/ and discussed in #440, Electron isn't made for writing browsers.  I think potential users deserve at least a note about it in the readme, as the general expectation otherwise is for web browsers to prioritise security.

This is just a suggested wording. Feel free to reword.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,8 @@ Min is a smarter, faster web browser. It includes features such as:
 
 More information, and prebuilt binaries are available [here](https://minbrowser.github.io/min/).
 
+> :warning: [Browsers are officially not a supported Electron use-case](https://electron.atom.io/docs/tutorial/security/).  Any security updates may come very late!  Use min at your own risk!
+
 ## Screenshots
 
 ![The search bar, showing information from DuckDuckGo](http://minbrowser.github.io/min/tour/img/searchbar_duckduckgo_answers.png)


apm:
This is like saying "Min is a minimal web browser that never should be used for that purpose".

PalmerAL:
Hmm, I'm not sure I like linking to that page, because most of the problems it mentions don't apply to Min, as I mentioned in #440 (and I also don't think it's right to suggest that no one should ever use Min because of this). I'm thinking about this:

> Note: Min uses an older version of Chromium, which may be missing security fixes from later versions. [More Information](https://github.com/minbrowser/min/issues/440#issuecomment-338080554)

anko:
@PalmerAL I like it!

PalmerAL:
Great. Added in 1c28905ad7a14ed4cdad34daa1b47d73222a59d5.

apm:
Maybe you could adjust the #440 title to be more gentle as well.  I reacted to 'Electron based browsers are insecure'.  Nobody wants an insecure browser.  To note that your browser may not have the latest security fixes since it lags behind Chromium releases is appropriate, but I'm under the impression that you do not think the browser is insecure, just less than optimal.  If it is insecure, then that needs to be addressed.  I think Electron should be more interested in keeping Chromium up to date, since the first thing people think when seeing the project is "Hey, I could make a custom browser with html and javascript... cool".

PalmerAL:
Yeah, it probably does make sense to change the issue title to something better. To clarify what is mentioned there, it's hard to judge what "insecure" means exactly. Min definitely isn't as secure as using Chrome would be, but it's not horrible either; I don't think the risk from it is high enough that I would be concerned about using Min. Regarding Electron updates, it's tricky, because Electron is really closely integrated with the Chromium's internal API's, which are changing constantly with every release. Given the difficulty of upgrading (see this PR for example: https://github.com/electron/electron/pull/10213) and the amount of people working on Electron, they actually do a pretty decent job.

PR #452, opened 2017-11-18
Add ja translation

LEoREo2247:


PalmerAL:
Thanks for translating!

Could you also add a translation for the PDF viewer string that's at the end of the en-US.json file: https://github.com/minbrowser/min/blob/e1ee07870997a86d4d1f740a7cd7cd01b0621b49/localization/languages/en-US.json#L152-L153

Otherwise, looks good!

LEoREo2247:
Added!

PalmerAL:
Thanks!

Issue #52, opened 2016-04-17
Add `loading activity` or `progrss bar` when loading web page


clonezer:
Please add something to make sure the browser is loading when clicking link or  open new tab

jkanarek:
Something like this could work:
![loadingbar](https://cloud.githubusercontent.com/assets/1163147/14659902/d5ee3370-0654-11e6-9827-b551ec9f33c9.png)

PalmerAL:
It seems kind of busy, though, since there are three different shades of green in the tabstrip. Also, having two different colors under the same text will likely be a problem, since the text may be readable on top of one color but not the other.

jkanarek:
True, could just do a faint line on the bottom of the tab, like this:
![untitled_2](https://cloud.githubusercontent.com/assets/1163147/14660129/75ef8b9c-0657-11e6-9a5d-c501e847640d.png)

TheInitializer:
There already is one...

PalmerAL:
@TheInitializer No, there isn't.

TheInitializer:
I get a little blue progress indicator directly under the tab bar... maybe I'm seeing something else?

idle-zealot:
There's a bar like that on Google websites.

TheInitializer:
It must have been something else, I tried loading a few pages and didn't see it.

TheInitializer:
@idle-zealot I've only been on GitHub since I installed min, lol

TheInitializer:
Oh lol, Github has one for its issue thing. *facepalm*

jkanarek:
@PalmerAL how do you feel about the thin line approach? Mobile Safari and FB's browser both use this technique and I think it's pretty effective.

PalmerAL:
A thin line would work pretty well. What should the algorithm for changing the width of the line be? There's no way to measure actual progress, so the progress shown will just be a guess.

Smona:
Rather than a progress bar, maybe a loading indicator. Like Chrome has. But instead of that circular icon, you could use a sliding line on bottom, fading animation on the tab, etc.

ashishwadekar:
@PalmerAL Considering the issue about algorithm, a spinner or an animated line would do just fine which will indicate that the page loading process is not yet complete. On internet connections have slow speed, I have found myself rechecking the tab if the page is loading or not and have refreshed page lot many times.

abejfehr:
[...]

PR #453, opened 2017-11-18
This commit adds visual feedback while the page is loading

danfragoso:
When using this browser on a slow connection i've missed some feedback if the page was loading or not.

captura de tela 2017-11-18 as 13 20 58

### DIFF
--- a/css/navbarTabs.css
+++ b/css/navbarTabs.css
@@ -260,7 +260,7 @@ body.is-focus-mode .tab-item {
 	position: absolute;
 	top: 0;
 	left: 0;
-	margin: 0 0.75em;
+	margin: 0 0.8em;
 }
 
 .p0 {
--- a/css/navbarTabs.css
+++ b/css/navbarTabs.css
@@ -254,25 +254,33 @@ body.is-focus-mode .tab-item {
 
 /* progress bar */
 
-.progress-bar {
+
+.progress-bar-container {
 	height: 2px;
-	background-color: currentColor;
+	width: 100%;
 	position: absolute;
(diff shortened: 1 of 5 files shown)


PalmerAL:
I like the idea of showing some kind of progress bar while the page is loading, but it's kind of distracting to have the icons in the tab suddenly shift when clicking a link.

I wonder if an actual bar (maybe just a thin line under the tab, like in https://github.com/minbrowser/min/issues/52#issuecomment-212186989) would work better.

danfragoso:
Maybe it is, i will try to work on the bar.

danfragoso:
I've changed it, the behavior now is similar to safari.

PalmerAL:
I like it! I'll add some commits to fix a few minor issues I've noticed.

One thing I'm wondering about is that on a page that loads pretty quickly, the progress bar seems to make the page load feel slower. I'm wondering if it would make sense to do something like we do on the PDF viewer, where we show no progress bar at all for the first couple seconds, and then move to the p25 state if the page still hasn't loaded. That way, pages that load quickly would still seem to just appear without any wait, and for pages that take longer to load, you would still get feedback that the page is loading. What do you think?

danfragoso:
That would be great, but i don't really have the time to do it for now. It would be nice if you did it.

PalmerAL:
OK, that's fine, I will do it.

danfragoso:
Nice, thanks!

PalmerAL:
I think I've finished everything, so I'm going to go ahead and merge this. Let me know if you notice something that should be changed.

PR #454, opened 2017-11-18
Updated pt-br translation with new back key option

danfragoso:


PalmerAL:
Thanks!

Would you also be willing to add a translation for the new PDF Viewer string?

https://github.com/minbrowser/min/blob/e1ee07870997a86d4d1f740a7cd7cd01b0621b49/localization/languages/en-US.json#L152-L153

Issue #451, opened 2017-11-17
Switch to reading settings from a file instead of IndexedDB


PalmerAL:
This should allow us to avoid the slow startup time of IndedDB, so we can read settings right at startup (which in turn allows us to introduce settings like making session restore optional, or changing the menu bar keybindings).

PalmerAL:
This will be fixed in #766.

Issue #357, opened 2017-05-20
mod+w from the task overlay doesn't work correctly


PalmerAL:
This should close the tab and update the overlay.

code-hunger:
(let me call the task overlay *TO*)
There's a more general problem with the TO currently: keyboard events are sent to the browser "*behind*" it and not to the TO itself.

For example, if you press Ctrl+Tab in the TO you'll switch to the next tab "*behind*".
Or, if you open a youtube video and open the TO, pressing "k" will stop the video.

This behaviour is undesired - I think the TO should be a standalone tab-like page with its own keyboard shortcuts. This will fix the issues above and will allow a much easier impl of keyboard control of the TO (which, in turn, will close a number of issues).

code-hunger:
I just noted another similar bug and fixed it in https://github.com/minbrowser/min/pull/352/commits/897d6e0bed37c1c7b61901d464ec82c1b0fc4437
Ctrl+Tab now updates the focused tab element successfully.

PalmerAL:
Rather than adding code for each individual listener, I think it would be better if tabState had a way to listen for changes to the tab state. Then the task overlay could just listen for changes and update its contents whenever the state changes. That would reduce a lot of duplication, and fix all of these bugs without adding too much complexity.

code-hunger:
Looking at the [MDN Guide](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Creating_and_triggering_events) is should be pretty simple to implement this.

For example, for all close buttons we can add this:
```
var tabClosedEvent = new CustomEvent('tabClosed', {
  task: task,
  tab: tab,
  source: /* e.g. TaskOverlay or NavbarTabs */
});
document.dispatchEvent(tabClosedEvent);
```
Then we could create various listeners for this event, like:
```
// in taskOverlay:
document.addEventListener('tabClose', function(e) { taskOverlay.update(t.task); });
// in navbarTabs:
document.addEventListener('tabClose', function(e) {
  destroyTab(e.tab);
  // remove navbar tab element
});
```

PalmerAL:
It would probably be even simpler to have something like ```tasks.listenForChanges(callback)``` that would be called whenever the state of a task or its tabs change.  Then the task overlay could re-render the task element corresponding to the task that was changed.

Syndamia:
@PalmerAL is this still relevant? Also, I'm not sure I understand what the issue is.

PalmerAL:
[...]

PR #463, opened 2017-11-26
Fixes #462: Show previous match in Find in page when Shift is pressed

### LINKED ISSUE #462: Find on page - Shift+F for previous match
Generally, in GUIs where Ctrl+F searches strings on a page, pressing Enter/Shift+Enter goes to the next/previous match, respectively. Min currently supports only forward searching, and it would be great if previous matches are also accessible.

code-hunger:
It turned out to be very easy to achieve. Here's the solution.

I don't know why all these commits are shown. The only difference between my and min's master is the file I just edited. I suppose it's beause you squashed my previous PR.


PalmerAL:
Thanks! I think it would be a good idea to eventually convert both ```return``` and ```shift+return``` to use Mousetrap, so that they can be customized from the settings page (see the example here: https://github.com/minbrowser/min/pull/437/files), but if you'd prefer to leave it this way for now that's fine as well, and I'll go ahead and merge this.

The extra commits shouldn't matter too much - I'll just squash this into one commit when I merge it.

code-hunger:
I'll do it - and I'll change the mapping to `/` - as in vim.

Now as I see the code there will need some refactoring to avoid too much duplication; hope you don't mind it.

PalmerAL:
I wouldn't make that the default mapping, since that would make it impossible to search for a / in the page. Otherwise, sounds good!

code-hunger:
I think I ran into a problem. If I set `Enter` for `findInPageNext` in the default keymap, wouldn't this make every `Enter` in the page a `findInPageNext`? Also, I think this is such a 'deep' option that it's not woth adding it to the settings. Who would ever want to change `Enter` to go to the next match anyway? It's the most natural and almost everything else would be harder to use and remember (I mean ctrl/mod combinations. Other keys are out of question because they'll change the text searched).

PalmerAL:
Yeah, I kind of see your point. I guess it's fine to just leave this as a hardcoded option (but to answer your question, you could have the shortcut only trigger when find in page is active by checking whether ```findInPage.activeWebview !== null``` before doing anything).

PalmerAL:
If we're not going to make the shortcut customizable, this looks fine; I'm going to merge it.

code-hunger:
You're quick; I've done some changes I think are helpful anyway so I'm going to commit them, too.

PalmerAL:
OK, great! I don't think GitHub will let me merge a PR twice, so you'll probably have to open a separate PR once you're done.

code-hunger:
They're not showing up here so you seem to be right. It's not a big change - I made the findInPage shortcut customizable; this leads me to a question though - where is this shortcut _currently_ stored? I ran a search through the entire repo for `mod+f` but I got only the lines I personally edited.

PalmerAL:
There's a find in page menu item, so the shortcut is here: https://github.com/minbrowser/min/blob/master/main/main.js#L316 .

I don't think you want to remove that shortcut, since it's nice to have the shortcut show up in the menu item next to the name. Once #451 is fixed, it will be possible to make the menu item shortcuts customizable, but for now we can either leave it the way it is, or make the customization happen in addition to the default mod+f.

code-hunger:
Hmm.. so the settings (especially the keybindings) are retrieved _after_ the menubar is loaded, and you can't read them before displaying the menu?

Another thing - I've tried to do this some time ago - and I think I've mentioned it somewhere already - to make the task overlay keyboard-operatable. When having a lot of tabs it'd be quite useful if we could search through them easily, which is currently impossible as all the keybindings go to the page 'behind' the overlay.

Do you have an idea how to make it behave like a separate page? Or at least send key events to it?

PalmerAL:
> Hmm.. so the settings (especially the keybindings) are retrieved after the menubar is loaded, and you can't read them before displaying the menu?

Yes, because the settings are stored in IndexedDB which takes a couple of seconds to load, and we can't wait that long to create the menu. The plan is to switch to a file, which would be fairly quick to read from, allowing us to read the settings and then create the menu.

> Another thing - I've tried to do this some time ago - and I think I've mentioned it somewhere already - to make the task overlay keyboard-operatable. When having a lot of tabs it'd be quite useful if we could search through them easily, which is currently impossible as all the keybindings go to the page 'behind' the overlay.
> Do you have an idea how to make it behave like a separate page? Or at least send key events to it?

You'd have to set the focus on the element you want to receive keypresses (and in order to make an element focus-able, you have to add ```tabindex="-1"``` to it). We'd probably want to do something like the searchbar, where the current tab (the one highlighted in gray) is focused, and pressing the arrow/tab keys changes the focus to the next one (the code for that is here: https://github.com/minbrowser/min/blob/master/js/searchbar/searchbar.js#L252). Beyond that, I think the idea in #357 would be best: have some kind of ```tabState.onChange``` listener, and re-render the overlay whenever that changes. That way, shortcuts like mod+w would work automatically without any additional code.

Also, you actually can search through open tabs right now from the searchbar. If you enter a few characters of the title or URL, you should see something like this:

screen shot 2017-11-26 at 5 12 45 pm

PR #469, opened 2017-12-08
add setting to disable task overlay on task switch

toraora:
This adds a setting to disable the task overlay popup on task switch (default mod+[ and mod+]). 

For me, it's visually jarring and doesn't provide much utility (especially in dark mode, the selected tab is barely visible).

### DIFF
--- a/js/keybindings.js
+++ b/js/keybindings.js
@@ -1,194 +1,89 @@
 /* defines keybindings that aren't in the menu (so they aren't defined by menu.js). For items in the menu, also handles ipc messages */
 
-ipc.on('zoomIn', function () {
-  getWebview(tabs.getSelected()).send('zoomIn')
-})
+settings.get('taskAnimationDisabled', function(taskAnimationDisabled) {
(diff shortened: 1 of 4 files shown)


PalmerAL:
I'm not sure adding a setting for this is the right idea - it seems more like a design flaw that we should fix in general. Maye if we made the animation less distracting? Another option would be to not show anything if you press and release the keys immediately, but show the task overlay if you hold the keys down for a bit.

The thing about dark mode sounds like a bug, I've created a new issue for it: #470.

toraora:
I like the latter idea. Maybe after holding the keys for long enough, you enter a mode where you can cycle through tasks in the overlay by repeating the key patterns (I think this is how the current implementation was intended to feel)

PalmerAL:
BTW, I've fixed the bug with dark mode in 02d1276661fce45dc61f49bcb6b99ac9f231b104.

PalmerAL:
The animation has been removed completely in #784, so this shouldn't be necessary any more.

PR #479, opened 2018-01-13
Translate settingsKeyboardShortcutsHelp (fr-FR)

g-ongenae:
Hello,

I have a little doubt about this one. 
The user should write something like my example?
```
^n: open, ^q: close
```

Also, I saw that it's written `"Swipe navigation" // same in french`, however I would (rather) say `Navigation par swipe`. We don't really have word/verb for swipe, I think that the cause of the issue.

### DIFF
--- a/localization/languages/fr-FR.json
+++ b/localization/languages/fr-FR.json
@@ -100,13 +100,13 @@
         "settingsAppearanceHeading": "Apparence",
         "settingsDarkModeToggle": "Activer le mode nuit",
         "settingsHistoryButtonToggle": "Activer le boutton navigation",
-        "settingsSwipeNavigationHeading": "Swipe navigation", //same on french
-        "settingsSwipeNavigationToggle": "Activer swipe pour changer de page.",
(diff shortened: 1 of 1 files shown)


PalmerAL:
Thanks for the PR!

> The user should write something like my example?

What that text means is that in each of the textboxes, if you want multiple key combinations that trigger that action, you can place commas between each key combination, like "command+a, command+b, command+c". For example, the default setting for "switch to previous tab" is "option+command+left, shift+ctrl+tab, shift+command+pageup", which means that any of those three shortcuts will switch tabs.

> Also, I saw that it's written "Swipe navigation" // same in french, however I would (rather) say Navigation par swipe. We don't really have word/verb for swipe, I think that the cause of the issue.

That seems reasonable, but I don't know enough about French to answer that. Maybe @cubmann has an opinion?

cubmann:
We can replace swipe navigation by navigation en glissement maybe, in
french there is no word for this.

PalmerAL:
OK, sounds fine to me.

PR #483, opened 2018-01-21
Update ja.json & Swipe gestures indicators

LEoREo2247:
I just filled missing translation and fixed some very small translation error and added horizontal swipe gestures indicators like chrome

### DIFF
--- a/css/webviews.css
+++ b/css/webviews.css
@@ -32,3 +32,40 @@ body.windows #webviews {
 	height: 100vh;
 	z-index: 99;
 }
+
+.arrow-indicator {
+	width: 140px;
+    height: 140px;
+    border-radius: 70px;
+    background: rgba(0,0,0,0.8);
+    position: fixed;
+    top: 50%;
+    z-index: 1;
+    transition: opacity 0.3s, visibility 0.3s;
+  	opacity: 0;
+  	visibility: hidden;
+}
+.arrow-indicator.shown {  
+	opacity: 1;
+	visibility: visible;
+}
+#left {
+	left: -70px;
+}
+#right {
+	right: -70px;
+}
+
+.arrow {
+	color: white;
+    font-size: 40px !important;
+    margin-top: 50px;
+}
+#leftArrow {
+    float: right;
+    margin-right: 15px;
+}
+#rightArrow {
+    float: left;
+    margin-left: 15px;
+}
--- a/css/webviews.css
+++ b/css/webviews.css
@@ -42,8 +42,8 @@ body.windows #webviews {
     top: 50%;
     z-index: 1;
     transition: opacity 0.3s, visibility 0.3s;
-  	opacity: 0;
-  	visibility: hidden;
+	opacity: 0;
+	visibility: hidden;
 }
 .arrow-indicator.shown {  
 	opacity: 1;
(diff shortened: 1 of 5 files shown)


css/webviews.css
```diff
@@ -32,3 +32,40 @@ body.windows #webviews {
 	height: 100vh;
 	z-index: 99;
 }
+
+.arrow-indicator {
+	width: 140px;
+	height: 140px;
+	border-radius: 70px;
+	background: rgba(0,0,0,0.8);
```
PalmerAL: Since there's another rule below to set the opacity, can this just be ```#000```?

css/webviews.css
```diff
@@ -32,3 +32,40 @@ body.windows #webviews {
...
+	position: fixed;
+	top: 50%;
+	z-index: 1;
+	transition: opacity 0.3s, visibility 0.3s;
```
PalmerAL: There's no transition for ```visibility```, so I think that part can be removed without affecting anything.

css/webviews.css
```diff
@@ -32,3 +32,40 @@ body.windows #webviews {
...
+}
+#left {
+	left: -70px;
+}
+#right {
+	right: -70px;
+}
+
+.arrow {
+	color: white;
```
PalmerAL: The indentation here doesn't match the other lines

index.html
```diff
@@ -61,7 +61,14 @@
 
 	<div id="overlay"></div>
 
-	<div id="webviews"></div>
+	<div id="webviews">
+		<div id="left" class="arrow-indicator">
```
PalmerAL: Can you make the id's of these something more descriptive, like "leftArrowContainer"?

css/webviews.css
```diff
@@ -32,3 +32,40 @@ body.windows #webviews {
...
+	background: rgba(0,0,0,0.8);
+	position: fixed;
+	top: 50%;
+	z-index: 1;
+	transition: opacity 0.3s, visibility 0.3s;
+	opacity: 0;
+	visibility: hidden;
+}
+.arrow-indicator.shown {  
+	opacity: 1;
```
PalmerAL: I think it might be better to have this less opaque - maybe around 0.6?

js/webviews.js
```diff
@@ -274,6 +274,22 @@ webviews.bindIPC('goForward', function () {
     })
 })
 
+webviews.bindIPC('showBackArrow', function() {
+    var backArrow = document.getElementById('left')
+    backArrow.classList.toggle('shown')
+    setTimeout(function() {
+      backArrow.classList.toggle('shown')
+    }, 800)
```
PalmerAL: 800 seems a little too long, maybe somewhere around 500-600?

PalmerAL:
I've added a few minor comments. Otherwise, this looks pretty good!

LEoREo2247:
Alright, I've updated my code based on feedback.

```css
.arrow-indicator {
	width: 140px;
	height: 140px;
	border-radius: 70px;
	background: #000;
	position: fixed;
	top: 50%;
	z-index: 1;
	transition: opacity 0.3s, visibility 0.3s;
	opacity: 0;
	visibility: hidden;
}
```
Though, `visibility` seems to be needed. Otherwise, arrow indicator's fade-out animation won't be applied.

PR #490, opened 2018-01-28
Polish translation + improvements

JB1905:
+ Polish translation
+ window rendering improvements
+ Electron version update

### DIFF
--- a/css/webviews.css
+++ b/css/webviews.css
@@ -32,3 +32,40 @@ body.windows #webviews {
 	height: 100vh;
 	z-index: 99;
 }
+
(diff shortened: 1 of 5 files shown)


JB1905:
Added in https://github.com/minbrowser/min/pull/491
On iMac late 2012 (basic model) with macOS 10.13.4 app startup time almost identical
after the implementation of the window display delay.

PalmerAL:
Yeah, the actual startup time is probably the same, but I think it feels slower because the window doesn't appear immediately after you open the app. I've opened a new issue about this: #492.

PR #491, opened 2018-01-28
Polish translation + zoom fix

JB1905:
+ Polish translation
+ disabled browser interface zoom on macOS (double tap on Magic Mouse)


PalmerAL:
Thanks!

PR #497, opened 2018-02-15
fixed it

clunt-westwood:
Switched to native, runs much faster now.


PalmerAL:
I get that you apparently don't like Electron, but is there an actual problem that you're having? I've spent a lot of time making Min run really fast, and from my testing most actions complete in under 10-20ms, which means it should run fairly well even on an older computer. If that's not the case for you, can you open an issue so we can figure that out?

Issue #364, opened 2017-06-08
set homepage?


somenxavier:
Is it possible to set home page (default to min browser page). Thx.

matthewxu:
> Is it possible to set home page (default to min browser page). Thx.

I still cannot find how to set home page, could do me a favor?

PalmerAL:
@matthewxu There isn't a setting to do this. If you wanted, you could build Min yourself, and change [this line](https://github.com/minbrowser/min/blob/master/js/tabState/tab.js#L6) to something like:

```
      url: tab.url || 'http://example.com',
```
and it should work.

matthewxu:
I have change the code to set the home page when  minbrowser startup. It is hard coding.....

somenxavier:
@PalmerAL So it is not possible to set homepage at user level.

matthewxu:
I think will should add config on setting page for user

code-hunger:
In the context of min it doesn't really make sense to have a homepage. You have the navbar, which has access to most of the browser functionallity, implements search suggestions etc. The purpose of the blank tab (where the homepage should be opened) is just to show a white page until you choose to where to navigate.

It is an intermediate state which you only go through, you have nothing to there apart from entering a url or going to a search engine.

PalmerAL:
I don't think we want to support this as an option, for the reasons @code-hunger mentioned. However, it should be possible to modify the code and get this to work without many changes.

Everything works correctly on startup when I make the change suggested, above, I'm not sure why it isn't working for you. In order to see the homepage immediately, you would need to change the `addTab` function so it doesn't open the searchbar automatically [here](https://github.com/minbrowser/min/blob/master/js/api-wrapper.js#L49) - you could probably just remove this if statement entirely.

There isn't much documentation about how the code works unfortunately, I'll try to write some at some point. In general, everything is kind of split into a few different sections:

* `tabState`, which is the actual objects that store the data about the tabs you have open.
* `webviews.js`, which controls the actual `BrowserViews` that display a page.
* `tabBar.js`, which controls the tab bar.
* `searchbar.js`, which controls  the searchbar.

Almost everything else is just adding features that use those API's.
To add a homepage, you would probably need to make that URL the default URL that is used when a new tab object is created (which is the change I suggested above), and then modify `addTab` to change what happens when a new tab is opened.

PR #501, opened 2018-02-17
Standard Startscreen/Website

ONIXTurtle:
Hey there,
I just wondered if it would be possible to have a Website always shown as a start website. I mean like in Google Chrome showing always Google on startup.

Thank you


PalmerAL:
I believe that's this issue: #364.

PR #508, opened 2018-02-27
Create norwegian language file

xhaavardx:


localization/languages/no.json
```diff
@@ -0,0 +1,156 @@
+{
+    "name": "Norwegian Bokmål",
+    "identifier": "no",
```
PalmerAL: I'm somewhat confused about what the right identifier for this is. The Electron docs list "no" as a generic "norwegian" option, while "nb" corresponds to "norwegian (Bokmal)". On macOS, there are two separate "norwegian" and "norwegian bakmal" options listed in the preferences, but both seem to correspond to "nb". Given that Wikipedia says Bokmal is the most popular form, maybe we want to update buildLocalization.js to accept an array of identifiers, so that this file will be used for both?

PalmerAL:
Thanks for working on this! Before merging this, would it be possible to translate the rest of the strings listed in the file? If you'd rather not do that, that's fine to, I can go ahead and merge this once we've addressed the issue with the identifier.

localization/languages/no.json
Val0Tnra: So, I know I'm comming at this from nowhere, but as a norwegian myself I would like to point out that the "correct" thing to do would be to **not** use the identifier "no" (norwegian) at all -- and then use "nb" for norwegian bokmål. If you at some point added norwegian nynorsk, the identifier should then be "nn". The reason for this is that -- though "bokmål" is more popular -- the two languages are *equal* and should thus not be treated as anything other than equals.

I know this seems like beeing way to strickt but, trust me, there is **a lot** of feelings involved in the whole "bokmål" vs. "nynorsk" discusion, and some people (myself very included) get extramely ofended when "bokmål" is refered to simply as "norwegian" in contexts like these.

xhaavardx:
That's totally fine with me. Whatever floats your (the other guy) boat.
It's just a minor thing. It'll be bokmål either way :)

Regarding the translation: I've been a bit busy lately so I haven't had
time to do finish them all, but I really want to get it done. I'll get at
it soon unless progress has been made by my fellow Norwegians ;)

PalmerAL:
@xhaavardx Ok, sounds good!

abrarShariar:
hey @xhaavardx did you get some time to finish off the translations then we can close this issue :)

xhaavardx:
To be honest, I never was able to finish it. But I could make a quick and
dirty attempt so we can have a working translation, and once it's
officially a part of the repository, make incremental improvements to it.

PalmerAL:
This is fairly outdated at this point, but feel free to open a new PR if you want to continue working on it again.

Issue #303, opened 2017-01-16
Add a text box in Preferences to choose an alternative search engine


PalmerAL:

PalmerAL:
Added in c90549c270a22a2ca2537935c9756f3a56327c79.

PR #523, opened 2018-03-26
Added Quadrazid to searchengines

vrkttr:


daseuter:
Great Job, a lot of potential, Love It

daseuter:
Gute Ergänzung fürs Konzept.
I'd like

PalmerAL:
I don't think it makes sense to add every possible search engine to the list, since it would quickly get really long and difficult to navigate. There's an issue open to add support for custom search URLs (#303); I'll try to implement that in the next version. I also like your suggestion of adding opensearch support, although that would likely be more complicated than just adding a place to enter a URL.

dave-4k:
You add Baidu and Wikipedia but doesnt want to add the only non commercial and anonymous search engine? That confuses me.
Opensearch will be great, but Quadrazid as preselectable is an must have.

I can also imagine that the operators of quadrazid would be grateful and would advertise for the minbrowse.

Just my 2 cents ;)

PalmerAL:
It's not about the features of the search engine; it's about the number of users it has. Baidu and Wikipedia are both extremely popular (Baidu is around <a href="https://www.quantcast.com/measure/baidu.com?country=US">1.5 million people/month</a>, Wikipedia is around <a href="https://www.quantcast.com/measure/wikipedia.org?country=US">181 million</a>). As far as I can tell, Quadrazid isn't particularly popular at the moment, although it's possible that there's something I'm missing. If that changes in the future, I'd be happy to reconsider.

I think adding support for custom search engines should solve this, because it will allow people to choose whatever search engine they want. If everything goes according to plan, that should be implemented in the next 1-2 months.

PR #525, opened 2018-03-30
implement user scripts support

### LINKED ISSUE #117: Userscripts/styles
I understand not wanting to implement extensions, as that sounds very difficult. As an alternative, I suggest having some way to load custom scripts on domains matching a regex of the user's choosing.

PalmerAL:
Fixes #117


localization/languages/cs.json
```diff
@@ -101,7 +101,7 @@
         "settingsAppearanceHeading": "Vzhled",
         "settingsDarkModeToggle": "Použít tmavý režim",
         "settingsHistoryButtonToggle": "Zobrazit tlačítko zpět",
-        "settingsSwipeNavigationHeading": "Swipe navigace",
+        "settingsadditionalFeaturesHeading": null, //missing translation
```
revolter: Shouldn't this be `settingsAdditionalFeaturesHeading` instead?

PalmerAL: Yup, fixed in 8c0d405dac066e941f87f5865de6a20360e214aa.

PR #528, opened 2018-03-30
Add alert for multiple running instances of Min

### LINKED ISSUE #526: There should be a warning if another instance is already open
Cloning this project, following the instructions in the README and opening the apps throws these errors in the console:

```
database.js:81 database error occured DOMException: Internal error opening backing store for indexedDB.open.
(anonymous) @ database.js:81
database.js:81 database error occured Error: Database not open. Following error in populate, ready or upgrade function made Dexie.open() fail: UnknownError occurred when opening database browsingData. Internal error opening backing store for indexedDB.open.
    at Dexie.min.js:1
    at k (Dexie.min.js:1)
    at new t (Dexie.min.js:1)
    at Dexie.min.js:1
    at Function.t.newPSD (Dexie.min.js:1)
    at ir._promise (Dexie.min.js:1)
    at u._transPromiseFactory (Dexie.min.js:1)
    at Object.resume (Dexie.min.js:1)
    at Dexie.min.js:1
    at Array.forEach (<anonymous>)
(anonymous) @ database.js:81
database.js:81 database error occured Error: Database not open. Following error in populate, ready or upgrade function made Dexie.open() fail: UnknownError occurred when opening database browsingData. Internal error opening backing store for indexedDB.open.
    at Dexie.min.js:1
    at k (Dexie.min.js:1)
    at new t (Dexie.min.js:1)
    at Dexie.min.js:1
    at Function.t.newPSD (Dexie.min.js:1)
    at ir._promise (Dexie.min.js:1)
    at u._transPromiseFactory (Dexie.min.js:1)
    at Object.resume (Dexie.min.js:1)
    at Dexie.min.js:1
    at Array.forEach (<anonymous>)
```

revolter:
Let me know if this makes sense or if there's a better way.

The problem would be that it doesn't show up directly on start, but most probably before a db related issue like incomplete preferences screen or unable to connect to an URL:

screen shot 2018-03-31 at 01 14 01

Fixes #526

### DIFF
--- a/js/util/database.js
+++ b/js/util/database.js
@@ -82,7 +82,7 @@ Dexie.Promise.on('error', function (error) {
   console.warn('database error occured', error)
 
   if (!dbAlertShown) {
-    window && window.alert && window.alert('Make sure no other Min instance is running and then restart Electron.')
+    window && window.alert && window.alert(l('multipleInstancesErrorMessage'))
 
     dbAlertShown = true
   }
--- a/js/util/database.js
+++ b/js/util/database.js
@@ -2,6 +2,8 @@
 // requires Dexie.min.js
 
 var db = new Dexie('browsingData')
+
+var dbErrorMessage = 'Internal error opening backing store for indexedDB.open'
 var dbErrorAlertShown = false
 
 // Min 1.1.0-1.3.1
@@ -81,7 +83,7 @@ db.open().then(function () {
 Dexie.Promise.on('error', function (error) {
   console.warn('database error occured', error)
 
-  if (!dbErrorAlertShown) {
+  if (error.message.indexOf(dbErrorMessage) !== -1 && !dbErrorAlertShown) {
     window && window.alert && window.alert(l('multipleInstancesErrorMessage'))
 
     dbErrorAlertShown = true
--- a/js/util/database.js
+++ b/js/util/database.js
@@ -82,7 +82,7 @@ Dexie.Promise.on('error', function (error) {
   console.warn('database error occured', error)
 
   if (!dbAlertShown) {
-    window.alert('Make sure no other Min instance is running and then restart Electron.')
+    window && window.alert && window.alert('Make sure no other Min instance is running and then restart Electron.')
 
     dbAlertShown = true
   }
--- a/js/util/database.js
+++ b/js/util/database.js
@@ -2,6 +2,7 @@
 // requires Dexie.min.js
 
 var db = new Dexie('browsingData')
+var dbAlertShown = false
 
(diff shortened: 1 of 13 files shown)


js/util/database.js
```diff
@@ -79,4 +80,10 @@ db.open().then(function () {
 
 Dexie.Promise.on('error', function (error) {
   console.warn('database error occured', error)
+
+  if (!dbAlertShown) {
```
PalmerAL: Instead of triggering this on every error, I think we should instead check whether or not the error message has the "Internal error opening backing store" string in it.

js/util/database.js
```diff
@@ -79,4 +80,10 @@ db.open().then(function () {
...
+    window.alert('Make sure no other Min instance is running and then restart Electron.')
```
PalmerAL: This needs an ```if (window && window.alert)``` around it, since this error could also occur from within a worker that doesn't have a ```window``` object.

Also, could you move the string to the translation file so that it can be translated into other languages? To do that, you would add a line to the ```localization/languages/en-US.json``` file, and then replace the string here with ```l(whatever the string name is)```.

PalmerAL: Also, for the string itself, it's probably better not to mention Electron, since it's also possible to start multiple instances without trying to develop it. Maybe something like "An error occurred. Please close any other open instances and restart Min"?

js/util/database.js
revolter: The problem is that this callback is called more than once with the same internal error message, so checking only the message shows the alert 6 times on a fresh open (with only the tour tab opened)

PalmerAL: OK, I didn't realize that, maybe check both the message and ```dbAlertShown```?

PalmerAL: That makes sense, I'm just wondering if there's other types of errors that can occur. If some other kind of error (unrelated to the multiple-instance issue) occurs, we probabably don't want to show the error message even if it is the first error to happen.

revolter: I see, then the whole logic will be changed. Will do it tomorrow 😃

PalmerAL: I'm not sure you would have to change much, I think you could just do this:
```
if (!dbAlertShown && error.toString().indexOf("..
```

But I haven't actually tested this either, so maybe it's more complicated than that.

revolter: I put the check at the beginning because if another error would be triggered, the flag would be set to `true` and the error message would never be checked, because the `!dbAlertShown` would be `false` so there would be no reason to evaluate the second part of the `and` operation.

PalmerAL:
Everything looks good, thanks!

PR #530, opened 2018-03-31
Make sure the files build on watch start too

revolter:
This removes the need for running "grunt" before "grunt watch:scripts"
the first time.


PalmerAL:
I never realized there was an option for that, thanks for adding it!

revolter:
@PalmerAL, Neither did I 😆 But it was annoying and looked for a solution.

Issue #89, opened 2016-04-21
Turn off zooming with "CMD + <scroll>"


Vilscon:
I use the command tab a lot, but I also scroll a lot. Whenever I just scrolled a bunch and press CMD + TAB to move to a different window, for example, the whole page gets zoomed in.

This should either be turned off or they should be a way to capture the difference between these moments. I mean, even if the page is stopped scrolling, the scrolling momentum is staying alive and will zoom the page when you press the CMD button.

abejfehr:
I now see why this is happening, but I can't think of a good fix.

The [`wheel`](https://github.com/minbrowser/min/blob/d1db4753aff7f4a8c038c1ef54371714e95a6cc9/js/webview/swipeEvents.js#L38) event being captured seems to fire repeatedly on macOS until the momentum from the scroll is gone, which can be seen [here](https://jsfiddle.net/38u9yjdL/).

I thought that a valid alternative would be to use the `scroll` event, but that wouldn't be accurate either, since even if the window isn't scrolling, some element inside of it could be. I've illustrated that [here](https://jsfiddle.net/38u9yjdL/1/).

Does anyone have any ideas?

revolter:
A very easy solution would be to add a preference setting that disables it, as you can also use <kbd>cmd</kbd> + <kbd>+</kbd>.

PalmerAL:
It looks like electron has a <a href="http://electronjs.org/docs/api/browser-window#event-scroll-touch-end-macos">```scroll-touch-end```</a> event that could work. The only issue would be that it fires in the main process, so we would be sending a lot of IPC messages to the window, which might not perform very well.

PalmerAL:
Another way we could fix this would be to check whether the command key is pressed at the beginning of the scroll gesture (we could probably reuse the swipeGestureTimeout for this). If the key isn't pressed at the beginning of the gesture, then we wouldn't zoom even if it is pressed later on in the gesture.

revolter:
Sounds good. Will you start working on it? I could test it if you want.

PalmerAL:
I can do it if you want, I don't think it will be too difficult.

PalmerAL:
@revolter PR is here: #534.

PR #531, opened 2018-03-31
Fix incorrect scrolling on mod + scroll zoom

### LINKED ISSUE #127: Pinch to Zoom
The current zooming solution (cmd + scroll) is clunky and scrolls while zooming. Implementing pinch to zoom would make it much more intuitive for users of apple laptops.

revolter:
Fixes #127

### DIFF
--- a/js/webview/swipeEvents.js
+++ b/js/webview/swipeEvents.js
@@ -79,6 +79,8 @@ window.addEventListener('wheel', function (e) {
         zoomIn()
         verticalMouseMove = 0
       }
+
+      e.preventDefault()
     }
     platformZoomKey = e.metaKey
   }
@@ -92,5 +94,9 @@ window.addEventListener('wheel', function (e) {
   if (verticalMouseMove < -55 && platformZoomKey) {
     verticalMouseMove = -10
     zoomIn()
+  }
+
+  if (platformZoomKey) {
+    e.preventDefault()
   }
 })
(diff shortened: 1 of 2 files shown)


js/webview/swipeEvents.js
```diff
@@ -8,6 +8,9 @@ var beginningScrollRight = null
 
 var hasShownSwipeArrow  = false
 
+// TODO: find out a way to read this from settings
+var isZoomWithScrollEnabled = true
```
revolter: @PalmerAL, Could you please help me figure out how to read the respective setting's value in this file?

PalmerAL: @revolter You can't, because you can only read from the database in a page with a file:/// origin. The way we worked around this for the swipe preference was to send an IPC to the parent process, which would then check the preference and do the navigation. (the code for this is here: https://github.com/minbrowser/min/blob/master/js/webviews.js#L272-L278). I'm guessing you'd have to do something similar here.

PalmerAL:
I guess we could add this if we really have to, but I'd prefer to find a way to fix the original issue instead. If we just add a preference, most people aren't even going to notice it, and if you turn it on you lose the functionality entirely. Is there some other way we could detect whether the page is actually being scrolled?

revolter:
@PalmerAL, I rebased the entire branch and only left the minor spelling fix, grunt task simplifier ❤️ and the #127 fix.

PalmerAL:
@revolter Looks good, thanks! I think I have an idea about how to fix #89, I'll add a comment to that issue.

PR #534, opened 2018-04-01
only zoom if the zoom key was pressed at the beginning of the gesture

### LINKED ISSUE #89: Turn off zooming with "CMD + <scroll>"

PalmerAL:
Fixes #89


js/webview/swipeEvents.js
```diff
@@ -43,13 +47,20 @@ window.addEventListener('wheel', function (e) {
...
+  var platformZoomKey = ((navigator.platform === 'MacIntel') ? e.metaKey : e.ctrlKey)
+
   if (!beginningScrollLeft || !beginningScrollRight) {
     beginningScrollLeft = document.scrollingElement.scrollLeft
     beginningScrollRight = document.scrollingElement.scrollWidth - document.scrollingElement.clientWidth - document.scrollingElement.scrollLeft
   }
 
   if (Math.abs(e.deltaX) >= 20 || Math.abs(e.deltaY) >= 20) {
     clearTimeout(swipeGestureTimeout)
+    swipeGestureTimeout = setTimeout(onSwipeGestureFinish, 70)
```
revolter: Didn't test it yet, but, why `70`? Looks like a magic number

js/webview/swipeEvents.js
```diff
@@ -82,21 +88,20 @@ window.addEventListener('wheel', function (e) {
...
-  if (verticalMouseMove < -55 && platformZoomKey) {
+  if (verticalMouseMove < -55 && platformZoomKey && initialZoomKeyState) {
     verticalMouseMove = -10
     zoomIn()
   }
 
-  if (platformZoomKey) {
+  if (platformZoomKey && initialZoomKeyState) {
     e.preventDefault()
   }
```
revolter: Not directly related to this PR, but these could be rewritten as:

```js
if (platformZoomKey && initialZoomKeyState) {
  if (verticalMouseMove > 55) {
    verticalMouseMove = -10
    zoomOut()
  }

  if (verticalMouseMove < -55) {
    verticalMouseMove = -10
    zoomIn()
  }

  e.preventDefault()
}
```

so `platformZoomKey && initialZoomKeyState` wouldn't be duplicated anymore.

PalmerAL: Fixed in 31088da3bd5227cd4bab0358751589c9c64aae39.

js/webview/swipeEvents.js
PalmerAL: Yeah, pretty much. Increasing the number means there's more time to finish the gesture (which can be more accurate), but also means that there's more of a delay between when you finish swiping and when the navigation actually happens. 70 seemed to work pretty well when I tested it.

revolter: I'm not a big fan of magic numbers, but when you really need them, you could either create a constant (or a simple variable) so it has a name (if possible), or add a comment for future readers, like:

```
/* The timeout duration represents the time between the swipe gesture's start and end. Increasing the number means there's more time to finish the gesture (which can be more accurate), but also means that there's more of a delay between when you finish swiping and when the navigation actually happens. 70 seemed to work pretty well. */
```

or how you see fit.

revolter:
I forgot to run `npm run buildTranslations`. We should automate that somehow.

revolter:
It's not working for me for some reason :/

PalmerAL:
@revolter You have to run `grunt` again after ```buildTranslations```. There's probably a way to run an NPM task from grunt, I'll have to look into that.

js/webview/swipeEvents.js
PalmerAL: Good idea about the comment, fixed in e4240d3981658ffea459283dc0c69cc22aba4d43. I would agree that magic numbers in  general aren't great, but I'm not sure how you would avoid it here, since there's no objectively right amount of time to wait.

revolter:
I have already done it but was waiting for you to merge those PRs to avoid conflicts.

Regarding this PR, does it work for you on macOS?

PalmerAL:
@revolter It works for me, is it not workign for you?

revolter:
No. I opened a long webpage, scrolled down fast, then hit <kbd>cmd</kbd> + <kbd>tab</kbd> to swith to the next tab and that tab's webpage got zoomed out.

PalmerAL:
@revolter Oh, now I see what you're talking about. The problem is that the second tab doesn't start recieving `wheel` events until you switch to it, and at that point the ctrl/cmd key is pressed. I'm not sure how to fix this.

code-hunger:
> I'm not sure how to fix this.

I see two ways:
1. A global flag indicating whether Ctrl was pressed.
 * On ctrl press, set it true, unless already wheel-ing.
 * On wheel start check it to decide whether to zoom/scroll.
 * On wheel end set it to false. This way, when you switch to another tab while ctrl is being pressed and the endWheel event is fired, the ctrl flag will be false. The next tab will receive wheelStart but ctrl will be false and it won't zoom.
2. Take the job of zooming out of the tab and make the whole winodw responsible for it. This way changing tabs won't disturb the scroll/zoom logic, because it will be separated from the tabs.

PalmerAL:
Sorry for the late response, I've been pretty busy this week. I think option 2 makes the most sense; that should eventually allow us to remember zoom levels for websites as well. I'll start working on that.

PalmerAL:
@revolter @code-hunger I've moved the gesture code to the window in 14cc294784bce4a39fba42f2e55a86f39b4535b4, so this should actually work now. Would you mind trying it out and seeing if it works for you?

revolter:
Looking great! The only "problem" now is that the scroll continues on the new tab, but it's not even 1% as annoying as the initial zoom problem.

code-hunger:
First impression - I had to wait a few minutes for grunt to download electron and "package the app". I don't know what happened but now it's downloading electron for the third time.

Finally grunt gave up and exited; here's it's output after `Running "concat:main" (concat) task`:

```
Running "electron:linuxBuild" (electron) task                                                                                                        
WARNING: The version parameter is deprecated, use electronVersion (or --electron-version in the CLI) instead                                         
WARNING: The app-version parameter is deprecated when used from JS, use appVersion instead. It will be removed in the next major version.            
Downloading SHASUMS256.txt
Downloading electron-v2.0.0-beta.4-linux-x64.zip
[============================================>] 100.0% of 48.56 MB (2.49 MB/s)
Downloading electron-v2.0.0-beta.4-linux-armv7l.zip
[============================================>] 100.0% of 45.5 MB (2.72 MB/s)
Packaging app for platform linux armv7l using electron v2.0.0-beta.4                                                                                 
                                                                                                                                                     
                                                                                                                                                     
Running "electron-installer-debian:linux32" (electron-installer-debian) task                                                                         
Creating package (this may take a while)                                                                                                             
>> Error creating package: Error executing command (127):                                                                                            
>> fakeroot dpkg-deb --build /tmp/electron-11841-10008-1rdhk69.rks5/min_1.7.1_i386
>> /usr/bin/fakeroot: line 178: dpkg-deb: command not found
Warning: Error executing command (127): 
fakeroot dpkg-deb --build /tmp/electron-11841-10008-1rdhk69.rks5/min_1.7.1_i386
/usr/bin/fakeroot: line 178: dpkg-deb: command not found
 Use --force to continue.

Aborted due to warnings.
```
(I hit Enter after the second download so that the terminal correctly shows the third download; otherwise it would have overwritten the two animated "download progress" lines)

It makes no sense to download electron. I have it installed locally. So far `grunt` has always finished for about 10 seconds and then I told electron to start min. Now it tried to package itself...

How to build it without this "packaging app" stuff?

revolter:
@code-hunger, I'm pretty sure `grunt` doesn't download `electron`. That's `npm`'s job. What command are you running? It looks to me that you ran `grunt linuxBuild`.

PalmerAL:
@code-hunger `grunt linuxBuild` builds a .deb package that you can distribute; if you want to run from source, you would run `grunt` and then `/path/to/electron .`.

The re-downloading of electron is controlled by electron-packager, I don't think theres anything we can do anything to change that. In most cases, re-downloading makes sense, because you'd want to make sure you have the correct version of electron from package.json, and because you need binaries for all three CPU types (32/64-bit x86 and ARM). I think for the error message you need to install dpkg, but I'm not sure. Regardless, none of this should matter unless you want to create a package anyway.

revolter:
@code-hunger Glad to know you successfully built it 😃.

PalmerAL:
I'm glad you figured it out! Since this seems to work for everyone, I'm going to merge it; please open an issue if you notice any problems with it.

The scrolling problem is kind of annoying, but we don't do anything to modify the built-in scrolling, so I'm not sure what we can do about it.

PR #537, opened 2018-04-02
Automate npm scripts using Grunt

revolter:


PalmerAL:
The translation changes look good. I'm not sure auto-updating EasyList is a good idea; in the past, there's been instances where bugs in our filter parser caused filtering to be broken with newer versions of EasyList, so it's good to test updates for a while before a release.

revolter:
@PalmerAL, Done! All should be good now ❤️

PalmerAL:
Looks good, thanks!

PR #553, opened 2018-05-05
Error pages add input

code-hunger:
After a long fight with git I finally managed to open this PR without conflicts, but I still have no idea why my commits from last year still show up, even though they're submitted to the repository. I tried rebasing onto upstream/master but after fixing a few conflicting files I gave up, seeing it would take too much time to fix everything that has been updated since then.

I've merget `upstream/master` to this branch but still in 'Files changed' I see some changes that I haven't done recently at all - those under `js/`. If I run `git diff master` I see only the files under `pages/` (those I edited). If I run `git diff upstream/master` then I get the files under `js/`, too. I have no idea why this happens.

Aaand - finally coming to the point, here's a PR introducing an input field on the error pages so that the user can fix the url (which is mangled (urlencoded) in the searchbar & it's un-editable).


code-hunger:
Yes, after my latest successful pull & build I got it working. Sorry for the extra PR, perhaps I've forgotten git & missused it after a long break. I hope it won't happen again! (But I'm not sure what to do about all the extra commits... merging upstream says 'up to date'. Maybe a full rebase is required...)

PalmerAL:
I'm not sure, I'm not too great at using git either.

PR #556, opened 2018-05-14
Various back/forward related improvements & Updating Japanese localization

LEoREo2247:


js/webviews.js
```diff
@@ -23,6 +23,23 @@ function onPageLoad (e) {
...
+      goBackButton.classList.remove('fa-angle-left')
+      goBackButton.classList.add('fa-angle-right')
+      goBackButton.style.opacity = '1'
+      goBackButton.title = l('goForward')
+    } else {
+      goBackButton.classList.remove('fa-angle-right')
+      goBackButton.classList.add('fa-angle-left')
+      goBackButton.style.opacity = '0.5'
+      goBackButton.title = l('goBack')
+    }
```
code-hunger: Avoid repetition by moving all the `classList` modifications out of the if-clauses and use `toggle`:

```
goBackButton.classList.toggle('fa-angle-left')
goBackButton.classList.toggle('fa-angle-right')
```

Also, this code seems suspiciously copy-pasted from tabBar. Move it to a common function to keep common logic to avoid the need to support one code on two places.

js/webviewGestures.js
```diff
@@ -128,10 +128,14 @@ window.addEventListener('wheel', function (e) {
...
     if (horizontalMouseMove < -150 && Math.abs(horizontalMouseMove / verticalMouseMove) > 2.5 && !hasShownSwipeArrow) {
       hasShownSwipeArrow = true
-      webviewGestures.showBackArrow()
+      if (webviews.get(tabs.getSelected()).canGoBack()) {
+        webviewGestures.showBackArrow()
+      }
     } else if (horizontalMouseMove > 150 && Math.abs(horizontalMouseMove / verticalMouseMove) > 2.5 && !hasShownSwipeArrow) {
       hasShownSwipeArrow = true
-      webviewGestures.showForwardArrow()
+      if (webviews.get(tabs.getSelected()).canGoForward()) {
```
code-hunger: The `wheel` event callback calls `webviews.get(tabs.getSelected())` three times. You can make it cleaner by putting it into a variable in the beginning and avoding so many identical calls.

js/navbar/goBackButton.js
```diff
@@ -1,11 +1,76 @@
...
+	},500)
+})
+
+function historyItemClicked(index) {
+  var currentIndex = webviews.get(tabs.getSelected()).getWebContents().getActiveIndex()
+
+  if (webviews.get(tabs.getSelected()).canGoBack()) {
+    webviews.get(tabs.getSelected()).goToIndex(currentIndex - index)
+  } else {
+    webviews.get(tabs.getSelected()).goToIndex(currentIndex + index)
```
code-hunger: As I commented in the other file, no need to call `webviews.get(tabs.getSelected())` 3 (+1) times in a single function. Put it in a variable in the beginning. Then you'll be able to make it much shorter, for example:

```
var currInd = X.method(),
    newInd = X.canGoBack() ? currInd + index : currInd - index;
X.goToIndex(newInd);
```

PalmerAL:
Thanks for adding this! The translation and the swipe gesture changes look OK. For the back button, there are couple of things that I think could be changed:

* I'm not sure I really like the behavior of showing the forward arrow when you can't go back. Instead of just being able to click the button and know that I'll go back, now I have to look at the icon every time and figure out what's going to happen. I think it's better to just leave the existing controls for going forward that we have now.
* The code to update the back button appearance is duplicated twice, and it still doesn't work in all scenarios (such as switching tabs using a keyboard shortcut). I think a better way to do this would be to have an ```updateBackButton``` function in backButton.js, and then call that directly from ```switchToTab```. That way, we can catch every scenario where the button would change, and only have one copy of the code.
* Rather than trying to create our own menu, I think the navigation history dropdown would look better as a [native menu](https://electronjs.org/docs/api/menu). Doing this would also make the code a lot simpler, because then you wouldn't need a lot of the logic for setting the menu appearance, deciding when to close the menu, etc. ```webviewMenu.js``` has an example of how this could be implemented.

js/navbar/goBackButton.js
```diff
@@ -1,11 +1,77 @@
...
+    navigationHistoryOpened = true
+	},500)
+})
+
+function historyItemClicked(index) {
+  var webview = webviews.get(tabs.getSelected())
+
+  var currentIndex = webview.getWebContents().getActiveIndex()
+  var delta = webview.canGoBack() ? -index : index
+  console.log(delta)
```
PalmerAL: Can you remove this?

js/webviewGestures.js
```diff
@@ -128,10 +130,14 @@ window.addEventListener('wheel', function (e) {
 
     if (horizontalMouseMove < -150 && Math.abs(horizontalMouseMove / verticalMouseMove) > 2.5 && !hasShownSwipeArrow) {
       hasShownSwipeArrow = true
```
PalmerAL: ```hasShownSwipeArrow``` would need to be moved inside the if(canGoBack) statement, both here and in the section below.

PalmerAL:
The webview and back button code has changed enough that this can't be merged as-is, but if you want to continue working on it, feel free to open a new PR.

PR #557, opened 2018-05-15
New build system

PalmerAL:
This implements a new build system for the browser using Browserify. All of the existing scripts are concatenated together, and then the result is run through browserify to bundle all of the modules. Over time, we would gradually convert the items in ```legacyModules```  to real modules that are loaded using ```require()```, and when everything is converted we could eventually just run browserify directly on ```default.js```.

@code-hunger I think this addresses what you mentioned regarding implementing features as modules. Does this approach make sense?

Once we decide on what the right approach is, I'll implement something similar for the webview bundle as well, and remove the old grunt-based system.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -8,104 +8,6 @@ module.exports = function (grunt) {
   // Project configuration.
   grunt.initConfig({
     pkg: grunt.file.readJSON('package.json'),
-    concat: {
-      options: {
-        separator: '\n;'
-      },
-      browser: {
-        src: [
-          'dist/localization.build.js',
-          'js/default.js',
-          'js/windowsCaptionButtons.js',
-          'js/util/database.js',
-          'js/util/defaultKeyMap.js',
-          'js/util/settings.js',
-          'js/util/searchEngine.js',
-          'js/util/relativeDate.js',
-          'js/menuBarVisibility.js',
-          'js/tabState.js',
-          'js/util/urlParser.js',
-          'js/filteringRenderer.js',
-          'js/webviews.js',
-          'js/phishingWarning.js',
-          'js/webviewMenu.js',
-          'js/bookmarksHistory/bookmarksHistory.js',
-          'js/api-wrapper.js',
-          'js/searchbar/searchbar.js',
-          'js/searchbar/searchbar-plugins.js',
-          'js/searchbar/searchbar-autocomplete.js',
-          'js/searchbar/placesPlugin.js',
-          'js/searchbar/instantAnswerPlugin.js',
-          'js/searchbar/openTabsPlugin.js',
-          'js/searchbar/bangsPlugin.js',
-          'js/searchbar/customBangs.js',
-          'js/searchbar/searchSuggestionsPlugin.js',
-          'js/searchbar/placeSuggestionsPlugin.js',
-          'js/searchbar/hostsSuggestionsPlugin.js',
-          'js/readerview.js',
-          'js/navbar/tabActivity.js',
-          'js/navbar/tabColor.js',
-          'js/navbar/tabBar.js',
-          'js/taskOverlay/taskOverlay.js',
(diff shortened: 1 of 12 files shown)


PalmerAL:
Right, that part is the same as what Grunt does; the part that's new is the browserify part (lines 66-74). Basically, we eventually want to get to the point where we don't have any global variables, and everything is split into modules. For example, instead of having a ```tabs``` global variable, you would do ```const tabs = require("./tabState.js")``` everywhere you want to use the tabs API. The problem is that we have so many global variables that it's not possible to convert everything all at once. So this script does two things:

* First, it combines all of the global scripts we already have together, like what Grunt did.
* Then, it takes that result and runs it through Browserify, which looks for require() calls and copies that code into the file.

Over time, we would convert the files that are being concatenated together into modules, so that eventually the concatenation is gone entirely, and we end up with just the Browserify part.

I believe there is a Grunt plugin for Browserify as well, but if all we need to do to build the script is to run Browserify, it doesn't seem like we really need Grunt for anything.

code-hunger:
Ok, so your plan is to remove grunt as a dependency because it only serves as a middleware for a single command.

Another question - will this browserify allow us finally use ES6 syntax? Arrow functions, destructuring & all other fancy syntastic sugar?

PalmerAL:
ES6 isn't really related to this, that's because we're using UglifyJS, which doesn't support minifying it. Originally, I assumed that minifying the JS would reduce the time needed to load and parse the script, but it doesn't seem like that's the case, so we could probably just get rid of it.

PalmerAL:
> Ok, so your plan is to remove grunt as a dependency because it only serves as a middleware for a single command.

Basically, yes. We would also need to do something similar for the packaging components, but I wanted to get feedback on this first.

PalmerAL:
I think everything should work at this point. The JS is all built using custom scripts instead of grunt:

* `npm run buildBrowser` builds the main JS file.
* `npm run buildMain` builds the main process file.
* `npm run buildPreload` builds the preload script.
(There's also an `npm run build` command that builds everything).

Browserify is used for the browser JS (so we can start using modules there); for everything else, the files are just concatenated. If the main process or preload scripts get large enough that we want to use modules there as well, it wouldn't be too hard to enable browserify for those also.

I left the packaging part in grunt for now; I'll probably convert that at some point after the 1.8 release when I have more time to test that everything works (or if someone else is interested in working on this, that would be great as well).

@code-hunger Is there anything else you think I should change before merging this?

PalmerAL:
I'm going to go ahead and merge this so I can start converting everything to modules; if anyone has additional feedback about how this should work, please open an issue about it.

code-hunger:
Sorry for the late reply, I didn't notice you @tagged me because I jumped directly to the commit changes from the notification.

The only thing I'd note is the repetition of the concatenation code, but I guess we can live with that too.

Another question though - you say _Browserify is used for the browser JS (so we can start using modules there); for everything else, the files are just concatenated._. This means, only the result from buildBrowser is browserified (i.e. resolution of `require`), but in `main.js` I see `require`s, too... 

Doesn't matter too much, just curious how it works. So, now we're ready to start moving `legacyModules` to exported & required modules, right?

PalmerAL:
> The only thing I'd note is that there's some repetition of the concatenation code, but I guess we can live with that too.

We probably should fix that eventually, but I'm not going to worry about it too much at this point, especially since one of those instances should be going away in a bit.

> Another question though - you say Browserify is used for the browser JS (so we can start using modules there); for everything else, the files are just concatenated.. This means, only the result from buildBrowser is browserified (i.e. resolution of require), but in main.js I see requires, too...

Any require() call that doesn't go through browserify (or is something that browserify can't transform, like `require("electron")`) will just use the require() function that is built-in to node, and load the module at runtime.

> So, now we're ready to start moving legacyModules to exported & required modules, right?

Yup!

PR #578, opened 2018-07-09
updated missing translations

aliencoder16:
Updated missing translation lines, and fixed a typo.


aliencoder16:
Hi @PalmerAL , I thought the strings that have //missing translation are the only ones

PalmerAL:
They aren't, because I haven't been very consistent about adding a //missing translation comment to the other files when adding new strings. I'll try to fix that so it's easier to see which ones need to be added.

PalmerAL:
I've added comments for all of the strings that weren't listed in 0097a1b4ebc2143bcd8c5e2bf34f0669ec71fdf2; in the future, I'll make sure we do this when we add new strings to the UI.

PR #587, opened 2018-07-27
Use BrowserView instead of webview

PalmerAL:
Webviews are currently broken in Electron 3.0 because of electron/electron#13712. I'm hoping that we can switch back to Webviews eventually, but unless this gets fixed soon, we'll need to start using BrowserViews so that we can upgrade to a newer version of Chromium.

I'm expecting that a lot of things will be broken because of this (and there are still some features that haven't been re-enabled), so please try this out, and add a comment if you notice any issues!


PalmerAL:
I think everything works now, so I'm going to merge this.

PR #588, opened 2018-08-01
Corrected Simplified Chinese and added Traditional Chinese language

slemony:
corrected grammar mistake and complete all of the translation of simplified chinese language and added traditional chinese language

### DIFF
--- a/localization/languages/zh-CN.json
+++ b/localization/languages/zh-CN.json
@@ -1,4 +1,4 @@
-{
+﻿{
     "name": "中文 (简体中文)",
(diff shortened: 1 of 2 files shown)


PalmerAL:
Thanks! The only thing that I think needs to be changed is the "identifier" field in zh-TW.json; it should be "zh-TW" to match the filename. Otherwise, everything looks good!

slemony:
Thanks for remind! I've change the identifier.

PalmerAL:
Thanks!

PR #595, opened 2018-08-14
Disable back button when it's not active

revolter:
This makes the back button transparent when you can't go back because you're on the first page.

### DIFF
--- a/css/tabBar.css
+++ b/css/tabBar.css
@@ -72,6 +72,9 @@ body.windows #navbar {
 	font-size: 1.5em;
     padding: 0.8em 0.6em 0.8em 0.4em;
 }
+#back-button:disabled {
+	opacity: 0.2;
+}
 
 body.is-edit-mode #back-button,
 body.is-edit-mode #menu-button {
--- a/js/webviews.js
+++ b/js/webviews.js
@@ -68,7 +68,7 @@ function captureCurrentTab () {
 function updateBackButton () {
   const webview = webviews.get(tabs.getSelected())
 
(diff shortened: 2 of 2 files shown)


revolter:
It looks like refreshing a tab causes the `canGoBack` method to return `true` instead of `false, and you can indeed go "back" to the same page. But I think this is another issue and not related to this PR.

PalmerAL:
Thanks! A couple of suggestions:

* Since this function is called more often than just on navigations, maybe call it something like `updateBackButton`?
* Instead of directly setting the opacity, can we instead add and remove the `disabled` attribute of the button, and then have CSS that changes the opacity? That shoul dbe better for accessibility, and would also allow us to reuse the same styling for the other buttons if we want to.

I'm guessing that if this function is called from `did-navigate`, it's redundant to also call it from `onPageLoad`, but I haven't actually tested this, so feel free to leave it the way it is if it's necessary.

PalmerAL:
I've opened a new issue about the reloading issue: #596.

revolter:
Will update the PR tonight with those 2 changes.

> I'm guessing that if this function is called from `did-navigate`, it's redundant to also call it from `onPageLoad`, but I haven't actually tested this, so feel free to leave it the way it is if it's necessary.

For some reason, when opening up the browser, `did-navigate` is not called, so the back button doesn't get updated.

revolter:
@PalmerAL, Updated 😃

PalmerAL:
LGTM! I think I know what's causing the reloading issue, I'll try to fix that soon.

PR #597, opened 2018-08-20
Add Bangla translations

abrarShariar:


PalmerAL:
Thanks for doing this! A couple of things I noticed:

* There are several places in the translations where we leave placeholders to insert other strings, like %i, %s, %n, etc. It looks like you've added spaces between the % and the letter, or changed the letter to uppercase for some of these, which will prevent the replacement from working - can you change that back?
* In a couple places (like `sslErrorMessage`), it looks like you've translated "Min" to a word that means "minimum" (although I'm reviewing this using Google translate, so I could be completely wrong about that). Since this is referring to the name of the browser, and we typically leave that in English in our other translations, I think it might make more sense to leave the name in English here as well?

Other than that, everything looks good!

abrarShariar:
hey @PalmerAL fixed the issues and pushed code. Please let me know if there are other issues. Thanks!

PalmerAL:
Everything looks good to me!

PR #603, opened 2018-08-28
fix zh-tw translation

z-huang:
I think zh-TW.json is a copy from zh-CN, just changed the simplified words to the traditional words, so I modified to the 'real' Traditional Chinese and fix the grammar.
Sorry for my poor English!


PalmerAL:
Looks good to me, thanks!

PR #612, opened 2018-09-09
Turn tabState.js into 2 smaller module parts

code-hunger:
Took the "tabPrototype" and the "taskPrototype" out of tabState.js and turned them to modules.

I think it works fine. I started making the TaskOverlayBuilder a module in the meantime, and if there had been some error here, I've probably fixed it already in the next commits, but I think the tabState.js must be bug-free. Let me know if it's not!


js/webviews.js
```diff
@@ -430,7 +435,7 @@ ipc.on('view-ipc', function (e, data) {
 
 setInterval(function () {
   captureCurrentTab()
-}, 30000)
+}, 3000)
```
code-hunger: Ops... I forgot that. Will fix it shortly, but in the meantime, what is it used for?

PalmerAL:
I think this makes sense; I'll look at it more closely when I have time tomorrow. I don't think we actually need the error checks - can you remove those? (and the package-lock.json changes should also probably be separate).

js/webviews.js
PalmerAL: Because we can't draw anything directly on top of browserviews, there's no way to show the page in the background while the searchbar is open (or during the task overlay transition). To work around this, we capture a low-resolution image of the page. When the searchbar is opened, the view gets hidden, and the image is shown instead. We normally make a new image every time a new page is loaded, but in some situations (such as single-page apps, or if you scroll to a different part of the page) it's possible for the content of the page to change without the URL changing. This code makes a new image at 30-second intervals, so that the image is always at least somewhat close to what the actual page looks like, even if the URL hasn't changed in a while.

(I should probably add a comment explaining this to webviews.js; I'll probably do that after this PR is merged).

code-hunger: Oh.. so it's doing a screenshot. I had interpretated "capture" as "grab keyboard and mouse" and was wondering why it was here.

code-hunger:
Undid package-lock and error changes and pushed a few unpushed commits from before. I've read package-lock.json is to be commited so I'd committed it, but if we want to stay clear I guess it's ok.

PalmerAL:
Looks good to me. I'm not sure I like using if statements without braces, because it makes it more difficult to tell what code is part of the if statement and what isn't, but we can fix that later.

I'm not exactly sure about the package-lock.json thing - you're apparently supposed to commit it, but it changes every time you do `npm install` anyway, so I'm not really sure what the point is. (There probably is a reason; I need to go read the NPM docs about it at some point.)

PR #615, opened 2018-09-11
Remove settings.js from the legacy modules list and fix imports

code-hunger:
`settings.js` no more needs to be concatenated with the others. I used the same technique from settings.js to check if the page is loaded in the browser from an html file or not, in order to decide whether to import settings.js or not. Otherwise one part of the browser was giving an error about `settings` being undefined.

### DIFF
--- a/js/keybindings.js
+++ b/js/keybindings.js
@@ -54,6 +54,27 @@ ipc.on('showHistory', function () {
   tabBar.enterEditMode(tabs.getSelected(), '!history ')
 })
 
+ipc.on('duplicateTab', function (e) {
+  /* new tabs can't be created in focus mode */
+  if (focusMode.enabled()) {
+    focusMode.warn()
+    return
+  }
+
+  // can't duplicate if tabs is empty 
+  if (tabs.isEmpty()) {
+    return
+  }
+
+  const newIndex = tabs.getIndex(tabs.getSelected()) + 1
+
+  const sourceTab = tabs.get(tabs.getSelected())
+  // strip tab id so that a new one is generated
+  const newTab = tabs.add({...sourceTab, id: undefined}, newIndex)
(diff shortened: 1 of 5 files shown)


code-hunger:
Actually I just thought about a way to allow the usage of const instead of var by removing searchEngine from the list, too. I don't have the time to do it now so I'll check it in a few hours.

code-hunger:
For now it'll be too hard to fix that alone - we need to isolate a few other components before we're able to do it cleanly. I've already started working on it on another branch to keep it simple and clear, but it's far from ready. Let's proceed it piece by piece :)

And btw, doing this refactoring & investigation of browser internals makes me udnerstand how it works much more deeply. Quite an interesting task :)

code-hunger:
I just saw there was a request for a tab duplicate feature, decided to give it a shot here.

PalmerAL:
In order to do this correctly, we'd also need to import `settings.js` from all the other places that use it - filteringRenderer.js, keybindings.js, menubarVisibility.js, goBackButton.js, phishingWarning.js, and probably some other ones. For now, this happens to work, because searchEngine.js is loaded in the global scope, and so once it loads `settings.js`, all the other code in the global scope will be able to access it as well, but this won't work once `searchEngine.js` is itself a module. The way I've been solving this for the other modules is to just do `var settings = require("settings.js")` at the top of each global script, and let them all overwrite each other. Browserify will bundle the scripts so that there is still only one instance of `settings`, and this way we can reduce the amount of breakage that will happen later.

PalmerAL:
I think the duplicate tab addition looks good for the most part; could you split that into a separate PR?

PalmerAL:
Settings.js was converted to a module as part of #766, so this isn't necessary any more.

PR #618, opened 2018-09-13
TaskOverlayBuilder & UrlParser taken out of legacy modules

code-hunger:
Removes the need to set a global function task_container_build_func, gave it a better name.

Also, I noticed you have turned the UrlParser to a module, but haven't removed it from the legacy modules list.

Btw, I understood why in previous PRs older commits appeared in the diff. It's because you squash the commits from my PR, and when I make new changes, my previous commits are still in my repo, but not on yours. When I started making this PR, github showed the changes from the entire previous PR. I had to rebase this branch onto your master to clean it. (And the effect is that they now appear to be made in the exact same second)

### DIFF
--- a/js/taskOverlay/taskOverlay.js
+++ b/js/taskOverlay/taskOverlay.js
@@ -1,6 +1,8 @@
 var browserUI = require('api-wrapper.js')
 var focusMode = require('focusMode.js')
 
+const createTaskContainer = require("taskOverlay/taskOverlayBuilder.js")
+
 var taskContainer = document.getElementById('task-area')
 var taskSwitcherButton = document.getElementById('switch-task-button')
 var addTaskButton = document.getElementById('add-task')
@@ -68,7 +70,7 @@ window.taskOverlay = {
 
     // show the task elements
     tasks.get().forEach(function (task, index) {
-      var el = window.task_container_build_func(task, index)
+      const el = createTaskContainer(task, index)
 
       taskContainer.appendChild(el)
       taskOverlay.tabDragula.containers.push(el.getElementsByClassName('task-tabs-container')[0])
--- a/js/taskOverlay/taskOverlayBuilder.js
+++ b/js/taskOverlay/taskOverlayBuilder.js
@@ -1,5 +1,6 @@
 var browserUI = require('api-wrapper.js')
 var searchbarUtils = require('searchbar/searchbarUtils.js')
+var urlParser = require('util/urlParser.js')
 
 function removeTabFromOverlay (tabId, task) {
   task.tabs.destroy(tabId)
@@ -154,4 +155,6 @@ var TaskOverlayBuilder = {
 // extend with other helper functions?
 }
 
-window.task_container_build_func = TaskOverlayBuilder.create.task.container.bind(TaskOverlayBuilder.create.task)
+module.exports = function createTaskContainer(task, index) {
+  return TaskOverlayBuilder.create.task.container(task, index)
+}
--- a/js/webviews.js
+++ b/js/webviews.js
@@ -1,6 +1,7 @@
 var browserUI = require('api-wrapper.js')
 const previewCache = require('previewCache.js')
 var getView = remote.getGlobal('getView')
+const urlParser = require('util/urlParser.js')
 
 /* implements selecting webviews, switching between them, and creating new ones. */
 
--- a/scripts/buildBrowser.js
+++ b/scripts/buildBrowser.js
@@ -36,7 +36,6 @@ const legacyModules = [
   'js/navbar/tabColor.js',
   'js/navbar/tabBar.js',
   'js/taskOverlay/taskOverlay.js',
-  'js/taskOverlay/taskOverlayBuilder.js',
   'js/navbar/addTabButton.js',
   'js/navbar/goBackButton.js',
   'js/navbar/menuButton.js',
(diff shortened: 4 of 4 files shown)


PalmerAL:
The task overlay changes look good! I think urlParser.js has the same issue as settings.js: not all of the files that use it import it, which will cause problems as we convert more things in the future. There's a bunch of files where we need to fix this (keybindings.js, tabBar.js, instantAnswerPlugin.js, and some other ones), but it should be fairly easy to fix. I should have time to do that in a few days, or if you want to make a PR now that's fine as well.

The diff issue is interesting; if it's really an issue I can switch to using merge commits, although the downside is that it sometimes makes the repository history harder to read. Do you have a preference?

code-hunger:
> The diff issue is interesting; if it's really an issue I can switch to using merge commits, although the downside is that it sometimes makes the repository history harder to read. Do you have a preference?

As I usually try to keep branches clean and do not do much "heterogenous" edits to the code base, the chance of a conflict introduced by you after I start and before I submit the PR are very small.  
Even in that case, conflicts shouldn't be hard to resolve, as long as there isn't much deviation of the two branches.  
So No, do not switch to fast-forward merging, as that would add to the history everything I have commited/reverted (like the package-lock.json).  

That would (should) also make devs stive to keep their PRs small and easy to review, which is better for the whole development process.

> There's a bunch of files where we need to fix this (keybindings.js, tabBar.js, instantAnswerPlugin.js, and some other ones), but it should be fairly easy to fix.

I think it'll be easier to do that incrementally, instead of at once. With the modularization of every new item in the legacy list, it'll be clear what has changed if the related imports are added in-place.

I've got 2 more branches to do PR, but I'd like to do one thing at a time and not open all PRs at once. In one of them, for example, in a place where urlParser is used, I found that the decoupling caused an error about the urlParser not defined, so I imported it there.

I hope it won't be too far away that time where we'll be able to use the linter on the whole code base to instantly find any undefined values and these problems will just disappear.

code-hunger:
BTW, look at the list of all types of errors produced by `standard`, sorted:

[sorted_errors_full.txt](https://github.com/minbrowser/min/files/2382967/sorted_errors_full.txt)

Here's the command to produce them:

    standard | sed 's/^\s*[a-z\/\.0-9:-]*\s*//i'|sort|uniq -c|sort -nr > sorted_errors.txt

An excerpt of error with >10 occurences, that is the top 15:

    204 'tabs' is not defined.                                                                                                                       
    138 'webviews' is not defined.
     64 'l' is not defined.
     53 'tabBar' is not defined.
     51 'ipc' is not defined.
     40 'tasks' is not defined.
     27 'taskOverlay' is not defined.
     23 'remote' is not defined.
     20 'empty' is not defined.
     19 'currentTask' is not defined.
     16 'bookmarks' is not defined.
     12 'urlParser' is not defined.
     11 'registerCustomBang' is not defined.
     10 Use path.join() or path.resolve() instead of + to create paths.
     10 'settings' is not defined.

PalmerAL:
> So No, do not switch to fast-forward merging, as that would add to the history everything I have commited/reverted (like the package-lock.json).

OK, sounds good!

Regarding modules, the issue I'm concerned about is not breaking things unexpectedly in the future. For example, suppose you have scripts A, B, and C, which are all loaded globally, and all depend on module M. If you only import M from A, everything will work at first, because B and C can reuse the same import. But if you convert A to a module, B and C will break, which requires you to go back and fix a lot of unrelated code before you can convert A to a module.

But I would agree, the eventual goal is to just convert everything, use a linter, and avoid this entirely.

code-hunger:
> If you only import M from A, everything will work at first, because B and C can reuse the same import. But if you convert A to a module, B and C will break, which requires you to go back and fix a lot of unrelated code before you can convert A to a module

That already happened to me :) It was quite annoying until I figured out what was it, 

Initially, after making a file a module, I grepped the whole codebase and imported the module in every file it was used. With const, of course.

I expected that to solve the problem you describe.And it almost did. When concatenated, the files using const for the same variable crashed, of course, because the value was overwritten.

So, we should use var. It will not crash now, and will solve the problem for the future. Are you ok with this solution?

PalmerAL:
I think so; that's what I've been doing for the modules that I've converted so far, and I haven't really run into any problems with it. Once the conversion is done, we can probably go back and change everything to use const.

code-hunger:
Ok, I'll add the 'var' imports into all files with urlParser/TO_builder references then, in an hour or two.

code-hunger:
I also had to make sure I don't overwrite the hosts variable because I was getting some strange errors. It's really a quickfix that'll be removed anyway in my next branch where I take hosts out of urlParser.

PalmerAL:
Looks good to me! Can you remove the "sorted_errors" files? The hosts issue is probably because that gets used in hostsSuggestionPlugin.js, I would agree splitting hosts into a separate module is the best solution for this.

code-hunger:
Sorry about these files… I didn't expect GitHub would add to the PR the files I uploaded to my comment. I used the link bellow the comment input box to 'attach' them, but it turns out it doesn't work as I thought.

Cleaned them.

code-hunger:
Actually you're right. After fixing the imports I added the whole js/ directory to the tree, without noticing the error files among the bunch of other files. That's clearly seen in the commit history, I need to look carefully when adding a lot of files together.

PR #621, opened 2018-09-16
Modularize hosts list generation

code-hunger:
The hosts list are not exclusively bound to the urlParser. This way is more testable and reusable.

### DIFF
--- a/js/searchbar/hostsSuggestionsPlugin.js
+++ b/js/searchbar/hostsSuggestionsPlugin.js
@@ -1,8 +1,7 @@
 var searchbarPlugins = require('searchbar/searchbarPlugins.js')
 var searchbarUtils = require('searchbar/searchbarUtils.js')
 var urlParser = require('util/urlParser.js')
-
-// hosts are parsed in util/urlParser
+const hosts = require('util/hosts.js')
 
 function showHostsSuggestions (text, input, event, container) {
   empty(container)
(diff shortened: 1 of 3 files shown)


PalmerAL:
Looks good to me! One (really minor) thing: can we avoid using if statements without braces? They do take up a bit less space, but it's not that much of a difference, and it makes it harder to see what is inside the if statement.

Also, do you mind making a list (either here or in the other issue) of what modules you're working on? I was planning to convert some more stuff in the next few days, but I want to make sure I don't unnecessarily duplicate something you're already working on.

code-hunger:
Oh, sorry about the if-s. Fixed that. Currently I'm not working on anything, rather I have a few branches for which I'm just opening PRs separately so that it's easier for you to see the changes in each step.

After this, I have a tabState_refactor branch and then a webviewMenu_refactoring. I'll open a PR after you merge this with the next one, but will wait for you so that I can rebase on the latest changes and make it clean, otherwise you might run into merge conflicts.

code-hunger:
BTW I was tempted to rewrite the function even more and completely remove its body by directly providing the return statent via the arrow syntax:
```
const truncatedHostsFileLines = (data, limit) =>
  data.length > limit
    ? data.substring(0, limit).split('\n').slice(0, -1)
    : data.split('\n')
```

But I decided I've probably gone too far... :)

PalmerAL:
OK, good to know!

It is interesting how much you can shorten things with ES6, but I would agree we don't necessarily need to! Ultimately, making it readable is much more important.

PR #622, opened 2018-09-16
Convert the taskPrototype to a TaskList class

code-hunger:
Rewrote it as a class. It has a mroe clear interface, with helpers for map/forEach/… on the inner `tasks[]` array (65ed6c4224fc742aaf0ef1ec5cb66b4bd06f5a7c), some optimizations (54ab215fbf9a53c3e56d3f8c83ad2edfa16330c8 - arr.map().indexOf() is inefficient, 013c01f9b2b5e2735cf42c8f9e16da2e272f0772 - no need to sort the whole array just to take the lowest value; here I've also ran `standard`, which made some extra styling changes), removed all direct references to the inner task array (helps us think about `tasks` as a *task list*, and not an *enhanced array of task objects* - 65ed6c4224fc742aaf0ef1ec5cb66b4bd06f5a7c), also rewrote some parts using the provided helpers to clean the code and avoid duplication (54ab215fbf9a53c3e56d3f8c83ad2edfa16330c8).

It got a bit longer than the previous PRs because there was more to clean, but if you look at it commit-by-commit, it should be clear. Let me know if you have any questions/critics.

### DIFF
--- a/js/api-wrapper.js
+++ b/js/api-wrapper.js
@@ -79,13 +79,13 @@ function closeTask (taskId) {
   if (taskId === currentTask.id) {
     // the current task was destroyed, find another task to switch to
 
-    if (tasks.get().length === 0) {
+    if (tasks.getLength() === 0) {
       // there are no tasks left, create a new one
       return addTask()
     } else {
       // switch to the most-recent task
 
-      var recentTaskList = tasks.get().map(function (task) {
+      var recentTaskList = tasks.map(function (task) {
         return { id: task.id, lastActivity: tasks.getLastActivity(task.id) }
       })
 
--- a/js/api-wrapper.js
+++ b/js/api-wrapper.js
@@ -38,7 +38,6 @@ options
   options.openInBackground - whether to open the tab without switching to it. Defaults to false.
 */
 function addTab (tabId = tabs.add(), options = {}) {
-
   tabBar.addTab(tabId)
   webviews.add(tabId)
 
@@ -89,12 +88,12 @@ function closeTask (taskId) {
         return { id: task.id, lastActivity: tasks.getLastActivity(task.id) }
       })
 
-      // sort the tasks based on how recent they are
-      recentTaskList.sort(function (a, b) {
-        return b.lastActivity - a.lastActivity
-      })
+      const mostRecent = recentTaskList.reduce(
+        (latest, current) =>
+          current.lastActivity > latest.lastActivity ? current : latest
+      )
 
-      return switchToTask(recentTaskList[0].id)
+      return switchToTask(mostRecent.id)
     }
   }
 }
@@ -109,7 +108,8 @@ function closeTab (tabId) {
 
   if (tabId === tabs.getSelected()) {
     var currentIndex = tabs.getIndex(tabs.getSelected())
-    var nextTab = tabs.getAtIndex(currentIndex - 1) || tabs.getAtIndex(currentIndex + 1)
+    var nextTab =
+      tabs.getAtIndex(currentIndex - 1) || tabs.getAtIndex(currentIndex + 1)
 
     destroyTab(tabId)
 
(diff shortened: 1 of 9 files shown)


PalmerAL:
I haven't had time to review all the details of this yet, but it looks pretty good overall! I'll try to finish reviewing it in the next day or two.

js/api-wrapper.js
```diff
@@ -79,22 +78,22 @@ function closeTask (taskId) {
   if (taskId === currentTask.id) {
     // the current task was destroyed, find another task to switch to
 
-    if (tasks.get().length === 0) {
+    if (tasks.getLength() === 0) {
```
PalmerAL: The equivalement method for tabs is `tabs.count()`, we probably want to use the same name for both (whichever one you want to use is fine with me).

js/keybindings.js
```diff
@@ -353,15 +353,10 @@ settings.get('keyMap', function (keyMapSettings) {
...
-      return task.id
-    }).indexOf(currentTask.id)
+    const currentTaskIdx = tasks.indexOf(currentTask)
 
-    if (tasks.get()[currentTaskIdx + 1]) {
-      browserUI.switchToTask(tasks.get()[currentTaskIdx + 1].id)
-    } else {
-      browserUI.switchToTask(tasks.get()[0].id)
-    }
+    const nextTaskIndex = (currentTaskIdx + 1) % tasks.getLength()
```
PalmerAL: It's neat that you can do this, but I'm not sure using it is a good idea. With the old way, it's pretty easy to tell that it's switching to the next task, or the first task if that one doesn't exist, whereas this way it takes a second to think through the math and figure out what it's doing. If we wanted to shorten this, something like this might work:

```
(tasks.byIndex(currentTaskIdx + 1) || tasks.byIndex(0)).id
```

js/tabState/task.js
```diff
@@ -1,93 +1,79 @@
...
-    }
-    return -1
-  },
-  setSelected: function (id) {
-    tabState.selectedTask = id
-    window.currentTask = tasks.get(id)
+  }
+
+  get (id) {
+    return this.tasks.find(task => task.id == id) || null
```
PalmerAL: Array.find is like 10x slower than a for loop ([test](https://jsperf.com/find-vs-loop)). I'm not sure if this actually gets called enough that it would matter, but we might want to check how much of an impact this has.

code-hunger: In that case we can keep the benefits of both ways by only changing the implementation of the TaskList::find () method, and use that instead of "this.tasks.find ()"

js/api-wrapper.js
code-hunger: Of course! Since `getLength()` does in fact `this.tasks.length`, I think its name is quite appropriate. 
So I'd suggest that the tabs' equivalent is changed to this one - and that would probably happen when the tabs prototype goes through the same type of change as this PR does to the task list.

js/tabState/task.js
code-hunger: Addressed in https://github.com/minbrowser/min/pull/622/commits/cd468e78c4f4a5460b36ee86f5dce1f8e0e7331a

js/keybindings.js
code-hunger: Addressed in 3486a1a.

code-hunger:
Done. Let me know if you find any further issues.

js/keybindings.js
```diff
@@ -374,15 +369,13 @@ settings.get('keyMap', function (keyMapSettings) {
...
+          taskCount = tasks.getLength()
 
-    if (tasks.get()[currentTaskIdx - 1]) {
-      browserUI.switchToTask(tasks.get()[currentTaskIdx - 1].id)
-    } else {
-      browserUI.switchToTask(tasks.get()[tasks.get().length - 1].id)
-    }
+    // The addition of taskCount fixes 0 --> -1
+    // E.g. with 5 elements: (0-1+5)%5 = 4, i.e. the last index
+    const previousTaskIndex = (currentTaskIdx - 1 + taskCount) % taskCount
```
PalmerAL: Can you do something similar here as above, like this?
```
    const previousTask = tasks.byIndex(currentTaskIdx - 1) || tasks.byIndex(tasks.getLength() - 1)
```

code-hunger: Oh sorry, didn't see it. Will fix it ASAP.

code-hunger:
I just looked at the PRs overall LOC stats - `+123 −130`. I got the source 7 lines less :laughing:

PalmerAL:
Looks good to me! (and less code is nice as well!)

PR #625, opened 2018-09-21
Convert tabColor to a module

PalmerAL:
@code-hunger Any suggestions?

I had to add an additional file, `deferredLoad.js`, that initializes the tabColor module. The reason for this is that in order to attach the events correctly, tabColor has to be initialized after the `webviews` object is created, but before the first view is created. I expect the solution to this is to convert webviews to a module and have tabColor depend on it, but that's going to require a bunch of other changes first.

### DIFF
--- a/js/navbar/tabColor.js
+++ b/js/navbar/tabColor.js
@@ -2,6 +2,12 @@ var colorExtractorImage = document.createElement('img')
 var colorExtractorCanvas = document.createElement('canvas')
 var colorExtractorContext = colorExtractorCanvas.getContext('2d')
 
+const defaultColors = {
+  private: ['rgb(58, 44, 99)', 'white'],
+  lightMode: ['rgb(255, 255, 255)', 'black'],
+  darkMode: ['rgb(40, 44, 52)', 'white']
+}
+
 function getColorFromImage (image) {
   var w = colorExtractorImage.width
   var h = colorExtractorImage.height
@@ -83,12 +89,6 @@ function getColorFromImage (image) {
   return res
 }
 
-const defaultColors = {
-  private: ['rgb(58, 44, 99)', 'white'],
-  lightMode: ['rgb(255, 255, 255)', 'black'],
-  darkMode: ['rgb(40, 44, 52)', 'white']
-}
-
 function getHours () {
   return new Date().getHours() + (new Date().getMinutes() / 60)
 }
@@ -98,38 +98,10 @@ var hours = getHours()
 // we cache the hours so we don't have to query every time we change the color
 setInterval(function () {
   hours = getHours()
-}, 4 * 60 * 1000)
+}, 5 * 60 * 1000)
 
 function getRGBString (c) {
   return 'rgb(' + c[0] + ',' + c[1] + ',' + c[2] + ')'
-}
-
-function updateTabColor (favicons, tabId) {
-  // private tabs always use a special color, we don't need to get the icon
-  if (tabs.get(tabId).private === true) {
-    return
-  }
-
-  requestIdleCallback(function () {
-    colorExtractorImage.onload = function (e) {
-      let backgroundColor = getColorFromImage(colorExtractorImage)
-      let textColor = getTextColor(backgroundColor)
-
-      let backgroundString = getRGBString(backgroundColor)
-
-      tabs.update(tabId, {
-        backgroundColor: backgroundString,
-        foregroundColor: textColor
-      })
-
-      if (tabId === tabs.getSelected()) {
-        updateColorPalette()
-      }
-    }
-    colorExtractorImage.src = favicons[0]
-  }, {
-    timeout: 1000
-  })
 }
 
 var getTextColor = function (bgColor) {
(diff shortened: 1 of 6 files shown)


js/navbar/tabColor.js
```diff
@@ -243,7 +200,61 @@ function setColor (bg, fg) {
   }
 }
 
-// theme changes can affect the tab colors
-window.addEventListener('themechange', function (e) {
-  updateColorPalette()
-})
+const tabColor = {
+  initialize: function () {
+    webviews.bindEvent('page-favicon-updated', function (e, favicons) {
```
code-hunger: You moved this method here probably with the hope to make tabColor.js easier to maintain, but in fact it makes it harder because it couples tabColor to webviews.

webviews.js binds all kinds of events and I think it makes sense if it serves some sort of an initializator for other modules. It'll be cleaner if they stay webviews-independent, and instead webviews loads them and initializes them.

But, as I don't have such a good picture in my mind of the entire browser, maybe you have strong reasons to think that's not the case. In that situation, to avoid doing this funny deferred loading, you can take a "webviews" parameter from `tabColor::initialize()`. Then, in the webviews, you'd call `tabColor.initialize(webviews)`. This way, `tabColor` is independent of the global `webviews` and you keep yourself from doing strange deferred loadings :)

js/navbar/tabColor.js
```diff
@@ -1,149 +1,123 @@
...
+  return res
+}
 
-      var obj = {
-        r: c[0] / 255,
-        g: c[1] / 255,
-        b: c[2] / 255
-      }
+function getHours () {
+  return new Date().getHours() + (new Date().getMinutes() / 60)
```
code-hunger: If you're making a function out of it anyway, you can accept an optional `date` parameter to avoid instantiation `new Date()` twice:

    function getHours (date = new Date()) {
      return date.getHours() + date.getMinutes() / 60
    }

code-hunger: Or even:

    const getHours = (date = new Date()) => date.getHours() + date.getMinutes() / 60;

js/navbar/tabColor.js
```diff
@@ -207,33 +181,16 @@ var runNetwork = function anonymous (input) {
...
 function setColor (bg, fg) {
-  var background = document.getElementsByClassName('theme-background-color')
-  var textcolor = document.getElementsByClassName('theme-text-color')
+  var backgroundElements = document.getElementsByClassName('theme-background-color')
+  var textElements = document.getElementsByClassName('theme-text-color')
 
-  for (var i = 0; i < background.length; i++) {
-    background[i].style.backgroundColor = bg
+  for (var i = 0; i < backgroundElements.length; i++) {
+    backgroundElements[i].style.backgroundColor = bg
```
code-hunger: An alternative style:

    backgroundElements.forEach(el => { el.style.backgroundColor = bg })
    textElements.forEach(el => { el.style.color = fg })

Just sayin'

code-hunger: Or with the traditional way, at least don't use `var i =…` for iteration, because it makes a function-scoped variable `i`, which is [might be] also used in any further loops. Use `let` instead.

js/navbar/tabColor.js
```diff
@@ -245,7 +200,61 @@ function setColor (bg, fg) {
...
+  },
+  updateFromImage: function (favicons, tabId) {
+    // private tabs always use a special color, we don't need to get the icon
+    if (tabs.get(tabId).private === true) {
+      return
+    }
+
+    requestIdleCallback(function () {
+      colorExtractorImage.onload = function (e) {
+        let backgroundColor = getColorFromImage(colorExtractorImage)
```
code-hunger: May I express my fondness towards `const`? :-P

js/navbar/tabColor.js
```diff
@@ -1,149 +1,123 @@
...
+  }
+  var output = runNetwork(obj)
   if (output.black > 0.5) {
     return 'black'
   }
   return 'white'
 }
 
+// generated using http://harthur.github.io/brain/
 var runNetwork = function anonymous (input) {
```
code-hunger: Might be a copy-paste typo, but looks out of place here - `var runNetwork = function anonymous (input)`. I also feel it should get out of the `tabColor.js` file, it occupies a good 60 lines here and is not strictly bound to the tab color logic, it's rather an implementation detail.

js/navbar/tabColor.js
```diff
@@ -243,7 +200,61 @@ function setColor (bg, fg) {
...
+    })
+  },
+  refresh: function () {
+    var tab = tabs.get(tabs.getSelected())
+
+    if (tab.private) {
+      // private tabs have their own color scheme
+      return setColor(defaultColors.private[0], defaultColors.private[1])
+    // use the colors extracted from the page icon
+    } else if (tab.backgroundColor || tab.foregroundColor) {
```
code-hunger: You don't need the `else` blocks, because you're doing `return X` in every if-block. Removing them will also flatten the function, making it flowing more straightforwardly.

js/navbar/tabColor.js
PalmerAL: getElementsByClassName returns an `HTMLCollection` that doesn't have a `forEach` method, so it would be something like `Array.prototype.forEach.call(backgroundElements, el => ...`. Maybe it's still worth it?

(I guess we could use querySelectorAll instead, but that seems kind of pointless.)

js/navbar/tabColor.js
PalmerAL: This is actually external code, so I've moved it to ext/tabColor/tabColor.js

js/navbar/tabColor.js
PalmerAL: Done in e30edd7a7d1bd6d5ad57d8431c0ccba7c9f8a325, everything that's constant should be a `const` now.

js/navbar/tabColor.js
code-hunger: > Interesting, I was actually planning in doing the opposite (moving all the events out of webviews.js and into the modules that they relate to.

I support that. I wouldn't have mentioned it at all if I knew you had these intentions.

> Keeping the events in the module does make it a bit easier to understand how the module works, and I was thinking it should allow us to add and remove modules more easily than if everything was linked together in webviews.js

You're right. As I didn't know you plan to apply it to everything, it seemed strange to me to export some 10 lines out of a 500 LOC file, but it makes perfect sense to continually reduce the webviews.js to something more easy to work with.

:+1:

code-hunger: > maybe it is better to do what you're suggesting, and just have the webview decide which modules to notify based on the events that are happening.

Are you talking about receiving `webviews` as a dependency of `initialize()`? Or about not having `initialize()` at all?

We can have an `initialize(webviews)` per module, and call it inside `webviews.js`.

PalmerAL: No, I just meant leaving the events inside the webviews.js file - sorry that wasn't worded very well!

Eventually, we should just be able to have each module use `const webviews = require("webviews.js")`, but since the webviews aren't a module yet, we can't do that.

code-hunger: In that case, you can just bring `tabColor::initialize()` back to `js/webviews.js` (with the theme change listener) and remove `deferredLoad.js` to resolve this (last) conversation.

PalmerAL: Assuming that tabColor is a completely separate module with all the webview events in it, I think the eventual goal would be to initialize it from `default.js` and have the tabColor module import the webview module. I can move the `require("tabColor.js")` into webviews.js, but I don't really see why that would be any better than leaving it separate.

PalmerAL:
I'm going to merge this, and leave the deferred.js fix for later. We aren't really able to fix that right now, and having it in a separate file is less confusing than including it in webviews.js is.

PalmerAL:
deferredLoad.js is gone in 0123ec11b02720708b6bdc2b998841e277b7baaf.

PR #626, opened 2018-09-21
Webview menu refactoring

code-hunger:
Uhhhhm… I'm sorry it became that long, but everything here is strictly related to splitting the menu into manageable parts (ok, that's a lie - 16a2b4f90ae8656f9aaaa969a927ed4e73c11ecc. But it's the only exception, and it's very simple, I promise). I also think if you look at the commits one by one, it should be clear what I'm doing. I think there shouldn't be anything bigger than a few lines change (except maybe for 89efa9e13bfc0f80284a453d0ca35d14edd1550a, which just looks scary but most of it is indentation fix) or a file split (where I do pure copy-paste, to allow git to keep tracking changes on split files)

I'll make a summary of the split-and-extract commits (the harder to keep an eye of) so that you know 'the big picture' here:

1. I extract the different menu parts from the `showMenu` function, like link menu/image menu/selection etc. 1198e92c583e360f0c6e115b5bf42a1f6fefabf6
2. I make a `makeMenuSections` function which combines the sections from above. This further simplifies the `showMenu()` function 89efa9e13bfc0f80284a453d0ca35d14edd1550a
3. I put the sections & `makeMenuSections` into different files under the `webviewMenu/` directory. 5cd3aee506f9b3039c964e18050ca636c8556030

Everything else are shorter commits which make small steps towards improving the way it looks. I also added an option to get out of the focus mode, though I'm not sure how to implement it.

I hope you like it because I really think it makes dealing with this part much more easy. Just look at the finished webviewMenu.js and menuSections.js, 25 and 62 LOC each.


PalmerAL:
Thanks for doing this! I haven't had a chance to review everything in detail yet, but a couple of general comments:

* I like separating the menu items based on section (makeNavigationMenuItems, makeLinkMenuItems, etc), but these are all similar enough to each other that I don't really see the purpose of splitting them into separate files. Having them in one file does seem kind of long and repetitive, but separating them doesn't really do anything to reduce the complexity, and it makes it harder to see how the overall menu fits together.
* I feel like makeMenuSections should just be part of createWebviewMenu, since the purpose of that function is to create the menu anyway.
* We could reconsider showing both the image and the link actions if you think it would be better that way - can you open an issue about that?
* Having the menu items be plain objects, and then calling new MenuItem() on them at the end is nice.
* I'm not sure I see what the issue with having webviewMenu depend on searchEngine is - did you have something in particular in mind with that?

PalmerAL:
I'm not sure I agree with adding a focus mode item. There are two possible scenarios when you open the menu:

* You enabled focus mode by accident, or enabled it and forgot about it.
* You enabled it intentionally, because you don't want to be distracted.

I think the current behavior does a good job addressing both of these scenarios: if you enabled focus mode by accident, you'll still see the familiar menu items, but if you try to click on them, you'll get instructions about how to disable it. If you enabled focus mode intentionally, disabling it is somewhat difficult: you have to close the instructions, then go to the app menu, and select the option to disable it, so you're less likely to become distracted.

With this change, the menu becomes more difficult to understand if you are in focus mode accidentally - because all the normal menu items will be gone, replaced by something new. Additionally, if you're in focus mode intentionally, it becomes much easier to disable it, which probably isn't a good thing.

PalmerAL:
I'll have to think more about the alt vs option change - internally we store everything as `option`, but we're still showing the right thing in the settings page, and I think mousetrap will use the right one automatically depending on the platform. I'll try to look into that soon.

Reading back over my comments, they sound kind of negative, but I think this is actually pretty good overall! I think there's a bit too much abstraction in some places, but separating out the items does make it easier to understand, and it's definitely better than what we had before.

code-hunger:
>  Having them in one file does seem kind of long and repetitive, but separating them doesn't really do anything to reduce the complexity, and it makes it harder to see how the overall menu fits together.

Can you really see how the overall menu fits together? By just "having a glimpse" over lines from 16 to [183](https://github.com/minbrowser/min/blob/75c1bffef1938fe0644773ff8171ae4431c08db6/js/webviewMenu.js#L183)? I can't. When the whole logic is kept in js/webviewMenu/menuSections.js between L11 and L59, at least for me, it's more doable.

> I feel like makeMenuSections should just be part of createWebviewMenu, since the purpose of that function is to create the menu anyway.

I made a separate file for it because makeMenuSections' job is to glue together the different sections. It's different from createWebviewMenu as the latter builds a `Menu` instance using `MenuItems`, which are none of makeMenuSections' concern.  
These two are logically independent of each other and each one can be easily replaced with something else if needed, or can be reused in other code. The menu sections are Menu-API agnostic, while `createWebviewMenu` is sections-agnostic, i.e. it can be given any section builder and will work with it, while the section builder can be given to any Menu API to make the UI elements.

> We could reconsider showing both the image and the link actions if you think it would be better that way - can you open an issue about that?

In https://github.com/minbrowser/min/pull/626/commits/d3260d1219bbc717e9182afbe4fc4d5ad8b41dcf I revert this change as I see the comment you've left about keeping the 2 parts mutually exclusive. I'd make it configurable, but currently it is quite hard to take parameters out of the depths of the webviewMenu and expose them to public modification. (That'd be a *very* good design architecture, especially for a browser that should let the user configure it. We'd need to make extensive use of dependency injection, but that'd be possible only after code is well modularized)

> Having the menu items be plain objects, and then calling new MenuItem() on them at the end is nice.

Yep. That's part of achieving point 2 - make the menu generation independent of the menu API used.

code-hunger:
> I'm not sure I see what the issue with having webviewMenu depend on searchEngine is - did you have something in particular in mind with that?

I've left no descriptive comment on that 11 days ago and now I can only guess what I meant. I think it was needed because I'm taking webviewMenu.js out of the file concatenation list, hence we've got no access to the global variables.

The best way I could see to have the searchEngine inside webviewMenu was to mark it as a dependency. We might also `import` it (in the future, as currently searchEngine is not an importable module), but then it would be a fixed value, and we'd loose flexibility if we want to support live searchEngine configuration.

For example, if the user changes the searchEngine, on the next `showMenu()` call we'd just pass the new value. If we had imported it, we'd be unable to do that - we'd have to mess with settings and add unnecessary dependencies and coupling to irrelevant parts.

code-hunger:
> I'm not sure I agree with adding a focus mode item. […] I think the current behavior does a good job addressing both of these scenarios: if you enabled focus mode by accident, you'll still see the familiar menu items, but if you try to click on them, you'll get instructions about how to disable it.

I've never really used that feature myself, so my opinion doesn't have much weight here. I just think from a UX POV it's better to hide some option, rather than have it present and make it give errors. I can revert it. BTW,  "Allow edits from maintainers." is checked on, so you must be able to push commits here too, if you want.

> I'll have to think more about the alt vs option change - internally we store everything as option, but we're still showing the right thing in the settings page, and I think mousetrap will use the right one automatically depending on the platform.

I made this change because the `Alt+m` shortcut I've set for hiding the menu was not working. After I debugged the settings page and inspected through the settings object, I saw it was replaced with 'option', which according to the docs must be done only on OSX, and in the code did exactly the opposite:

https://github.com/minbrowser/min/blob/75c1bffef1938fe0644773ff8171ae4431c08db6/pages/settings/settings.js#L259-L267

So I moved the alt→option replace inside the check for mac os.

PalmerAL:
> I've never really used that feature myself, so my opinion doesn't have much weight here. I just think from a UX POV it's better to hide some option, rather than have it present and make it give errors. I can revert it. BTW, "Allow edits from maintainers." is checked on, so you must be able to push commits here too, if you want.

OK, I've reverted those two commits. Avoiding errors is generally good, but I think in this case it makes it easier to understand what's happening than showing a different set of options.

> I made this change because the Alt+m shortcut I've set for hiding the menu was not working. After I debugged the settings page and inspected through the settings object, I saw it was replaced with 'option', which according to the docs must be done only on OSX, and in the code did exactly the opposite:

OK, I think I understand what's happening now. Currently, all the shortcuts in `defaultKeyMap.js` are stored using `option`. When we show them on the settings page, we replace all instances of `option` to `alt` for non-mac platforms, and when you edit the keybinding, we convert it back to `option` before saving. Previously, this worked, because Mousetrap would [convert the option to alt when deciding what shortcuts to run](https://github.com/ccampbell/mousetrap/blob/master/mousetrap.js#L130). However, the new method of binding keyboard shortcuts (in keybindings.js, line 483) doesn't check for that, so any shortcut with `option` or `alt` won't work on any platform, mac or not.

Converting the `option` to `alt` won't work completely, because it won't affect saved shortcuts, and it will still be broken on mac. I think the best short-term fix is to make the `before-input-event` handler match Mousetrap's behavior, and treat the `option` and `alt` keys like the same thing. In the long-term, we should figure out how to make the shortcut handling simpler so this doesn't happen (seriously, we have 3 different ways of handling a shortcut now), but that's going to involve a lot more work.

code-hunger:
> The currentSearchEngine variable is set directly on the window, so you should be able to access it from inside a module - I haven't tested it in this case, but I can't think of any reason why it wouldn't work.

That's the problem - the module has a hidden dependency, which will make it harder to find what breaks in the future.

If `webviewsMenu` requests no `searchEngine` as a parameter, or as an import, it won't be easily visible it is using it. Then, if we change the way `searchEngine` is retrieved (we put it in a module/make a function/etc.) it will silently introduce a bug, which will appear only after you test the webviewsMenu and click on the search engine option and see that it doesn't work.  
And even then, you'd have to debug, inspect the code, locate the error in the browserified thousand line file and track it back to the tiny webviewsMenu.js.

> menuItems.js has the items together. While this ends up being pretty long, it's pretty linear (and the items are all in the same order as what's in the menu), so I don't think it's that difficult to follow conceptually.

Ok, I'll put them back together shortly.

> webviewMenu.js has the code to combine the sections together and build the menu, which gives you the overview of how the sections fit together.

My idea was to split these two different jobs into their own files. Look at how you said "webviewsMenu has to do X and Y".  
It's useful to have X and Y in one place when they're interrelated and highly dependent on each other, but here X does not depend on Y at all, and is big enough (50 LOC) to deserve its own file.

Also, Y has external dependencies that X (Menu/MenuItem/Clipbord) that X doesn't need to use, and decoupling X into its own file makes that clear, and also ensures that in the future it won't mess up with other things.

---

I'm fine with having them together (the arguments above were just my reasons to do it the other way, for you to see *why* I did it), but to avoid having both webviewsMenu-related files in the global `js/` directory, I'd suggest having them in the `webviewsMenu` directory, with the `webviewsMenu.js` renamed to `index.js`. This should still allow us to `require('webviewsMenu')` while not polluting the main directory with feature-specifics.

PalmerAL:
> That's the problem - the module has a hidden dependency, which will make it harder to find what breaks in the future.

It's hidden right now (and I agree depending on global variables to be what you expect isn't great), but once searchEngine becomes a module, this problem should go away - you could just look for all the files that import searchEngine.js (and that file looks pretty easy to modularize, so I don't expect this to be a problem for very long - in fact, I'm probably going to do it this weekend if we can merge this PR).

> My idea was to split these two different jobs into their own files. Look at how you said "webviewsMenu has to do X and Y".
> It's useful to have X and Y in one place when they're interrelated and highly dependent on each other, but here X does not depend on Y at all

OK, I can kind of see that, if you want to leave them separate that's fine with me.

>  I'd suggest having them in the webviewsMenu directory, with the webviewsMenu.js renamed to index.js. This should still allow us to require('webviewsMenu') while not polluting the main directory with feature-specifics.

Sounds good!

PalmerAL:
The webview menu has changed a lot since this was written, so it's no longer possible to merge this as-is. If you want to continue working on it, feel free to open a new issue (or PR).

Issue #619, opened 2018-09-14
Turn off colors in tab bar


jackrieger:
Really like the design and viewpoint of this browser.

I wish there was a simple toggle in preferences to keep the tab bar always white, as it looks on the search page, instead of the colors that change dynamically.

PalmerAL:
Fixed in 8dbfdef8c36052d400c6fa8dfe84522778f8a2d4.

PR #632, opened 2018-10-02
Added Site Themes toggle for #619

PandawanFr:
So #619 asked for a simple toggle in the preferences to disable site colors (like when you go on Google and it changes tabs to blue or something). 

This will require adding new translations for every language, I have only added for en_us for now. 
Also it might conflict with other current PRs that are refactoring the tabColor module though it shouldn't be too hard to fix. 

Another idea to make this better would be to allow for custom theming (where users can select from a variety of colors instead of a simple toggle), but this would be outside of the scope of the original issue.

### DIFF
--- a/js/navbar/tabColor.js
+++ b/js/navbar/tabColor.js
@@ -214,7 +214,7 @@ window.updateColorPalette = function () {
     // private tabs have their own color scheme
     return setColor(defaultColors.private[0], defaultColors.private[1])
   // use the colors extracted from the page icon
-  } else if (tab.backgroundColor || tab.foregroundColor) {
+  } else if (window.allowSiteThemes && (tab.backgroundColor || tab.foregroundColor)) {
     return setColor(tab.backgroundColor, tab.foregroundColor)
   // otherwise use the default colors
   } else if (window.isDarkMode) {
--- a/js/util/theme.js
+++ b/js/util/theme.js
@@ -4,14 +4,9 @@ function shouldEnableDarkMode () {
 }
 
 settings.get('siteThemes', function(value) {
-  if (value === true) {
-    window.allowSiteThemes = true
-    window.dispatchEvent(new CustomEvent('themechange'))
-  } else {
-    window.allowSiteThemes = false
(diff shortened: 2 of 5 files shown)


Nybbit:
You can cut down on the code [here](https://github.com/minbrowser/min/blob/8b079640de745b00a969b51d4f110c8c536a47bd/js/util/theme.js#L6-L14) and turn it into
```js
settings.get('siteThemes', function(value) {
    window.allowSiteThemes = value
    window.dispatchEvent(new CustomEvent('themechange'))
});```

PalmerAL:
The code looks pretty good, but I'm wondering if this is something we want to do in the first place. A lot of the decisions about the UI (having everything be the same color and using opacity/weight/size to communicate things instead) were assuming that the background color would continually changing. If we assume that the background color is always the same, then we would probably want to design the rest of the UI differently as wel.

However, it does seem like this is something people want (aside from the linked issue, a couple of people have mentioned it in the survey as well), so I guess I'm OK with adding this if there's enough interest. At the same time, I'd like to understand why people don't like the effect. Is it just distracting? Are there certain sites where the color-selection algorithm doesn't work very well? Does it make the tab bar harder to see?

PandawanFr:
Hmmm, yeah that's understandable. I feel like some people just really like the aesthetics/simplicity of the pure white theme. 

Also, there really isn't a need to redesign the UI, make it so that site themes are a default and that if people don't like it, they can always make it white. This would make it so that you keep designing under the assumption that people will use the site themes with varying background colors so that the UI works with any background (even a basic white color).

PalmerAL:
Yeah, I can see how it might be less distracting. Keeping everything monochrome is probably what we would do, I'm just not sure it looks very good - it seems very "empty". I think I'm going to wait and see if more people request this.

PalmerAL:
There were more requests to add this in #790, so I went ahead and added this option in 8dbfdef8c36052d400c6fa8dfe84522778f8a2d4 (with a couple changes from this PR to account for the settings redesign in #766).

PR #654, opened 2018-11-03
Added Persian localization

hossainalhaidari:
This commit adds Persian (Fa) localization to the project


PalmerAL:
Everything looks good, thanks! We may need to make some changes to the interface to support RTL text correctly; if you notice anything, please open an issue about that.

PR #659, opened 2018-11-04
Suggest that OS X users install node with Homebrew

rjoyal:


rjoyal:
The installer for OS X at https://nodejs.org/en/ installs to /usr/local/ with files owned by root. This conflicts with Homebrew which wants to manage files in /usr/local/ owned by the user.

PalmerAL:
Thanks for the tip! This seems like general advice that would apply to any Node project, so I'm not sure we need to include it, although maybe it's helpful if this is your first time using Node.

Also, I'm curious what the issue with using the nodejs installer with homebrew is? I have both of those installed, and I haven't noticed any issues so far - is there a specific part of homebrew that won't work with node installed this way?

rjoyal:
Maybe not a real runtime issue, but "brew doctor" complains since you have
stuff owned by 'root' in /usr/local after using the nodejs installer.

PalmerAL:
I didn't know that, thanks! Since this seems like general advice rather than something specific to Min, and the impact is fairly low, I don't think it makes sense to include it in the instructions, but thanks for the advice!

Issue #639, opened 2018-10-13
Enable content blocking by default


PalmerAL:
While we currently have a setting to enable content blocking, it's not very obvious that the setting exists in the first place, or how to enable it. Ads on many websites tend to be very slow, which makes the experience of using Min fairly unpleasant, and many ad networks have pretty invasive tracking, which is also a problem.

To address this, I think we should start prompting people to pick a content-blocking setting during the tour that shows up when you install Min. If we do this, more people will realize that content blocking is an option, and people will hopefully have a better experience with Min.

There are (valid) concerns that a lot of websites' business models are based on ads, and I think it is worth addressing that. To that effect, I think we should do a couple of things:

* Add an option to allow ads on certain sites (#204), which would be editable from the settings page (we could probably also add a searchbar command to make this even easier).
* Provide three levels of blocking - "block nothing", "block third-party ads", and "block everything". The default option would be for only third-party ads to be blocked, meaning that sites that run their own ads (I'm thinking of something like https://krebsonsecurity.com/) could continue to display ads. Because most of the performance and privacy issues are coming from third-party ad networks, this should address most of those issues while still providing websites a way to make money.

To summarize:
* If you're an existing user with content-blocking enabled, you would be switched to the "block all ads" level, which matches what the content blocking setting does currently.
* If you're an existing user who enabled content blocking, and later disabled it, you would be switched to the "block nothing" level.
* If you've never modified content blocking settings, you would be switched to the "block third-party ads" level, but could disable it from the settings page.
* If you're a new user, you will be prompted to pick one when you first run the browser. If you close the tour without picking anything, "block third-party ads" will be enabled.

Any thoughts?

LukaFX:
It should be in on the startpage of the browser. There should also be an easy way to turn it on and off.

PalmerAL:
@LukaFX I definitely agree about turning it off, there will be an option on the settings page (where the existing content blocking settings are now). Regarding the tour/welcome page, I think I'm just going to include instructions on going to the settings page for now, and probably add an actual checkbox when the tour page gets redesigned at some point.

PalmerAL:
This is done in 0cfcb67a41cddab6c839291fce932026251da5d3; I'm going to leave this issue open until we update the tour to reflect these changes.

PR #691, opened 2019-01-04
Add options to block third-party ads and add site exceptions

PalmerAL:
This addresses most of #639:

* It adds a third content-blocking level (block only third-party ad and tracking requests), which will become the default in 1.9.
* It adds an option to disable content-blocking for a specific site.

Screenshot:
contentblockingui

I think it looks a little cluttered - maybe separating the script and image options more clearly would help - but it should be OK for now.

Once this is merged, I'll change the default setting to third-party blocking in a separate PR.


LukaFX:
Will you be able to whitelist sites with the block all ads and trackers option?

PalmerAL:
@LukaFX Yes, when you select that option the UI changes to this:

contentblockingui2

I can see how that would be confusing, though, maybe there's a better way to do this.

PR #699, opened 2019-01-15
ar.json

anejjar:
arabic translation

### DIFF
--- a/localization/languages/ar.json
+++ b/localization/languages/ar.json
@@ -1,0 +1,345 @@
+{
+    "name": "Arabic",
+    "identifier": "{
+    "name": "Arabic",
+    "identifier": "ar",
+    "translations": {
+        /* Context menu items
+        these are the items displayed in the menu when you right-click on a page
+        */
+        "openInNewTab": "فتح الرابط في علامة تبويب جديدة",
+        "openInNewPrivateTab": "فتح الرابط في التصفح المتخفي",
+        "viewImage": "مشاهدة الصورة",
+        "openImageInNewTab": "فتح الصورة في علامة تبويب جديدة",
+        "openImageInNewPrivateTab": "فتح الصورة في التصفح المتخفي",
+        "saveImageAs": "حفظ الصورة باسم",
+        "searchWith": "%s بحت مع", //%s will be replaced with the name of the current search engine
+        "copyLink": "نسخ الرابط",
+        "copy": "نسخ", //this is a verb (as in "copy the currently-selected text")
+        "paste": "لصق",
+        "goBack": "الرجوع الى الوراء",
+        "goForward": "الانتقال الى الامام",
+        "inspectElement": "تدقيق العنصر",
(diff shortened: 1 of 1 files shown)


PalmerAL:
Thanks for translating everything!

While looking through this using Google Translate (in order to make sure there aren't issue like strings being matched to the wrong ID), I noticed that in a couple instances, the translation provided doesn't match what I would expect, and that Google Translate is showing "did you mean" suggestions that change a single character. Specifically:

* For many of the strings, it suggests changing "ت" to "ث", and if I make that change, then the translation matches what I would expect.
  * Examples: searchWith, searchHistory, searchInPage, archiveSearchAction, genericError, settingsSearchEngineHeading, customSearchEngineDescription
* For several other error message, it fails to translate "هادا", and suggests changing it to "هذا", which results in the expected translations for the errors.
* For "settingsDarkModeToggle" it translates it as "enable instant mode", and suggests changing "اليلي" to "الليلي", which results in the expected translation of "enable night mode".

I realize Google Translate is oftentimes inaccurate, and I don't have any way of judging for myself which way is correct, so it's perfectly fine to leave these the way they are if that's the correct way - I just wanted to check with you and make sure that everything's correct before merging this.

PalmerAL:
I'm not saying we should use the google translation version directly! You probably know what the right thing to do here is a lot better than I do - I just wanted to point that out in case you'd made a mistake when writing it, but I'm fine with whatever you think is correct.

anejjar:
"settingsContentBlockingLevel0": "السماح لجميع الاعلانات و المتعقبين",
        "settingsContentBlockingLevel1": "حظر اعلانات الطرف التالت و المتعقبين",
        "settingsContentBlockingLevel2": "حظر جميع الاعلانات و المتعقبين",
        "settingsContentBlockingExceptions": "سيظل مسموحًا بعرض الاعلانات على هده المواقع",
         "settingsEasyListCredit": {
             "unsafeHTML": " استنادا الى EasyList و  EasyPrivacy, التي تم انشائها من طرف  <a target=\"_blank\" href=\"https://easylist.to/\"> EasyList المؤلفون ل</a> (<a target=\"_blank\" href=\"https://creativecommons.org/licenses/by-sa/3.0/legalcode\">مشاهدة الترخيص</a>)."
         },

PalmerAL:
Thanks! Whatever you want to do with the easylist string is fine.

Just to make sure, are the Google Translate suggestions accurate? If they're not, and the way you translated it originally is correct, then it's completely fine to do that instead.

anejjar:
everythings is okey!
the google translation is just literal translation and the result of it not making any sense 
the complex words needs translation of what that word mean.
but leave it as it is in english letters looks fine to me like the name of this browser cant translate it to arabic it should stay as it is 
finally you can merge this pull requests  :)
and i'm here for any future change

PalmerAL:
OK, great! Thanks for doing this!

PR #707, opened 2019-02-08
update zh-TW translation

z-huang:
Since Min has updated to 1.9.0, there's something new to translate, so I updated the zh-TW translation file and modified some text to make the meaning more complete.


PalmerAL:
Thanks!

PR #718, opened 2019-03-10
show error if duplicate settings

### LINKED ISSUE #717: Disallow duplicate keyboard shortcuts
Two keyboard shortcuts being assigned to the same combination (i.e. setting "Add Private Tab" and "Restore Tab" to the same keys) will cause both keyboard shortcuts to execute.  

Recommend updating pages/settings/settings::onKeyMapChange to check for duplicates before submit.

(I'm looking into this, I think this could be a great first PR for me for this project!)

tjwds:
Closes #717.

Please feel free to reach out if there's a convention that I didn't follow correctly!  This requires an additional translation string for all locales.

### DIFF
--- a/pages/settings/settings.js
+++ b/pages/settings/settings.js
@@ -354,8 +354,8 @@ function onKeyMapChange (e) {
     // if the keymap before change already contains this key combination...
     Object.keys(currentKeyMap).forEach(keySetting => {
       if (currentKeyMap[keySetting] === keyMapSettings[action]) {
-        keyMapSettings[keySetting] = 'unassigned';
-        document.getElementById(keySetting).value = 'unassigned';
+        keyMapSettings[keySetting] = 'unassigned'
+        document.getElementById(keySetting).value = 'unassigned'
       }
     })
 
--- a/pages/settings/settings.js
+++ b/pages/settings/settings.js
@@ -2,6 +2,7 @@ document.title = l('settingsPreferencesHeading') + ' | Min'
 
 var container = document.getElementById('privacy-settings-container')
 var banner = document.getElementById('restart-required-banner')
+var bannerError = document.getElementById('settings-error-banner')
 var darkModeCheckbox = document.getElementById('checkbox-dark-mode')
 var historyButtonCheckbox = document.getElementById('checkbox-history-button')
 var swipeNavigationCheckbox = document.getElementById('checkbox-swipe-navigation')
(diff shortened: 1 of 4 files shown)


PalmerAL:
Rather than showing an error, I think it might be better to just update the settings so there aren't duplicates - if you have a key combination assigned to action A, and you assign the same combination to B, then I think we should just remove the combination from A and only assign it to B. In most cases, if you are re-using a shortcut, it's probably because you forgot about the old shortcut and aren't using it anyway, and that way we could avoid having a separate UI for this, which I think might be nice since this probably isn't a scenario that will happen very often. What do you think?

tjwds:
That's fair! Let me make some changes.

tjwds:
@PalmerAL 👍

tjwds:
The more I think about this, this seems a little too hacky.

Setting the overwritten setting to `''` doesn't work, because it'll then be ignored by the keybindings controller.

Let me think about this a little more and see if there's a way to refactor things that won't be too intrusive!

PalmerAL:
I think ignoring empty strings was intentional; if you remove all the shortcuts for an action, it will revert back to whatever the default is. I can see how that might not be a good idea, though, if we wanted to get rid of that I think we could just get rid of the if statement here: https://github.com/minbrowser/min/blob/master/js/util/defaultKeyMap.js#L29 so that all the settings get applied even if they're empty, and then maybe check [here](https://github.com/minbrowser/min/blob/master/js/keybindings.js#L185) that the shortcut isn't an empty string before registering the shortcut.

pages/settings/settings.js
```diff
@@ -347,7 +347,18 @@ function onKeyMapChange (e) {
       keyMapSettings = {}
     }
 
+    var currentKeyMap = {...defaultKeyMap, ...keyMapSettings}
```
PalmerAL: There's a `userKeyMap` function (from defaultKeyMap.js) that will do this already.

pages/settings/settings.js
```diff
@@ -347,7 +347,18 @@ function onKeyMapChange (e) {
...
+
     keyMapSettings[action] = parseKeyInput(newValue)
+
+    // if the keymap before change already contains this key combination...
+    Object.keys(currentKeyMap).forEach(keySetting => {
+      if (currentKeyMap[keySetting] === keyMapSettings[action]) {
```
PalmerAL: Another thing to consider here is that it's possible to have multiple shortcuts for a single action separated by commas; you probably need to split the shortcut value into an array of items, check if any of the items in the array are identical to the new shortcut, and remove them from the array if they are.

PalmerAL:
I'm going to close this PR since it's been open for a while, but feel free to make a new PR if you want to continue working on it!

PR #726, opened 2019-03-26
View Source

AwesomeTech10:
Untested!

What this is _supposed_ to do is add a View Source option to the menu. I have added the translation to en-US only, and NOTHING IS TESTED! Please make sure this code works. If not, than I have a backup code to try.

### DIFF
--- a/js/webviewMenu.js
+++ b/js/webviewMenu.js
@@ -185,6 +185,16 @@ var webviewMenu = {
         }
       })
     ])
+    
+    /* view source (havent tested this! if not, add view-source: to beginning of url) */
+    menuSections.push([
+      new MenuItem({
+        label: l('viewSource'),
+        click: function () {
+          webviews.get(tabs.getSelected()).viewSource(data.x || 0, data.y || 0)
+        }
+      })
+    ])
 
     menuSections.forEach(function (section) {
       section.forEach(function (item) {
--- a/localization/languages/en-US.json
+++ b/localization/languages/en-US.json
@@ -18,6 +18,7 @@
         "goBack": "Go Back",
         "goForward": "Go Forward",
         "inspectElement": "Inspect Element",
+        "viewSource": "View Source",
         /* searchbar */
         "DDGAnswerSubtitle": "Answer", //this is a noun - it is used as a subtitle when displaying Instant Answers from DuckDuckGo in the searchbar
         "suggestedSite": "Suggested site", //this is used to label suggested websites from the DuckDuckGo API,


PalmerAL:
This won't work because viewSource isn't a method of [webContents](https://electronjs.org/docs/api/web-contents); if we wanted to add this, putting view-source: in front of the URL is probably the correct way to do it.

However, I'm not sure we really need to add a menu item for this; if you want to view the page source, you can go to inspect page > elements, or just type view-source in front of the URL itself (although to be fair, I'm guessing not everyone knows about that).

Thanks for working on this, though!

AwesomeTech10:
I’ll work on my other idea, because I do prefer to right-click and see the View Source option?
    	
    	

    	Sent from my awesometech10@icloud.com email address.
    
  




On Sun, Mar 31, 2019 at 7:22 PM -0700, "PalmerAL" <notifications@github.com> wrote:












This won't work because viewSource isn't a method of webContents; if we wanted to add this, putting view-source: in front of the URL is probably the correct way to do it.


However, I'm not sure we really need to add a menu item for this; if you want to view the page source, you can go to inspect page > elements, or just type view-source in front of the URL itself (although to be fair, I'm guessing not everyone knows about that).


Thanks for working on this, though!

PalmerAL:
I guess I'm just not convinced many people would find this useful. What advantage does "view source" have over the inspector for you?

AwesomeTech10:
The advantage is it is easier for me to see what the pure source code is, and most browsers have this, which makes the transition neater

PalmerAL:
If you want to access this easily, I think you could do it as a [userscript](https://github.com/minbrowser/min/wiki/userscripts):
```
// https://stackoverflow.com/questions/1815021/programmatically-open-view-source-html-window-in-browser-with-javascript

function showSource(){;
    var source = "<html>";
    source += document.getElementsByTagName('html')[0].innerHTML;
    source += "</html>";
    //now we need to escape the html special chars, javascript has escape
    //but this does not do what we want
    source = source.replace(/</g, "&lt;").replace(/>/g, "&gt;");
    //now we add <pre> tags to preserve whitespace
    source = "<pre>"+source+"</pre>";
    //now open the window and set the source as the content
    sourceWindow = window.open("data:text/html," + encodeURIComponent(source))
}  

window.addEventListener("keypress", function(e) {
    //when "v" key is pressed, show source
    if (e.keyCode === 118 && !["INPUT", "TEXTAREA"].includes(e.target.tagName)) {
        showSource()
    }
});
```
I just don't really see this being useful to enough people to justify adding it to the menu. If there's data that indicates that this is frequently used in other browsers, or if more people request this feature, then  we could reconsider adding it.

danrobi11:
@PalmerAL 

> I just don't really see this being useful to enough people to justify adding it to the menu

I dont know for others but I do quite often have to see the page source. 
That option is usually in the `right-click-menu` in other common browser. 
I was pretty surprised not seeing it in min.

> if you want to view the page source, you can go to inspect page > elements, or just type view-source in front of the URL itself

Thanks for the tips, I wasnt aware about that.

PalmerAL:
What do you typically do with the page source? My thinking is that "inspect element" is a more useful way to look at the page source in most cases; if you need to copy the source, you can either 1) right-click in the inspector and use the "copy" option, or 2) use the userscript above.

PR #738, opened 2019-04-27
custom bang for closing tasks

brakkum:
I was getting annoyed having to manually open the task overlay, and mouse over to the trash icon to remove tasks. This custom bang just gets rid of tasks by name/number or the current if nothing is specified.

### DIFF
--- a/js/searchbar/customBangs.js
+++ b/js/searchbar/customBangs.js
@@ -160,6 +160,26 @@ registerCustomBang({
 })
 
 registerCustomBang({
+  phrase: '!closetask',
+  snippet: l('closeTask'),
+  isAction: false,
+  fn: function (text) {
+    if (text) {
+      taskToClose = getTaskByNameOrNumber(text)
+      if (taskToClose) {
+        browserUI.closeTask(taskToClose.id)
+      }
+    } else {
+      browserUI.closeTask(tasks.getSelected().id)
+    }
+    taskOverlay.show()
+    setTimeout(function () {
+      taskOverlay.hide()
+    }, 600)
+  }
+})
+
+registerCustomBang({
   phrase: '!bookmarks',
   snippet: l('searchBookmarks'),
   isAction: false,
--- a/js/searchbar/customBangs.js
+++ b/js/searchbar/customBangs.js
@@ -164,18 +164,24 @@ registerCustomBang({
   snippet: l('closeTask'),
   isAction: false,
   fn: function (text) {
+    var currentTask = tasks.getSelected()
+    var taskToClose
+
     if (text) {
       taskToClose = getTaskByNameOrNumber(text)
-      if (taskToClose) {
-        browserUI.closeTask(taskToClose.id)
-      }
     } else {
-      browserUI.closeTask(tasks.getSelected().id)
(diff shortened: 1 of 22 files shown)


PalmerAL:
Thanks, this looks really good!

One small suggestion: I think it only makes sense to show the task overlay when switching between tasks, so if the task being closed isn't the current task (and therefore the active task won't change), it would probably be better not to show anything there. Other than that, I think everything looks good.

brakkum:
Good point! Refactored it a bit to match what I believe you're getting after.

PalmerAL:
Looks great, thanks!

PalmerAL:
By the way, I just added the ability to tab through items in the task overlay in 0b10907a06f6cc00da0302c87fbc48d487a96749. Right now, the close task button isn't selectable from the keyboard, because I was thinking it might be to easy to close tasks by accident, but since you're mentioning that it's annoying to click on it, I'm wondering if it would be better to make it selectable. What do you think (or is there anything else you think could be improved about keyboard navigation)?

brakkum:
That might be useful, it would certainly be nice for situations where you know you want to close a previous task, but aren't certain of the number/name. As far as it being easily executed on accident, that is a risk, especially since there isn't any task history. I'll check out the task overlay tabbing when I get a chance and ruminate on it a little more.

PalmerAL:
A task history is a pretty good idea, actually. We probably don't want to store it permanently (since otherwise it would be difficult to actually delete anything), but I'm thinking that when you click the delete button, we could replace the task with a placeholder that says something like "Task deleted. Undo?", which would disappear after a few seconds.

brakkum:
I think that'd be a great thing to add, just as a fail safe if anything. Or even just tracking the most recently closed task might be good too, in case a user thinks later that they could use something from the last task they had open. I know I've had moments like that in the past.

PalmerAL:
Is it more that you delete it and realize shortly afterwards you needed something, or you realize a longer time later and after multiple tasks have been deleted?

One thing I've been thinking about is changing the task overlay so that tasks are sorted based on how recently you've used them, and tasks you haven't opened in a long time would get collapsed so they don't take up as much space. That way, you could leave tasks open for longer without them getting in the way as much, so that you wouldn't have to delete anything unless you're really sure you don't need it anymore (although the actual state of the pages open would probably be discarded to save memory). But I'm not sure if that would address what you're talking about?

I think I'm going to include a survey about the tasks feature in the next release; it would be good to know what other people think about this as well, since I don't want to disrupt anyone's workflow too badly if they're relying on the existing feature.

brakkum:
Usually it's that I'm looking something up, find the answer, and promptly close the window. Then need it again withint a few minutes or so. At least that's my browsing experience.
I tend to keep as few tasks open as possible, so for me It's less about cleanliness and more about closing things prematurely. Granted, I should know myself better by now and not do that, but that's usually how it goes for me.
It would probably good to poll users about this, especially if you're thinking about introducing large changes. I think your idea of sorting them by recently used would be really good, I'd say when I do have multiple tasks open, I tend to work in the most recent ones, but can sometimes want to find something in one further back.

Issue #753, opened 2019-05-18
Request: context menu with "cut, paste, etc" not show when right click on search bar.


MairwunNx:
It would not be bad to add a context menu to the search field, in some situations when I hold a beer in my left hand it is not convenient for me to press `ctrl + c \ v`, it’s more convenient with a mouse in this situation, but it’s a pity that the context menu wasn’t :(

Tomadelostacos:
Nice comment ! I'm on it, it works, i just need to translate it before pushing. (Cut, Copy, Paste, Select All, Undo or Redo)

Tomadelostacos:
#754 it's here (translated in french, and in english atm for all other languages, with a comment to let people translate by themselves, i'm not polyglot enough for all those languages)

MairwunNx:
@Tomadelostacos good job! I review your PR and suggest strings for russian strings.

Tomadelostacos:
Resolved in #754 ! (Not by me though, sorry ! )

PR #754, opened 2019-05-18
[WIP] ✨feat (adress bar input): implements context menu (copy, cut, paste, …  …select all, undo,redo)

Tomadelostacos:
Added some words when "Missing translation" was said in comments
Modified some words, in accordance with what i often see in modern web browsers

#753

### DIFF
--- a/js/contextMenu.js
+++ b/js/contextMenu.js
@@ -2,9 +2,7 @@ document.getElementById('tabs').addEventListener('contextmenu', (e) => {
   e.preventDefault();
   e.stopPropagation();
 
-  if (e.target.classList.contains('tab-input')){
-
-    remote = electron.remote, Menu = remote.Menu;
+    Menu = remote.Menu;
 
     InputMenu = Menu.buildFromTemplate([{
       label: l('undo'),
@@ -31,7 +29,6 @@ document.getElementById('tabs').addEventListener('contextmenu', (e) => {
     },
     ]);
       InputMenu.popup(remote.getCurrentWindow());
-  }
 });
 
 
--- a/js/contextMenu.js
+++ b/js/contextMenu.js
@@ -1,34 +1,41 @@
-document.getElementById('tabs').addEventListener('contextmenu', (e) => {
-  e.preventDefault();
-  e.stopPropagation();
+document.addEventListener('contextmenu', (e) => {
+  e.preventDefault()
+  e.stopPropagation()
 
-    Menu = remote.Menu;
+  Menu = remote.Menu
 
-    InputMenu = Menu.buildFromTemplate([{
-      label: l('undo'),
-      role: 'undo',
-    }, {
-      label: l('redo'),
-      role: 'redo',
-    }, {
-      type: 'separator',
-    }, {
-      label: l('cut'),
-      role: 'cut',
-    }, {
-      label: l('copy'),
-      role: 'copy',
-    }, {
-      label: l('paste'),
-      role: 'paste',
-    }, {
-      type: 'separator',
-    }, {
-      label: l('selectAll'),
-      role: 'selectall',
-    },
-    ]);
-      InputMenu.popup(remote.getCurrentWindow());
-});
+  InputMenu = Menu.buildFromTemplate([{
+    label: l('undo'),
+    role: 'undo'
+  }, {
+    label: l('redo'),
+    role: 'redo'
+  }, {
+    type: 'separator'
+  }, {
+    label: l('cut'),
+    role: 'cut'
+  }, {
+    label: l('copy'),
+    role: 'copy'
+  }, {
+    label: l('paste'),
+    role: 'paste'
+  }, {
+    type: 'separator'
+  }, {
+    label: l('selectAll'),
+    role: 'selectall'
+  }
+  ])
 
+  let node = e.target
 
+  while (node) {
+    if (node.nodeName.match(/^(input|textarea)$/i) || node.isContentEditable) {
+      InputMenu.popup(remote.getCurrentWindow())
+      break
+    }
+    node = node.parentNode
+  }
+})
--- a/js/contextMenu.js
+++ b/js/contextMenu.js
@@ -1,40 +1,37 @@
-window.electron = require('electron')
-window.remote = electron.remote
-window.Menu = remote.Menu;
+document.getElementById('tabs').addEventListener('contextmenu', (e) => {
+  e.preventDefault();
+  e.stopPropagation();
+
+  if (e.target.classList.contains('tab-input')){
+
+    remote = electron.remote, Menu = remote.Menu;
+
+    InputMenu = Menu.buildFromTemplate([{
+      label: l('undo'),
+      role: 'undo',
+    }, {
+      label: l('redo'),
+      role: 'redo',
+    }, {
+      type: 'separator',
+    }, {
+      label: l('cut'),
+      role: 'cut',
+    }, {
+      label: l('copy'),
+      role: 'copy',
+    }, {
+      label: l('paste'),
+      role: 'paste',
+    }, {
+      type: 'separator',
+    }, {
+      label: l('selectAll'),
+      role: 'selectall',
+    },
+    ]);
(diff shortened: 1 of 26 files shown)


localization/languages/ru.json
```diff
@@ -13,6 +13,10 @@
         "saveImageAs": "Сохранить изображение как...",
         "searchWith": "Искать в %s", //%s will be replaced with the name of the current search engine
         "copyLink": "Копировать ссылку",
+        "select-all": "Select All", // Missing Translation
+        "undo": "Undo", // Missing Translation
+        "redo": "Redo", // Missing Translation
+        "cut": "Cut", // Missing Translation
```
MairwunNx: ##### Please, change strings for russia

```
        "select-all": "Выделить все",
        "undo": "Отменить",
        "redo": "Повторить",
        "cut": "Вырезать",
```

MairwunNx:
Also please, mark PR as `[WIP]` for protection against merging with the main branch.

js/default.js
```diff
@@ -43,6 +44,47 @@ if (window.platformType === 'windows') {
   })
 }
 
+// ISSUE 753: Someone wants to right click on address bar to copy/paste/stuff
```
PalmerAL: Can this go in a separate file? (and also the comment probably isn't necessary)

js/default.js
```diff
@@ -43,6 +44,47 @@ if (window.platformType === 'windows') {
...
+InputMenu = window.Menu.buildFromTemplate([{
```
PalmerAL: This is going to send a synchronous IPC to the main process, which could potentially slow down startup, so it might be better to delay creating the menu until the contextmenu event actually happens.

localization/languages/ar.json
```diff
@@ -13,6 +13,10 @@
         "saveImageAs": "حفظ الصورة باسم",
         "searchWith": "%s بحث مع", //%s will be replaced with the name of the current search engine
         "copyLink": "نسخ الرابط",
+        "select-all": "Select All", // Missing Translation
```
PalmerAL: Can you make these `null` instead, and for the other languages as well? localizationHelpers.js will use the English version automatically if there isn't a translated version available, and that way if we decide to change the strings later, we can do so without updating it in multiple places (although in this instance, changing the English strings does seem rather unlikely).

PalmerAL: Also, could this be `selectAll` to match how the other strings are named?

localization/languages/fr-FR.json
```diff
@@ -89,7 +93,7 @@
         "addressInvalidTitle": "Cette adresse est invalide.",
         "genericError": "Il y a une erreur",
         /* pages/phishing/index.html */
-        "phishingErrorTitle": "Ce site peut vous nuire.",
+        "phishingErrorTitle": "Ce site peut vous nuire. (Tentative de Phishing)",
```
PalmerAL: This looks like it doesn't match the English string, if the title's confusing we should probably try to fix it for every language. However, the phishing warning page is currently disabled completely, so it shouldn't really affect anything to leave it as-is.

localization/languages/ko.json
```diff
@@ -13,6 +13,14 @@
         "saveImageAs": "이미지를 다른 이름으로 저장",
         "searchWith": "%s에서 검색", //%s will be replaced with the name of the current search engine
         "copyLink": "주소 복사",
+        "select-all": "Select All", // Missing Translation
+        "undo": "Undo", // Missing Translation
+        "redo": "Redo", // Missing Translation
+        "cut": "Cut", // Missing Translation
+        "select-all": "Select All", // Missing Translation
```
PalmerAL: It looks like these maybe got duplicated, can you remove the second set of strings?

localization/localizationHelpers.js
```diff
@@ -21,7 +21,7 @@ function getCurrentLanguage () {
   } else {
     // nothing worked, fall back to default
   }
-
+console.log(language);
```
PalmerAL:
```suggestion
```

localization/languages/fr-FR.json
Tomadelostacos: Okey, i suggest so "Ce site pourrait vous nuire", as it matches more the english translation. ("peut" is more like "can harm you") 
It's going to be committed

Tomadelostacos:
All the changes are on the new commit ! Except for :  
> This is going to send a synchronous IPC to the main process, which could potentially slow down startup, so it might be better to delay creating the menu until the contextmenu event actually happens.

I deleted some lines, but i'm not an Electron senior (yet ? ), it surely can be optimized

Tomadelostacos:
Moved the menu in the tabBar.js file, it was stupid from me to create another file for this

MairwunNx:
Please, see my review with added russian strings, not ignore it!

PalmerAL:
@MairwunNx Don't worry, I'm going to add it.

MairwunNx:
@PalmerAL thanks.

PalmerAL:
@Tomadelostacos Thanks! I've fixed a couple of other minor things, I think all the translations look good now.

Regarding the context menu, the call to `buildFromTemplate` has to send the contents of the menu to the main process, which then sends the result back to the renderer process, which ends up taking around 4ms. On its own, this isn't a whole lot, but when you add a bunch more things that are all sending their own messages, it can start to make startup slower. So I think it would be better to create the menu when the contextmenu event happens (when the extra delay isn't as likely to be noticeable):

```
    function buildContextMenu() {
        return Menu.buildFromTemplate({
        ...
    }
    ...
     if (node.nodeName.match(/^(input|textarea)$/i) || node.isContentEditable) {
        buildContextMenu().popup(...
```

Tomadelostacos:
> Please, see my review with added russian strings, not ignore it!

Damn, i'm brainless, added them then removed when i removed all other languages strings, but we got them now !  
For the context menu, i'm going to have a look on that, but i'm not confident, i'm quite new at Electron, so i focus on simple issues to train myself.

PalmerAL:
OK, let me know if you have any questions, I don't think you should have to change any of the context menu code very much, just place it inside a function that gets called from the `contextmenu` event.

js/contextMenu.js
```diff
@@ -0,0 +1,37 @@
+document.getElementById('tabs').addEventListener('contextmenu', (e) => {
+  e.preventDefault();
+  e.stopPropagation();
+
+  if (e.target.classList.contains('tab-input')){
```
PalmerAL: Is there a reason the menu needs to be only for the tab input? Unless you're planning to add more actions that are specific to the address bar, I think it would be helpful to have this work in the task overly as well.

js/contextMenu.js
```diff
@@ -0,0 +1,37 @@
...
+
+    remote = electron.remote, Menu = remote.Menu;
```
PalmerAL: `remote` is already defined in default.js, so you don't need to re-define it here.

PalmerAL:
@Tomadelostacos I updated this to use the code from the original commit you made, which checks for the node name to make sure the menu only shows for `input` elements, and not other parts of the tab bar, and also made the listener apply to the whole document so it will work in the task overlay as well.

PalmerAL:
With that, I think everything looks good! Thanks for working on this!

Tomadelostacos:
Nice ! Thanks to you ! I'm quite new to this framework, so i'm gonna check if there's some little things i can do to improve this browser, and my skills by the way.

PalmerAL:
Sounds good, let me know if you have any questions and I'm happy to answer them.

Now that there's a UI menu with cut/copy/paste/undo, it would be good to support these in the menu that shows up when you right-click on a webpage as well. That menu currently supports copy and paste ([here](https://github.com/minbrowser/min/blob/master/js/webviewMenu.js#L136-L152)), but not undo, redo, or select all. Let me know if you're interested in working on that, or if you're not I'll probably do it myself at some point.

Tomadelostacos:
I did the select all option, it's in the next pull request ! I'm gonna go on undo and redo asap, as i think i know how to add them

Issue #703, opened 2019-01-24
Active tab highlight


Gabrieleiro:
Highlight color(s) for active tabs should blend with the current page's main color. 

The way it is now is counter intuitive and it leads you to think inactive tabs are the ones you're looking at.

PalmerAL:
That's an interesting idea, it does seem to help make it somewhat easier to see which tab is selected (example: [before](https://user-images.githubusercontent.com/10314059/51722257-6e1acf00-201a-11e9-957b-a5525e007f00.png) vs [after](https://user-images.githubusercontent.com/10314059/51722267-75da7380-201a-11e9-9c88-c5fa1ba3f3cc.png)). However, I think it does have a few downsides:

* If the majority of the tab bar is either much lighter or darker than the accent color (which we would need to do to provide contrast for the current tab), it makes the tab bar look kind of washed out.
* A lot of sites have a fixed header that's the same color as their theme color. When most of the tab bar is this same color, it creates the appearance of everything being a single bar, which reduces clutter, but the effect doesn't work very well when most of the tab bar is the accent color instead.

It's definitely something to consider, though.

Tomadelostacos:
I personally use a bold font for the active tab. I didn't commit anything, but it's a line i can commit if you think the idea is good

PalmerAL:
That sounds like a good idea, I'll try it for a few weeks and see how well it works.

PalmerAL:
@Tomadelostacos I've been testing this with `font-weight: 500`, and I think it does make it easier to see the active tab, which is nice! However, changing the font weight also makes the width of the text change as you switch tabs, which is really annoying. To fix this, I think we'd have to adjust the letter spacing on bold text so that it ends up being the same width; there's a couple ways we could do this:

1. Hard-code a bold letter-spacing for each [supported font](https://github.com/minbrowser/min/blob/master/css/base.css#L6), and just let the letter spacing be incorrect if we fall back to the generic `sans-serif` font.
2. Compute the correct letter-spacing on startup by putting some text in a container, measuring its width, and progressively shrinking its letter-spacing until it matches the width of non-bolded text. This would work regardless of the font, but it seems fairly complicated, and might also be slow (although maybe we could cache it?)

There's also an interesting difference depending on the background color - with white text on a color background, the bolding seems fairly reasonable, but with black text on a white background, it seems too bold. I think we can fix this by switching to a dark-gray text color instead of pure black (which is something we should probably do anyway).

[...]

PR #758, opened 2019-05-23
Select All option in context menu, small animation on add-tab button

Tomadelostacos:
I did :  
- In context menu, now you can select all webpage content
- Added a small animation on the "add tab button", i thought it needed some system "graphic event" on hover to make it look more like a button

- #703 made me think about the bold font for the active tab, to make it easier for the user to know where he is, "at first sight", i did revert it for this PR though

### DIFF
--- a/css/tabBar.css
+++ b/css/tabBar.css
@@ -114,6 +114,9 @@ body.is-edit-mode #navbar #tabs {
 	word-break: break-all;
 	position: relative;
 }
+.tab-item.active {
+	font-weight: bold;
+}
 .tab-item:not(.active):hover {
 	background-color: rgba(255, 255, 255, 0.1);
 }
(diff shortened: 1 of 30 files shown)


PalmerAL:
Hmm, I was thinking about select all more as an action for text fields in webpages than for every part of the page. Other browsers do include a "select all" option for everything, but it seems like it woudn't be very useful, since it's almost certainly going to include parts of the page (such as navigation elements) that you don't want to copy?

In order to make it only show up on text fields, it looks like you could check for `data.isEditable` first, and only append the item if it's true.

Also, could you move the item into the clipboard actions section (along with copy and paste)

I like the animation!

I'm still thinking about the bold tab, in general I think the idea's good, but it's going to be more complicated to implement than I thought; I'll leave a comment on the other issue about it.

PalmerAL:
Since this PR now has several conflicts, I'm going to close it, but feel free to make a new PR if you want to continue working on it.

(An animation for the new tab button is being worked on in #1116)

Issue #82, opened 2016-04-20
min behind a proxy


ljoli:
Hello everybody,

I would like to use min behind a proxy. 

How can I do that ?

Thanks for all

PalmerAL:
Most operating systems have a setting to use a proxy - For example, OS X instructions are <a href="https://support.apple.com/kb/PH18553">here</a>. If you enable a proxy in your operating system, it should work in Min.

There currently isn't a way to set a proxy for just Min, but it wouldn't be too difficult to add that using the <a href="http://electron.atom.io/docs/v0.37.7/api/session/#sessetproxyconfig-callback">session.setProxy</a> API.

ljoli:
@PalmerAL  I use XUBUNTU on my enterprise and I can't  change any file of system. So, That is why it is better to add proxy setting on web browser MIN. ;-)

ds2k5:
Hi,
are there any plans to get proxy settings in min-browser ?
thanks

michalpl7:
Would be nice if You add separate proxy settings.

20NE:
love min's simplty. Yes. I'm looking forward to  proxy setting feature, with proxy auto-config even better.

skygate2012:
It's really useful to have custom proxy settings because I don't want to proxy the whole system and I need to set different proxy ports for different applications.

d4tocchini:
to set proxy globally, for all sessions, you can set the relevant chromium command line flag before electron app ready event like so:

```
const proxyHost = '127.0.0.1:8888'
electron.app.commandLine.appendSwitch('proxy-server', proxyHost)
```

xboston:
TOR proxy:
```
const proxyHost = 'socks5://127.0.0.1:9050' 
app.commandLine.appendSwitch('proxy-server', proxyHost)
```

fudosakura:
I was able to add a simple proxy tor host to my session as @PalmerAL stated in `main.js`, now I'm developing a component in settings where you can switch proxy on or off and set proxy host and port by input.

Sestowner:
If you don't want to set proxy system-wide, just run Min with `--proxy-server` option.

HTTP/HTTPS: `min --proxy-server=address:port`

SOCKS: `min --proxy-server=socks5://address:port`

maxwtBvGBfwR:
That's not the case, then you can't switch back to normal middle session or switch to other proxy.

PR #820, opened 2019-09-06
Proxy settings

maxwtBvGBfwR:
Added proxy configurations for min, which will add two no proxy, and manual proxy mode to min
manual proxy is configurable for http, https, ftp, sock4 and sock5.

It does not have ip validator and also I think syntax is not min acceptable so it needs some review.
will close issue:  #82


pages/settings/settings.js
```diff
@@ -100,7 +100,8 @@ for (var contentType in contentTypes) {
   (function (contentType) {
     settings.get('filtering', function (value) {
       // create the settings section for blocking each content type
-
+      if ( value == undefined )
```
PalmerAL: Value is already checked below, so I don't think this is necessary.

pages/settings/settings.js
```diff
@@ -267,6 +268,8 @@ settings.get('keyMap', function (keyMapSettings) {
   var keyMapList = document.getElementById('key-map-list')
 
   Object.keys(keyMap).forEach(function (action) {
+    if (action == undefined)
```
PalmerAL: I don't think `Object.keys` can ever return an undefined item, because that would mean one of the keys in the object doesn't have a name, which as far as I know isn't possible.

PalmerAL:
Thanks for working on this!

In general, I'm not sure I'm comfortable with the amount of code this adds for this feature. One thing that stands out to me in particular is that there's a lot of code to save settings for different protocols, and then assemble a URL string based on the settings, but if I understand correctly, the generated URL looks like this in all cases:
```
[protocol]://[hostname]:[port]
```

Given that, I would be fine with just having a single textbox in the settings where you enter the proxy URL, and then have Min use that proxy URL directly (maybe with some validation to make sure the URL is formatted correctly). I'm guessing that in most cases, someone who's trying to configure a proxy has some understanding of how a URL is constructed, and we can create a wiki article and add a link to it in case some additional instruction is necessary. But as long as that allows us to provide the same functionality (which I think is the case?) this would be much simpler to implement.

As of #766, you can use `settings` in any of the Min processes, and changes will automatically sync between processes. Given that, instead of putting the settings listener in the renderer process and sending an IPC message to the main process, you can just run this code directly in the main process.

I don't actually have a proxy to test this with (unless you know of any free ones to try?), but I'm assuming the actual Electron API works correctly.

js/util/proxy.js
```diff
@@ -0,0 +1,75 @@
...
+  },
+  initialize: function () {
+    settings.listen("proxyMode", function(value){
+      if (value == 'no-proxy') {
+        proxy.changeHandler("")
+      }
+      else {
+        var proxyRules = {}
+        settings.listen('proxies', function(proxy){
+          if ( proxy.httpHost ) {
```
PalmerAL: `proxy` could be undefined here if the setting has never been modified, so there needs to be a check for that.

js/util/proxy.js
```diff
@@ -0,0 +1,75 @@
...
+  },
+  changeHandler: function(proxyRules) {
+    proxy.rules.proxyRules = proxyRules
+    window.dispatchEvent(new CustomEvent("proxy-change", {
+      detail: proxy.rules
+    }))
+  }
+}
+
+if (require.main === module) {
```
PalmerAL: `proxy.js` isn't ever loaded not as a module, so this check shouldn't be necessary.

main/main.js
```diff
@@ -280,6 +280,11 @@ ipc.on('showSecondaryMenu', function (event, data) {
   }
 })
 
+ipc.on('proxyChange', function(event, data) {
+  session.defaultSession.setProxy(data)
```
PalmerAL: Private tabs use a different session then the default, so if we do this, anything you load in a private tab won't be proxied. I'm not actually sure what the best way to get a list of all the current sessions is, I'm guessing viewManager.js would need to be modified to keep a list of them. I'll try to think about it more and see if I can come up with a solution.

js/util/proxy.js
```diff
@@ -0,0 +1,75 @@
+if (typeof require !== 'undefined') {
+  var settings = require('util/settings/settings.js')
+}
+const { BrowserWindow } = require('electron')
```
PalmerAL: These imports aren't actually used anywhere, so they can be removed.

pages/settings/settings.js
fudosakura: The looping function recreates elements everytime a value changes

pages/settings/settings.js
fudosakura: Object.keys returns an empty array for empty Object, but the forEach loop returns undefined for that empty arrray.

Test this:
```js
Object.keys({}).forEach(function (action) { console.log(action) } );
```

js/util/proxy.js
fudosakura: Yes I should remove this, as I first wanted to add it in legacy mode but latter changed it.

main/main.js
fudosakura: The main reason I used ipc listener was that I couldn't have access to session anywhere else rather than main class, for private sessions we have to do the same this as data is the same proxy string here.

pages/settings/settings.js
PalmerAL: I'm not sure I understand what you mean, can you clarify?

This function is creating a checkbox for each content type, and if the value is true, it sets the checkbox to checked based on that. `Value`s existence is already checked on line 114, so an undefined value shouldn't cause an issue, and we still want the rest of the function to run (so that the checkboxes still appear).

pages/settings/settings.js
PalmerAL: I don't think `forEach` does that with an empty array - the function is just never going to be called at all.

I do see undefined in the console if I run that, but that's just indicating that the forEach function didn't return any value, not that the `console.log` function actually ran. If you try this, you shouldn't see any result:
```
Object.keys({}).forEach(function (action) { console.log("the key is", action) } );
```

(also, keyMap should never be empty anyway).

main/main.js
PalmerAL: > The main reason I used ipc listener was that I couldn't have access to session anywhere else rather than main class,

I think you can just move the entire `proxy.js` script to run in the main process; that would let you use the session module without an IPC message.

> for private sessions we have to do the same this as data is the same proxy string here.

Right, I'm just not sure how to get the `session` object that represents each private tab.

PalmerAL:
This has several merge conflicts now, so I'm going to close it. If you want to continue working on it, feel free to make a new PR!

Issue #391, opened 2017-08-31
Password manager


PalmerAL:
From #390:

> Also, it would be nice if min remembered my passwords... i made another bash program for this too but perhaps min could integrate with KDE's wallet or something. Not sure if a standard exists for this or perhaps gnome and kde have different incompatible interfaces for this.

kodaxx:
Sorry about the duplicate issue - I'll go ahead and respond here then. This library might be of some help to get this feature added. https://github.com/buttercup/buttercup-core

alexandrestein:
No news on this one?

I think that those days your default web browser has to support password management.
Most of us have probably many credentials from many website. 
This makes almost impossible to remember all of theme. Unless you always use the same password which is definitely not a good security practice.

Personally, I use random generated password remembered by the browser.
I would love to switch to Min but I really mist this feature.

alexandrestein:
@code-hunger maybe not every browsers but many of them.

I'm using it in chromium and Firefox. 
I checked about brave and Opera and both dose.

I guess edge, ie or Safari does it as well.

This makes probably around 95% of the browser usage which supports password management.

PalmerAL:
Yeah, I tend to agree with @code-hunger here. Password management would be a nice feature to have, but it's tricky to do correctly, and the consequences if you do it wrong are pretty bad (exposing passwords to the wrong websites, etc.) Maybe in the future it will make sense to implement this, but it would require a lot of time and thought, and I'm not sure it's something I really want to be responsible for.

In the meantime, you could probably make a script to autofill passwords with a [userscript](https://github.com/minbrowser/min/wiki/userscripts) if you wanted to.

alexandrestein:
@PalmerAL thank you for this answer.

From this point of view I could agree.
This can easily became dangerous.

It doesn't look out of reach to parse the actual domain and to propose all saved password which are saved with the exact same domain and those which are saved for the parent domains.

Some tools parse domain name like [this one](https://www.npmjs.com/package/domain-name-parser).
This can be an helper to look for related domains.

It's not a big deal, I'm not dying for it.
I was just wondering if there was any improvement.

PalmerAL:
This has been requested enough (see https://github.com/minbrowser/min/issues/532#issuecomment-458806491) that I'm re-thinking whether we want to do this. Like I said before, we'll have to be really careful that this is secure, so I think there's a couple things we should do to simplify this and reduce the risk:

[...]

PR #822, opened 2019-09-07
#391: Bitwarden Integration for Password Autofill

cog1to:
This PR will hopefully provide a path for adding support for more password stores/managers inside Min. Currently only Bitwarden integration is implemented because that's what I am using.

#### Autofill in action
password-fill-demo

#### Bitwarden setup dialog
bitwarden-setup

### What was added

- `ProcessSpawner` class that can run custom commands and return STDOUT/STDERR.
- Bitwarden class that can communicate with `bw` CLI utility and fetch for password data based on provided domain name. Users `ProcessSpawner`
- New section in Settings for selecting password manager/provider.
- General purpose prompt dialog.
- Preload script that checks page content for form inputs and sends IPC events to query for available credentials data.
- A bunch of setup code, that is triggered on browser startup: it checks currently selected password manager, and if it's not ' properly configured', it triggers setup logic.
- Setup dialog for Bitwarden.
- A `browserUI` method `navigateInNewTab` that opens an URL in a new tab: currently I use it to open Bitwarden site when user clicks on a link in Setup dialog.

### Usage

I implemented 3 ways of triggering auto-fill:
1. Automatically on page load events.
2. Via keyboard shortcut (mod+a).
3. Via new button in the tab bar area.

(1) will only work if password manager is already unlocked. (2) and (3) will prompt user for unlocking the password store if hasn't been unlocked yet.


PalmerAL:
I've just started reviewing this; it'll probably take me a bit due to the size, but I should be able to finish at least by next week.

js/passwordManager/passwordManager.js
```diff
@@ -0,0 +1,220 @@
...
+    webviews.callAsync(tab, 'getURL', null, (err, src) => {
+      var domain = new URL(src).hostname
+      if (domain.startsWith('www.')) {
+        domain = domain.slice(4)
+      }
+      
+      var self = this
+      manager.getSuggestions(domain).then(credentials => {
+        if (credentials.length > 0) {
+          webview.sendToFrame(frameId, 'password-autofill-match', credentials)
```
PalmerAL: I'm pretty sure this is a security issue - if the frame that sent the `password-autofill` event is from a third party, then the frame will receive credentials from the first-party site. We need to send the URL of the frame to the password manager, not the URL of the tab.

js/util/keyMap.js
```diff
@@ -19,7 +19,8 @@ var defaultKeyMap = {
   'closeAllTabs': 'shift+mod+n',
   'reload': 'mod+r',
   'showAndHideMenuBar': 'ctrl+m',
-  'followLink': 'mod+enter'
+  'followLink': 'mod+enter',
+  'passwordAutofillShortcut': 'mod+a'
```
PalmerAL: This is already a built-in shortcut on Mac (it's "select all"), so we'll need to pick a different key combination for this.

js/util/process.js
```diff
@@ -0,0 +1,42 @@
...
+Helper class to control child process output.
+Bufferizes output from stdout and stderr, waits until the process exits,
+and then resolves the promise with gathered data.
+*/
+
+var { spawn } = require('child_process')
+
+class ProcessSpawner {
+  constructor(command, args) {
+    this.command = command
```
PalmerAL: We're passing potentially untrusted input (such as domain names) to this, so there needs to be some kind of filtering in order to prevent running arbitrary commands. Based on the session key, I would suggest something like `replace(/[^a-zA-Z0-9+/=.-]/, '')` on each item in the array by default, and then use a special "escape" method on the domain names that removes anything that's not letter/number/period.

PalmerAL:
One other thing I'm noticing here is that there's a ton of complexity involved in creating a separate window for the setup, and then passing data between the windows; it's fairly hard to visualize how the data is actually moving between windows, and I expect that's going to become a source of problems in the future. Instead of a separate window, would it be OK to show the dialog in the same window? We could use the existing `webviews.requestPlaceholder` method to hide the browserView, and then just layer a `<div>` on top for the setup dialog, which would remove most of the IPC complexity.

If that sounds OK to you, I'd be happy to implement it.

js/passwordManager/passwordManager.js
cog1to: Re-implemented by sending `document.location` as a parameter to this IPC call. As I understand it, frames have their own `documents`, so that should send frame's actual location/hostname.

js/util/keyMap.js
cog1to: Any suggestions? I'm fine with whatever key is free :)

js/util/process.js
cog1to: Ok, it makes sense, but

1. Node's `spawn` seems to already escape it's parameters. It's not possible to execute, for example, something like `echo 0 && cat /etc/passwd` with `spawn('echo', ['0', '&&', 'cat', '/etc/passwd'])`.
2. This class is intended to execute whatever is passed to it. If we want some additional security, I think it should be done at places where it is used. In our case: inside Bitwarden PM code.

cog1to:
@PalmerAL If you feel it would be better for the overall app architecture, go for it. I would like to better understand how your approach will work though. Does this placeholder block the whole browser window or just one tab? If it blocks just one tab, we can probably do something even more simple with the same effect: just open a new tab with an internal page, similar to how Preferences work.

js/util/process.js
PalmerAL: Oh, cool, I didn't know that; just sanitizing the domain should be fine then.

PalmerAL:
It blocks the whole window; you can switch tabs, but the dialog will still appear in front:



Having the page in a tab would be nice, assuming it's possible to get the tool path. I don't think it is, since internal pages in tabs don't have node access (the preferences page has special IPC api's to read/write preferences). I guess you could do something where you get the tool path in the parent process, send it to the tab over IPC, then the tab sends back a command to run the tool, etc. but that would probably end up being even more complicated. (And I want to avoid enabling node access for internal tabs, since that adds a lot of complexity to webviews.js, and any mistake would result in giving node access to normal webpages, which would be bad).

PalmerAL:
Also, I may just be misunderstanding how this works, and I haven't fully gone through the code yet to find out what's supposed to happen, but I don't understand how to sign in after dropping the tool. The dialog asks for email/password, but there seems to only be one field:

cog1to:
> It blocks the whole window; you can switch tabs, but the dialog will still appear in front:
Looks pretty nice :+1: 

> The dialog asks for email/password, but there seems to only be one field
Oh, my bad, I broke it :) Pushed a fix.

PalmerAL:
I've added the initial version of the in-page setup dialog. I think it does simplify the data flow a bit, although not quite as much as I'd hoped for; let me know what you think.

One issue I ran into is that since the dialog blocks the window, there isn't an obvious way to show the tab with the bitwarden download page, so I changed the link to use a direct download link instead.

cog1to:
@PalmerAL Looks nice. Is there a way to disable task/tab bar when the dialog is present? It's a minor thing, but I can still switch between tabs and open Tasks menu when the dialog is present.

Also, pushed a small fix for the dark mode detection.

PalmerAL:
> Is there a way to disable task/tab bar when the dialog is present? It's a minor thing, but I can still switch between tabs and open Tasks menu when the dialog is present.

Not easily, although it would be nice if there was (since it would make implementing focus mode a lot easier). You can listen for events like tab-selected on tabState but there's no way to cancel them; I think you'd have to add a way to do that, and then modify anything that creates or modifies tabs to check whether the event is cancelled before doing anything.

PalmerAL:
Thanks for fixing the dark mode after I broke it. I switched it to use the existing global dark-mode class in b1dfafd03d5860b0a7ae4c4c47723fbd4c9c0c0b, which also has the benefit of handling cases where the theme changes as the dialog is open.

js/passwordManager/passwordManager.js
```diff
@@ -0,0 +1,219 @@
...
+
+  // Checks if Bitwarden integration is configured properly by trying to
+  // obtain a valid Bitwarden-CLI tool path.
+  async checkIfConfigured() {
+    this.path = await this._getToolPath()
+    if (this.path == null) {
+      return false
+    }
+
+    isCommandValid = await this._checkCommand(this.path)
```
PalmerAL: I'm not sure I understand why this is necessary; _getToolPath already verifies the path before it returns it, so it shouldn't be possible for this to fail. Is there a situation where that could happen?

cog1to: I just forgot that _getToolPath alredy does that. >.< Reverted

cog1to:
> Not easily, although it would be nice if there was (since it would make implementing focus mode a lot easier). You can listen for events like tab-selected on tabState but there's no way to cancel them; I think you'd have to add a way to do that, and th

Can't we just add a couple of check to the `browserUI`'s methods? something like,
```
function addTab(tabId = null, options = {}) {
  // Early return if we're in restricted mode.
  if (focusMode.enabled() || modalMode.enabled()) {
    return
  }
}
```

cog1to:
What are the expected rules for focus mode and for this new _modal_ mode? I see that some actions are not allowed in focus mode, but not all. Like, we can't create new tabs or tasks, but we can navigate current tab.

For modal mode, I assue we should disable just about everything, including shortcuts, right?

PalmerAL:
Yeah that would probably make the most sense for now (if you do it, I'd vote for placing a semi-transparent overlay over the tab bar rather than hiding it), or I'd also be fine with just leaving it the way it is.

PalmerAL:
While trying to figure out why there was a long delay between clicking the key icon and the passwords appearing, I realized that calling `bw --version` takes around 800ms, so I changed _getToolPath to check if the file exists and is executable (in the local case), or just that the command exists (in the global case). This cuts the total time to retrieve passwords from ~1.6s to ~800ms total.

PalmerAL:
Also, autofill appears to not be working on Google, I'll probably look into it this weekend.

cog1to:
Added overlay for tab bar.

cog1to:
I also noticed a bug on my system with current `_getToolPath` implementation: calling `fs.promises.access` with `S_IXUSR` constant throws with `EINVAL` error. I don't know if it's a Linux thing or just my system thing, but changing that to `X_OK` fixes the issue.

PalmerAL:
The approach for the overlay looks good, although the tab bar appearance is kind of broken:

Screen Shot 2020-01-17 at 5 39 35 PM

We can probably just reuse the existing searchbar overlay here.

PalmerAL:
Fixed:
Screen Shot 2020-01-17 at 5 50 47 PM

PalmerAL:
> I also noticed a bug on my system with current `_getToolPath` implementation: calling `fs.promises.access` with `S_IXUSR` constant throws with `EINVAL` error. I don't know if it's a Linux thing or just my system thing, but changing that to `X_OK` fixes the issue.

Oops, I think I read the docs wrong; should be fixed now.

PalmerAL:
Google login should work now.

js/passwordManager/passwordManager.js
```diff
@@ -0,0 +1,249 @@
...
+  isUnlocked() {
+    return this.sessionKey != null
+  }
+
+  // Tries to get a list of credential suggestions for a given domain name.
+  // If password store is locked, the method will try to unlock it by
+  async getSuggestions(domain) {
+    if (this.lastCall != null) {
+      return this.lastCall
+    }
```
PalmerAL: What's the reason for this?

If I'm understanding correctly, if two websites request passwords at almost the same time, the second website could end up with the passwords for the first website, which would be fairly bad.

js/util/keyMap.js
PalmerAL: 1password uses cmd/ctrl + backslash, would that work?

js/passwordManager/passwordManager.js
cog1to: The idea is to guard from stacking up multiple unlock requests/dialogs in case when user activates the autofill multiple times. By accidently double-clicking on the "fill" icon, for example.

PalmerAL: OK, that makes sense, I think this can be fixed by keeping track of lastCalll separately per-domain then.

PalmerAL:
I simplified the wording on some of the strings, and tried to make them so they can be re-used for future password manager integrations (so they don't need to be re-translated).

I also removed the information about adding the tool to the PATH; if you've been using the tool already, it's almost certainly going to be there already, and I'm guessing at least a decent number of people aren't going to know what that is, which would potentially make it confusing.

PalmerAL:
I think this is almost done! There are only a couple things left to do that I can think of:

 - [ ] Convert the new files to modules and load them from `default.js` instead of including them in the legacyModules list
 - [ ] Decide which shortcut to use
 - [ ] Test this on Windows and make sure everything works.

Is there anything else I'm missing?

cog1to:
I'd like to test/fix it for some sites that track keyboard events on input fields to prevent auto login. Not sure how to approach it right now though, or if it is even possible to fully fix...

PalmerAL:
Do you have any examples?

cog1to:
@PalmerAL I checked again, and it seems that adding additional events to the list of events fired on auto-fill fixed the issue. Sorry for the confusion.

cog1to:
On another note, looks like the changes to the unlock icon positioning broke it for some sites. A short list I have tested it on where it's not visible, although autofill works there via keyboard shortcut:
```
coursera.org
twitch.tv
humblebundle.com
chrono.gg
```

This is probably not critical, since the feature works and it's just a cosmetic issue.

PalmerAL:
That was caused by the z-index not being high enough for the button to be on top of the modal; it should work now. On chrono.gg, the icon is inheriting a color that makes it almost invisible; there's probably a fix for that.

PalmerAL:
I'm good with merging this, unless there's anything else you think should be changed first.

cog1to:
I'd like to try adding a couple of cosmetic improvement before we merge:
1. Hover effect to the suggestions list items.
2. Either force the foreground/background colors for suggestion list, or try to derive that from the input field style.

Both are aimed to improve UX for dark mode and/or dark-styled sites. One quirk I detected is that Twitch in dark mode produces suggestion list that is very hard to read. I'll attach a screenshot when I get back to my main PC.

cog1to:
@PalmerAL I think we're good to merge after your review

PalmerAL:
While looking through the original issue to make sure I hadn't forgotten anything, I realized it would be a good idea to disable autofill for cross-origin frames; if this was allowed, it could be used to trick the user into signing into a site they didn't expect (although I don't have any specific idea of how this could be exploited), and I've yet to see any site do this legitimately.

PalmerAL:
Everything LGTM, I'm going to merge.

I really appreciate the amount of work you've put into this. It's been a fairly long process, and I apologize for delaying this for so long, but I think the end result has turned out really well. Thank you for being so thorough following up with fixes, and for investing so much time in this!

I'm hoping to support more than one password manager when this launches, so I'm going to try to add support for another one (probably 1password) over the next couple weeks, but if that ends up taking too long, I'll go ahead and release the next version with just Bitwarden. This should be released fairly soon either way.

cog1to:
@PalmerAL thank you as well for thorough reviews and myriad of fixes and catches. I'm really glad we were able to bring it to the conclusion :)

PalmerAL:
I'm running into an issue with 1password that makes it impossible to sign in: https://github.com/minbrowser/min/issues/915#issuecomment-592331543, and I haven't worked with `child_process` enough to know if there's a way to fix this. Would you happen to have any ideas?

PR #867, opened 2019-12-03
Add Raspbian(armhf) support, web browser and default browser settings for Debian based OS, update Korean strings.

lee-yunseok:
Add armhf build that works on Raspberry Pi 4 B Raspbian OS.
Update Korean translations.

PR #871, opened 2019-12-09
Add missing Korean translation, fix typos and grammar

toriato:


toriato:
#867 looks like he has already done it but still 'task' should translation to '작업' not a '테스크' which is just pronounciation.

PalmerAL:
Thanks for the translation! I went ahead and merged #867 since it was opened first, but it looks like there are still some differences between this translation and the other one.

@lee-yunseok Do the additional changes in this look reasonable to you? I don't know any Korean myself, so I can't tell which way is more correct.

Issue #882, opened 2020-01-01
Prompt to set Min as default browser breaks automated installs


chepec:
[Min 1.12](https://github.com/minbrowser/min/releases/tag/v1.12.0) comes with this new feature:

> You will now be prompted to set Min as your default browser after installing on Linux (by @lee-yunseok)

This is what happened when I upgraded (Ubuntu 18.04):
```
$ sudo apt full-upgrade 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Calculating upgrade... Done
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
1 not fully installed or removed.
After this operation, 0 B of additional disk space will be used.
Do you want to continue? [Y/n] 
Setting up min (1.12.0-1) ...
Would you like to set Min as your default web browser? [Y/n]
```

This implementation is a very unusual way to do it, and **will break** all kinds of automation built on `apt`. For example, Ansible playbooks will choke and error out on this prompt.

Would you consider accomplishing this feature in a more traditional way, as a popup window issued by the Min GUI, for example?

PalmerAL:
Hmm, that's not good. Assuming there isn't a more standard way to do this after installation (is there?), I'd be fine with just making this a message telling you how to run the command yourself. @lee-yunseok does that sound OK to you?

lee-yunseok:
Sorry for late reply. I didn't expect this issue. Then remove postinst script.

PalmerAL:
I don't think the script needs to be removed entirely; would something like this work?

```bash
#!/bin/sh
set -e

if [ "$1" = "configure" ] || [ "$1" = "abort-upgrade" ]; then
    update-alternatives --install /usr/bin/x-www-browser \
        x-www-browser /usr/bin/min 80
    echo "Run sudo xdg-settings set default-web-browser min.desktop && sudo update-alternatives --config x-www-browser to set Min as your default browser."
fi
```

lee-yunseok:
In these days I'm using RPi only, so I can't test on Ubuntu or other Linux distro.

lee-yunseok:
Please check the pull request #886

christianrauch:
The `min` package should really not ask interactively to become the default browser. This breaks many `apt`-based GUI tools and the dpkg database. Interactive package installation is an annoying thing in any case.

What is wrong with asking the user about making it the default on the first start? This is how Firefox, Chrome and Vivaldi are doing it.

PalmerAL:
[...]

PR #886, opened 2020-01-15
#882 Hotfix

lee-yunseok:
This will prevent postinst_script on **upgrade** but works on **install** step. I've tested it on Raspbian with dpkg and I think this may work on any Debian based **apt upgrade/full-upgrade**.

The version number is what I temporarily modified for testing purpose.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "min",
   "productName": "Min",
   "author": "PalmerAL",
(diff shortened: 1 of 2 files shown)


PalmerAL:
I'm not sure that this fully fixes the issue; the example in #882 (installing across machines using Ansible) would still have this issue I think; but maybe I'm misunderstanding?

PR #890, opened 2020-01-21
WIP Fix this issue : timche/gmail-desktop#174

Ordissimo:


PalmerAL:
Changing the default user agent to something other than Chrome isn't a good idea, because it results in sites sending versions of their code that's intended for other browsers. However, I think it  would make sense to adding a setting for this so that you can override the UA to log in.

Ordissimo:
I've reset my conf and I can't log into the google account anymore.
I'll try to fix that.
Thanks for this nice little project!

Ordissimo:
That's where I'm at.
If I delete the chrome version I can sign up for a Google account.
```
let newUserAgent = defaultUserAgent.replace(/Min\/\S+\s/, '').replace(/Electron\/\S+\s/, '').replace(/Chrome\/\S+\s/, 'Chrome ')
```
After logging in I can restore the chrome version, I don't lose the connection.
Maybe use cookies
```
let newUserAgent = defaultUserAgent.replace(/Min\/\S+\s/, '').replace(/Electron\/\S+\s/, '')
```

Ordissimo:
Just apply this correction to this domain : https://accounts.google.com/

Ordissimo:
I would like to modify the user agent with userscrips but it doesn't seem to work!
```
cat ~/.config/Min/userscripts/accounts.google.com 
// ==UserScript==
// @name        Change navigator.userAgent
// @namespace   Rob W
// @description Changes navigator.userAgent to accounts.google.com
// @match       https://accounts.google.com/*
// @run-at      document-start
// @grant       none
// @version     1
// ==/UserScript==

let userAgentString =  navigator.userAgent.replace(/Chrome\/\S+\s/, 'Chrome '); 

Object.defineProperty(navigator, 'userAgent', {
    value: userAgentString
});
```

Ordissimo:
Resolve for me with: https://github.com/minbrowser/min/pull/893
Thanks Nand2

PR #891, opened 2020-01-21
Update electron

Ordissimo:
Update electron and dependencies.

### DIFF
--- a/main/UASwitcher.js
+++ b/main/UASwitcher.js
@@ -1,6 +1,6 @@
 /* Use the same user agent as Chrome to improve site compatibility and increase fingerprinting resistance
 see https://github.com/minbrowser/min/issues/657 for more information */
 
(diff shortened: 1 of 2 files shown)


PalmerAL:
I usually just update electron before a release (after building it with that version and testing it for a week or so); I don't think there's any reason to update now unless there's a specific fix we need.

I think PDFjs might have a breaking change after the current version (which is why it's set to a specific version), were you able to confirm the PDF viewer works after upgrading?

Ordissimo:
Thank you. Thank you, 
Indeed the Pdf is loaded but the rendering is bad.
I'm sorry.

PalmerAL:
The fix to get the latest version of PDFjs working turned out to fairly simple (56dfe20432e00ba7ccbb908ecbd2cf0e3dd26065) so I went ahead and upgraded it. So thanks for pointing this out.

PR #893, opened 2020-01-24
Fix accounts.google.com login ("This website is not secure")

nand2:
Accounts.google.com seems to use more than the User-Agent for fingerprinting Chrome (other headers, javascript objects, ...), so use a Firefox UA instead.

This is not the perfect solution (the Firefox UA will need manual updating every X years; the website may do bad things thinking it is an actual Firefox), but this is the best I can think of, since using a valid Chrome UserAgent is not enough.

Another approach would require us to find out how accounts.google.com is fingerprinting chrome, with which other header or javascript check... but this would be hard, and it could change any day.

Let me know if you think of another better approach.

### DIFF
--- a/main/UASwitcher.js
+++ b/main/UASwitcher.js
@@ -4,3 +4,20 @@ see https://github.com/minbrowser/min/issues/657 for more information */
 let defaultUserAgent = app.userAgentFallback
 let newUserAgent = defaultUserAgent.replace(/Min\/\S+\s/, '').replace(/Electron\/\S+\s/, '')
 app.userAgentFallback = newUserAgent
+
+
+/* Some websites use more than the User-Agent for fingerprinting, and so the above change is not enough. */
+
+app.once('ready', function () {
+  session.defaultSession.webRequest.onBeforeSendHeaders((details, callback) => {
+    const url = new URL(details.url);
+    // Google login : Will throw a "Your browser is unsecured" and will not permit authentification,
+    // even using a valid Chrome User-Agent
+    // See https://github.com/timche/gmail-desktop/issues/174
+    // Using a Firefox User-Agent works
+    if(url.hostname == 'accounts.google.com') {
+      details.requestHeaders['User-Agent'] = 'Mozilla/5.0 (X11; Linux x86_64; rv:72.0) Gecko/20100101 Firefox/72.0';
+    }
+    callback({ cancel: false, requestHeaders: details.requestHeaders });
+  })
+})
--- a/main/main.js
+++ b/main/main.js
@@ -1,6 +1,7 @@
 const electron = require('electron')
 const fs = require('fs')
 const path = require('path')
+const url = require('url');
 const app = electron.app // Module to control application life.
 const protocol = electron.protocol // Module to control protocol handling
 const BrowserWindow = electron.BrowserWindow // Module to create native browser window.


PalmerAL:
I'm not sure this approach is really a good idea in the long term; if it starts to become commonplace (which it looks like it is), it would be fairly easy for Google to adjust their approach to bypass this. Also, I'm not sure how Google would view changing the user-agent, and I don't want to create any issues with that.

At the same time, I don't have a better solution. I spent some time digging through the sign-in page JS to see if they were using any browser API's that could be fingerprinted easily, and I really don't see anything that would allow them to differentiate between Chrome and other Chromium-based browsers (but they clearly are somehow). It's also strange how this only affects some people - I've tried this with both a G suite and a regular account, and I can't reproduce it. I'm quite puzzled. (But if anyone reading this knows the answer, please let me know!)

I think maybe the best approach is to add a setting to change the user agent so you could set it to whatever you want. (Actually, it's already possible to do this from the devtools, but putting it in the settings page would make it easier to find).

nand2:
Thanks for your answer. Yes indeed this is a thorny issue... This patch is quick and pragmatic, but could stop working at any time...

The better approach would be as you say, a configurable setting (User-Agent override per domain), and then we add an automatic update of this setting (much like adblocks do with their blacklist updates). And if necessary going broader than the User-Agent because of other finger-printing methods.

But this becomes a quite bigger project... 
If this issue keep arising (I need to regularly update User-Agent overrides), I could work on that feature.

PalmerAL:
Another option would be to add a "use firefox user agent" option; when enabled, it would apply to all sites, not just Google. We could guess the Firefox version based on the date (73 was released on 2/11 + 4 weeks per release) to avoid having to update it manually.

nand2:
I am less at ease with this option, because we could end up with randomly broken websites, which use the user agent to choose to activate some browser-specific functionality. Most of the website would be using the javascript interface to detect the browser, but some some will use the user agent (as seen in this case with accounts.google.com)

Ordissimo:
Why not use the userscripts method, already implemented in Min, but this time for the UserAgent.
The modification is up to the user, and depending on the domain, he applies the appropriate UserAgent.

PalmerAL:
I've added a user agent option to the settings page in f427d200cd5fb0ba1249f4cbc289a3c3d4895ca9; I think that should be sufficient to solve this issue for now.

nand2:
Thank you for writing this code -- that's a good option to have, even though I am not at ease with it for this specific problem (changing the user agent globally to fix a website may break another).

I will attempt coding a per-domain override (on top of your newly added override), with config and UI and all, starting next week, and will submit a pull request.

PalmerAL:
I'm not sure having an entire configuration UI for this is worth it; since Google typically keeps sessions valid for a long time, you shouldn't have to sign in very often, so I think it would be fine to enable this when you want to sign in and then turn it off.

Ordissimo:
Often some badly made sites require the user agent to be overridden.
Your solution allows us to change the user agent permanently and for the whole browser.
Your application is multi tabbed.
So a site may need a user agent. The site of another tab may require another user agent.
Offering one solution per domain is more complete.

PalmerAL:
Do you have other examples of sites that don't work with the default user agent?

This isn't really intended to be a fix for all user-agent issues; it's supposed to be something you can turn on for a limited period of time in order to sign in to Google. If there really are widespread issues on different sites that require per-domain customization, we could do that, but I'd prefer not to if we don't have to.

AwesomeTech10:
https://security.googleblog.com/2019/04/better-protection-against-man-in-middle.html

"Because we can’t differentiate between a legitimate sign in and a MITM attack on these platforms, we will be blocking sign-ins from embedded browser frameworks starting in June. This is similar to the restriction on webview sign-ins announced in April 2016."

PalmerAL:
The interesting question here is how Google differentiates embedded frameworks from standard Chrome; I haven't been able to figure that out yet.

AwesomeTech10:
Min uses WebViews, correct? They have already restricted WebViews, and Electron is probably considered an embedded browser. 

Google wants embedded frameworks and WebViews to use OAUTH2.
https://developers.googleblog.com/2016/08/modernizing-oauth-interactions-in-native-apps.html

PalmerAL:
Pages are in a [BrowserView](https://www.electronjs.org/docs/api/browser-view). What's interesting about this is that there shouldn't be any difference between a BrowserView and Chrome that allows webpages to tell where they're being loaded, but there apparently is.

AwesomeTech10:
Here's my take: Browser Fingerprinting. Any website can get a lot of information from a browser. To test this, I tried to fingerprint the browser myself using an EFF website. The biggest difference I noticed was the browser plugins. Sadly, this is not how Google detects browsers. However, it is possible that Google is fingerprinting the browser in a way that I don't know. I then went on to Google's privacy policy to see what they collect. Google itself mentions they use advertising IDs, IMEI numbers, cookies, etc. I then went on to Google's sign-in page and searched for the word navigator. The results were a bunch of navigator.userAgent, navigator.onLine, navigator.bluetooth, navigator.cookieEnabled, navigator.sendBeacon and navigator.standalone. These might be used to fingerprint the browser and detect an embedded systems/Webview/etc.

PalmerAL:
Yeah, I agree that's what they're doing. However, all of those properties _should_ be identical between Min and Chrome, so there shouldn't be anything that allows them to block Min and allow Chrome. If we could find out what the difference is, we could fix whatever's causing that and make signin work again.

EnderDev:
@KraiiFox suggested something in our discord server.

If you include Microsoft Edge's user agent `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.61 Safari/537.36 Edg/83.0.478.37`, it will let you login. _And yes it really is Edg, and not Edge._

This solution is much better than the Firefox UA because it uses a Chromium-based UA, which might help with compatibility. I tried the Firefox UA and most sites didn't accept it.

Here is a video with Google login working:
https://www.youtube.com/watch?v=KZzL9P1Dkx0

This is the code I use and it works like a charm, feel free to use it if you want.
```js
view.webContents.userAgent = view.webContents.userAgent
    .replace(/ YOURAPPNAMEHERE\\?.([^\s]+)/g, '')
    .replace(/ Electron\\?.([^\s]+)/g, '')
    .replace(/Chrome\\?.([^\s]+)/g, `Chrome/85.0.4152.0 Edg/83.0.478.37`)
```

PalmerAL:
Thanks, that is interesting.

If you're willing to change your default user-agent, I think you could use webRequest to modify the user-agent header only for the Google login page, assuming it's a server-side check and not `navigator.userAgent`.

I spent a bit more time trying to figure out how this block works, and I'm even more puzzled than before. As someone on stackoverflow [pointed out](https://stackoverflow.com/a/59673634), an official Chrome build will get blocked as well if you run it with `--remote-debugging-port` enabled. But that's an internal browser feature that webpages shouldn't be able to detect; the only way I can think of to do that is if the page tried to actually connect to the debugger, but there aren't any requests showing in the network console for that.

I seriously wonder if Google sends additional data as part of the sign-in request if you're running an official build that's not shown in the network console (sort of like what they did with [x-client-data](https://9to5google.com/2020/02/06/google-chrome-x-client-data-tracking/)). I think it should be possible to set up Wireshark to capture the traffic coming from Chrome and see if this is the case, but I haven't been able to get it to decrypt the connection correctly.

AwesomeTech10:
@EnderDev found this in the sign-in page:

```(a.xa&&a.da.zt&&a.da.zt(a),a.da.Yz&&a.da.Yz(a))},Yn=function(){if(!(window.chrome&&window.chrome.runtime&&window.chrome.runtime.getManifest&&window.chrome.runtime.getManifest()))throw Error("U");};Wn.prototype.getUrl=function(){return String(this.$.clone())};
_.Zn=function(a,b,c){a.ia||(a.ia=new _.ln);```

Said chrome was injecting code

EnderDev:
It could be possible Chrome communicates with the login with injected APIs, but from my testing it looks like it just uses it as a last resort measure, as long as the UA is Edge's, it seems to ignore the injected API all together.

EnderDev:
@PalmerAL We would probably only apply the Edge UA on `**://*.google.*` to make sure that it doesn't display incorrect information on sites like `whatismybrowser.com`.

AwesomeTech10:
You'd also have to change YouTube.com as Google login makes a few requests to YouTube servers for login

PalmerAL:
@EnderDev `chrome.runtime.getManifest` is undefined in Chrome for me (and it looks like its only defined when the page is loaded from an extension), so I don't think that's it.

AwesomeTech10:
From what I've gathered, Google checks the user agent to detect if the service is known browser (Edge, Firefox, Chrome). If the UA is Chrome or Unknown but very similar to Chrome (like Electron's UAs are), Google uses fingerprinting techniques to try and verify the browser. If the browser is not related to chrome (Firefox) or is a separate browser based on Chromium but is known to Google (Edge, Vivaldi, Brave, etc) then Google does not fingerprint the browser

AwesomeTech10:
Just so you know, this isn't just a problem with Electron-based browsers:

"Now, we are making the Vivaldi browser. It is based on Chromium, an open-source project, led by Google and built on WebKit and KHTML. Using Google’s services should not call for any issues, but sadly, the reality is different. We still have to hide our identity when visiting services such as Google Docs."

PalmerAL:
@nand2 Google sign-in has been one of the biggest complaints on the user survey that's included in 1.15, and no one seems to have tried the user-agent setting, so I think you were right.

I'm planning to update this to use the Edge UA like EnderDev suggested and merge it; I guess we could still keep the UA setting around as a backup.

AwesomeTech10:
Just search for the domainsGoogle.comYouTube.comAnything with .googleAnd apply an Edge UA. Just note, Google doesn't like Edge either.On Jun 25, 2020 8:51 PM, PalmerAL <notifications@github.com> wrote:
@nand2 Google sign-in has been one of the biggest complaints on the user survey that's included in 1.15, and no one seems to have tried the user-agent setting, so I think you were right.
I'm planning to update this to use the Edge UA like EnderDev suggested and merge it; I guess we could still keep the UA setting around as a backup.

motoleon:
The users needs to login google, i get a surprise when i cant login youtube. I cant get my youtube videos, subscriptions, etc.

PalmerAL:
I've opened a new PR based on this one that uses the Edge user agent, and also fixes a couple minor issues: #1119 

If someone could try this out and confirm that you can sign into your account without issues, that would be great.

PalmerAL:
#1119 has been merged, so this should be fixed by default in the next release.

@nand2 thanks for the idea! (And sorry for not doing this sooner, in retrospect it was a better solution than what I came up with originally).

PR #894, opened 2020-01-24
Remove prompts.

lee-yunseok:
Remove prompts from **postinst/prerm** script. Just work for install/remove update-alternatives list. #882

### DIFF
--- a/resources/postinst_script
+++ b/resources/postinst_script
@@ -1,25 +1,7 @@
 #!/bin/sh
 set -e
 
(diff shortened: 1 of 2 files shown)


PalmerAL:
I think it's still fine to show instructions on how to set Min as the default, it just can't stop the installation from completing. I've opened an alternative PR here: #896; would that work?

lee-yunseok:
Then ok, it will work.

PalmerAL:
OK, I'll merge that version then.

PR #900, opened 2020-01-31
Update zh-CN.json for password manager.

lautis0503:


localization/languages/zh-CN.json
```diff
@@ -198,6 +198,29 @@
...
+        /* Password manager setup */
+        "passwordManagerSetupHeading": "完成 %p 中的设置以使用自动填充",
+        "passwordManagerSetupStep1": {
+            "unsafeHTML": "首先, 访问 <a id='password-manager-setup-link'></a> 并解压到你的系统中。"
+        },
+        "passwordManagerSetupLink": "下载 %p 中的CLI工具",
+        "passwordManagerSetupStep2": "然后把工具拖动到下面的框中：",
+        "passwordManagerSetupDragBox": "把工具拖动到这里",
+        "passwordManagerSetupInstalling": "安装中...",
+        "passwordManagerSetupSignIn": "登录密码管理器以使用自动填充。你的凭证只会存储在Min内部。",
```
PalmerAL: Google/Bing translate are showing this as "Your credentials will only be stored inside Min", but the string should be "Your credentials **won't** be stored anywhere inside Min." - can you confirm this is correct?

lautis0503: Fix it.

PalmerAL:
Thanks!

PR #908, opened 2020-02-15
Update Korean strings.

lee-yunseok:
- **Password Manager** has been added.
- **Task** has been shortened.

### DIFF
--- a/localization/languages/ko.json
+++ b/localization/languages/ko.json
@@ -211,7 +211,7 @@
         "passwordManagerSetupStep2": "그 다음 도구를 아래 박스 안에 끌어다 놓습니다:",
         "passwordManagerSetupDragBox": "도구를 여기에 끌어다 놓으세요.",
         "passwordManagerSetupInstalling": "설치하는 중...",
-        "passwordManagerSetupSignIn": "자동완성을 사용하려면 비밀번호 관리자에 로그인하세요. 귀하의 자격증명은 민(Min)의 그  어디에도 저장되지 않습니다.",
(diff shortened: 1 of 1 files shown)


localization/languages/ko.json
```diff
@@ -199,5 +199,28 @@
...
+        /* Password manager setup */
+        "passwordManagerSetupHeading": "자동완성을 사용하려면 %p 설정을 완료하세요.",
+        "passwordManagerSetupStep1": {
+            "unsafeHTML": "우선, <a id='password-manager-setup-link'></a>를 방문해 시스템에 내려받으세요."
+        },
+        "passwordManagerSetupLink": "%p CLI 도구를 내려받으세요."
+        "passwordManagerSetupStep2": "그 다음 도구를 아래 박스 안에 끌어다 놓습니다:",
+        "passwordManagerSetupDragBox": "도구를 여기에 끌어다 놓으세요.",
+        "passwordManagerSetupInstalling": "설치하는 중...",
+        "passwordManagerSetupSignIn": "자동완성을 사용하려면 비밀번호 관리자에 로그인하세요. 귀하의 자격증명은 어디에도 저장되지 않습니다.",
```
PalmerAL: Google Translate (which I realize isn't always right) is translating this as "Your credentials are not stored anywhere", which isn't accurate - Min itself doesn't save the credentials, but they're sent to your password manager, which could potentially save them (although it probably won't). Could you please confirm that the translation only refers to what Min will do, and not the credentials in general?

lee-yunseok: Done.

PalmerAL: Thank you!

PR #910, opened 2020-02-22
Update tabBar.js

idAlencar:
the idea is to show where people are researching

### DIFF
--- a/js/navbar/tabBar.js
+++ b/js/navbar/tabBar.js
@@ -152,10 +152,11 @@ window.tabBar = {
 
     var ec = document.createElement('div')
     ec.className = 'tab-edit-contents'
-
+    
+    var searchEngine = require('util/searchEngine.js')
     var input = document.createElement('input')
     input.className = 'tab-input mousetrap'
-    input.setAttribute('placeholder', l('searchbarPlaceholder'))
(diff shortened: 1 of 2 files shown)


PalmerAL:
I'm not sure I understand the reason for this. I'm assuming that most people change their search engine rarely (or never), and so you would most likely already know where your search is going to go. Is there another scenario I'm missing?

idAlencar:
most people just open the browser and search. I work daily with people who barely know what a browser is, I realized that google chrome became popular because it had the search service name in the browser name. the most talked about phrase about navigation is "open a google". the idea is that the person will see in the first contact where he will search

idAlencar:
the goal is to display the message "search on (search server) or type an address" in the bar, making it clear in the first contact where it will be searched

PalmerAL:
I think most people who don't know what a browser is won't be using Min, so that's less of a concern for us than it is for Chrome. Given that, and the fact that this would require re-translating this string, I'm inclined not to do it. If we get more feedback in the future that this is an issue, I'd be happy to reconsider it.

idAlencar:
I understand your side, I just want to leave the idea and the path that many projects sinned, develop for developers or who knows, the big projects (windows, mac, chrome, firefox, office etc.) were great because they reached simple people (they don't know a browser). this I would like to see in Min, a browser for anyone to install and use

PR #913, opened 2020-02-24
Adjustment of icons and effects on the bar

idAlencar:


idAlencar:
Windows version
win


Linux version
linux

PalmerAL:
I agree some kind of hover effect on the buttons would be nice. I'm still thinking about whether this is the best option or not; I'll try to get back to this in a few days.

idAlencar:
I tried to follow the color pattern making it uniform, it already has red and gray, I didn't want to insert more colors.

PR #922, opened 2020-03-06
Update ru.json

KagamineP:


PalmerAL:
Looks good, thanks!

It looks like I forgot to copy the [new strings for the password manager](https://github.com/minbrowser/min/blob/master/localization/languages/en-US.json#L205-L227) into the other language files; would you be willing to translate those as well?

KagamineP:
> Looks good, thanks!
> 
> It looks like I forgot to copy the [new strings for the password manager](https://github.com/minbrowser/min/blob/master/localization/languages/en-US.json#L205-L227) into the other language files; would you be willing to translate those as well?

Added.

PR #927, opened 2020-03-09
Update pt-PT.json

smarquespt:

PR #937, opened 2020-03-16
Update pt-BR.json

idAlencar:


localization/languages/pt-BR.json
```diff
@@ -5,7 +5,7 @@
         /* Context menu items
         these are the items displayed in the menu when you right-click on a page
         */
-        "pictureInPicture": null, //missing translation
+        "pictureInPicture": "Picture in picture", //missing translation
```
PalmerAL: I had the same question on #927 actually: is there not a Portuguese translation for this?

PalmerAL:
Could you please also translate the "enableBlocking" and "disableBlocking" strings?

idAlencar:
we use it that way (Picture in picture), the translation gives us another idea -> picture can mean (image, figure, photo), fica algo como (image in image).

"enableBlocking" and "disableBlocking" -> went unnoticed, but I will kkkkkkk

PalmerAL:
Thanks for the explanation!

Issue #778, opened 2019-07-03
SUID sandbox doesn't work correctly


eltonfabricio10:
[17241:0702/225020.308126:FATAL:setuid_sandbox_host.cc(157)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /usr/lib/min/chrome-sandbox is owned by root and has mode 4755.
Trace/breakpoint trap

PalmerAL:
This is https://github.com/electron/electron/issues/17972; it looks like we need to upgrade the version of electron-installer-debian we're using in order to get this change: https://github.com/electron-userland/electron-installer-debian/pull/184.

In the meantime, you can run without sandboxing by running `min --no-sandbox`, although that is somewhat less secure than having sandboxing enabled.

Yusef-Aslam:
This still happens on Deepin 20 Beta, which is based on Debian 10 stable

EchedeyLR:
Still happens on last stable version in Devuan 3 (Debian 10).

PR #942, opened 2020-03-19
Solution for sandbox bug in Debian based distro

eltonfabricio10:
https://github.com/minbrowser/min/issues/778

### DIFF
--- a/resources/postinst_script
+++ b/resources/postinst_script
@@ -4,5 +4,7 @@ set -e
 if [ "$1" = "configure" ] || [ "$1" = "abort-upgrade" ]; then
     update-alternatives --install /usr/bin/x-www-browser \
         x-www-browser /usr/bin/min 80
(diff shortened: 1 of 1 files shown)


PalmerAL:
I think this should have been fixed already by 99c406dbf862b5fa235bbba0af751c4e752624e7 (so for all releases >=1.11.1); are you still seeing this issue in the latest version?

eltonfabricio10:
Yes, it still has the problem in the latest version.

EchedeyLR:
@PalmerAL is there any update for this to be merged?

PalmerAL:
Yeah, I guess it's fine to merge this. The package generator [already does this same thing](https://github.com/electron-userland/electron-installer-common/blob/bd3f8cafff92e7bcbdb8ca4bccd9ce2c07911613/src/sandboxhelper.js#L33), so I'm confused why this is necessary, but I guess it's probably not worth the time to investigate.

Issue #946, opened 2020-03-28
Building from git master fails after https://github.com/minbrowser/min/commit/545cf5a177e1bde0cbfda467e54f28594b86636e


glitsj16:
Hi, I've been building min for a while now from git master. After https://github.com/minbrowser/min/commit/545cf5a177e1bde0cbfda467e54f28594b86636e I saw the below output:

```
$ npm install
...

$ npm run buildDebian
...
Error while building: Cannot find module 'readerView.js' from '/home/glitsj16/min/js/searchbar/placesPlugin.js'
Error while building: Cannot find module 'readerView.js' from '/home/glitsj16/min/js/navbar/tabBar.js'
Error while building: Cannot find module 'readerView.js' from '/home/glitsj16/min/js/menuRenderer.js'
```

After renaming `readerview.js` to `readerView.js` (note the camelcase pattern) the above errors disappear. Haven't gotten around yet to extensively test the browser, but it seems to be working just fine with that change. More a FYI I guess.

PR #947, opened 2020-03-28
Rename readerview.js to readerView.js

glitsj16:
Fixes building from git master --> https://github.com/minbrowser/min/issues/946


PalmerAL:
Thanks for noticing this! I'm not entirely sure I understand what's happening here; the file is already renamed to be camelcase on my local copy of the repository:



and git says it's in sync with the remote copy:


So I'm not sure why it would be wrong here. Maybe because I'm using a case-insensitive file system?

But I suppose it's fine to merge this.

PR #951, opened 2020-03-31
Update pl.json

JB1905:


PalmerAL:
Thanks! Can you also add the strings for the password manager (https://github.com/minbrowser/min/blob/master/localization/languages/en-US.json#L205-L227)?

JB1905:
> Thanks! Can you also add the strings for the password manager (https://github.com/minbrowser/min/blob/master/localization/languages/en-US.json#L205-L227)?

Done 😃

PalmerAL:
Thanks.

PR #963, opened 2020-04-08
Custom window controls on Linux

PalmerAL:


idAlencar:
hi @PalmerAL  how are you?
my experience with the electron is small, i will ask a question.
The control function could not be modified only:

default.js
```

if (navigator.platform === 'MacIntel') {
document.body.classList.add ('mac')
window.platformType = 'mac'
} else if (navigator.platform === **'Win32' || 'linux'**) {
document.body.classList.add ('windows')
window.platformType = 'windows'
}
```


windowsCaptionButtons.js

```
if (navigator.platform === 'Win32' || 'linux') {
updateCaptionButtons ()
```


main.js
```

minWidth: (process.platform === 'win32'? 400: 320), // controls take up more horizontal space on Windows
minHeight: 350,
titleBarStyle: 'hiddenInset',
icon: __dirname + '/icons/icon256.png',
frame: process.platform === 'darwin' || settings.get ('useSeparateTitlebar') === true,

```

so that the main index.html can be used
`<div class = "windows-caption-buttons">`

**I have this vision**
test


**with merged #963 , I see this**

atual

PalmerAL:
Pretty good!

Just to make sure, you're running Linux, but you want to use the Windows buttons?

`navigator.platform` has more information than just "linux" in it (you can try running it in the console to find out for sure). You probably want to use this instead: `navigator.platform.includes('Linux')`.

idAlencar:
I use manjaro, really like the appearance of the Windows version, think it is well positioned in the browser.
following this path, it would not be simpler to create o Min?

PalmerAL:
I'm not sure it makes sense to do that as the default, since I think the Windows buttons don't fit in very well on Linux, but like you mentioned, it's possible to use the Windows buttons if you want to.

Make these changes to default.js:
```
if (navigator.platform === 'MacIntel') {
  document.body.classList.add('mac')
  window.platformType = 'mac'
} else if (navigator.platform === 'Win32' || navigator.platform.includes('Linux')) {
  document.body.classList.add('windows')
  window.platformType = 'windows'
} else {
  document.body.classList.add('linux')
  window.platformType = 'linux'
}
```

If you're using the current master branch, windowsCaptionButtons.js has been replaced by windowControls.js, and it should work automatically once you change the platformType in default.js.

idAlencar:
the modification in default.js worked perfectly.

idAlencar:
As for the #963 merged, I noticed that, when maximizing the screen, the space (-control-space-top: 12px;) at the top is not removed, in the images I sent you must have noticed what I'm talking about.

what do you think of putting in the settings so that the user can choose the appearance he would like to use?

something like:

appearance
Style: Windows or Linux

PalmerAL:
I agree it probably makes sense to remove the drag space when maximized; that originally made sense with a 1-button design, but it doesn't with a 3-button one.

I'm not convinced that having an option to use the Windows buttons is necessary, since I'm expecting most people will be fine with either one, and if you do want them it's possible to modify default.js.

PalmerAL:
I've changed the drag space behavior in 9f66fcc2a3144751891f67bc33cc277b4fd34a8a.

idAlencar:
it was very good!!!!

PalmerAL:
I'm not sure what you mean?

idAlencar:
sorry, i expressed myself wrong.

9f66fcc, is function well, very good

PalmerAL:
Thanks!

Issue #174, opened 2016-05-20
Min does not support HTTP authentication


plrenaudin:
It directly goes 401 without displaying a popup for login

wb9688:
Do you mean that Min doesn't support HTTP Authentication?

plrenaudin:
Yes, in this website for example: http://www.pagetutor.com/keeper/http_authentication/
Click on "My Secret Stuff" link.

ChristiaanLock:
Related to #142

wb9688:
No, I don't think so. I think he's refering to [HTTP basic access
authentication](https://en.m.wikipedia.org/wiki/Basic_access_authentication)
and not to not having support for `window.opener`…

ChristiaanLock:
I think #142 has nothing to do with 'window.opener' and that he is talking about the HTTP basic auth login form, not an actual popup window. Browsers should show the login form automatically when the servers sends a 'WWW-Authenticate' header. 
Therefore, imho, the issues described in #174 and #142 are related.

wb9688:
Indeed, but I accidentally pressed the #140 link in #142, so I thought you
said that it's related to #140…

talentlessguy:
Looks like it is still an issue, neither Github auth, nor Google login work

PalmerAL:
@talentlessguy Google login is different from HTTP auth, can you clarify what isn't working for you?

talentlessguy:
@PalmerAL "Login with Google" doesn't work. In Figma at least. Looks like tabs can't exchange information.

PalmerAL:
@talentlessguy That's probably #140..

PalmerAL:
This has been fixed in #969.

PR #969, opened 2020-04-17
Added a login prompt for basic http authentication.

NikSays:
Basic HTTP authentication is not very popular but often used in routers and other local network devices
Min had no prompt when site asked for authentication, so server always returned 401 (#174)
I added this using "prompt" event


NikSays:
I just realized prompt title needs to be localized too. I will add this in next commit

PalmerAL:
Thanks for working on this!

You don't actually need to create a new page for the dialog; instead, you should be able to do this:
```
createPrompt({
 text: ...,
 values: ...
}, function (result) {
    callback(result.login, result.password)
})
```

And the code in `prompt.js` will create the popup window and return the results for you.

PalmerAL:
For the dialog strings, I would suggest:

* "Sign in to {site}" for the title
* "username" instead of "login"

NikSays:
> you should be able to do this: createPrompt()
Thank you, this made everything much easier. I thought I needed to send an event from renderer.

PalmerAL:
Everything looks good, thank you!

PR #971, opened 2020-04-22
Add 1Password Integration

### LINKED ISSUE #915: 1Password Integration
Using the [CLI](https://support.1password.com/command-line-getting-started/).

PalmerAL:
Fixes #915.

@cog1to Do you have any thoughts on this? I ended up rearranging the code a bit to share more between the two managers, but the general approach is still the same. On macOS, 1password uses an installer rather than just having a downloadable executable, so I added an alternate mode to the setup dialog that watches for the installer to complete and then shows the sign-in dialog.


js/passwordManager/managerSetup.js
```diff
@@ -0,0 +1,189 @@
...
+  let data = ipcRenderer.sendSync('prompt', {
+    text: l('passwordManagerSetupSignIn'),
+    values: signInFields,
+    ok: l('dialogConfirmButton'),
+    cancel: l('dialogSkipButton'),
+    width: 500,
+    height: 220
+  })
+
+  if (!data.email || !data.password) {
```
cog1to, line 162: probably better to iterate through `signInFields` here

js/autofillSetup.js
```diff
@@ -10,13 +10,12 @@ const AutofillSetup = {
...
-    }).then((result) => {
-      const { manager, configured } = result
-      if (manager && !configured) {
-        if (manager.name === 'Bitwarden') {
-          showBitwardenDialog()
-        }
+      let setupMethod = 'dragdrop'
+      if (platformType === 'mac' && manager.name === '1Password') {
+        setupMethod = 'installer'
+      }
```
cog1to, line 16: since `setupMethod` is determined by the manager type, I think it's better to move this logic to the manager classes, so they can tell us how to setup them. something like `manager.getSetupMethod(platformType)`

js/passwordManager/onePassword.js
```diff
@@ -0,0 +1,180 @@
...
+  }
+  
+  getFileName() {
+    return (platformType === 'windows' ? 'op.exe' : 'op')
+  }
+
+  // Returns a 1Password-CLI tool path by checking possible locations.
+  // First it checks if the tool was installed for Min specifically by
+  // by checking the settings value. If that is not set or doesn't point
+  // to a valid executable, it check the if 'bw' is available globally.
```
cog1to, line 33: minor copypasta: `bw` instead of `op`

cog1to:
Looks good! Left a couple of comments, but overall I like the approach, given the restrictions we have to work with.

Have you tried to spawn/execute the MacOS installer right from the setup code? If that's possible, I guess it would eliminate the need to poll the `isConfigured` method every 2 seconds.

PalmerAL:
I think I've addressed everything, let me know if you see anything else.

There is a way to run the installer programmatically, but it has to be run as an administrator, and I haven't found an easy way to do that in Electron.

cog1to:
that's a bummer... the only way to somehow improve the experience I can think of, is to drag-and-drop PKG file and spawn `open <filePath>` on that file to auto-launch the installer. the process returns immediately, but at least it keeps the flow inside the browser, kind of.

Also tried to spawn `osascript` with root priveleges, but no luck so far.

PalmerAL:
I don't think the current flow is actually that bad either; if there's a simple way to make it easier, that'd be great, but if it ends up being too complicated to do that I'd be fine with leaving it as-is.

cog1to:
Added an implementation in https://github.com/minbrowser/min/pull/979. Up to you if you want to merge it, I'm totally ok with current flow as well.

PalmerAL:
Thanks for reviewing this (and the installer fix)!

cog1to:
I'm glad to help. Thank you for making Min :)

PR #975, opened 2020-04-24
Update Korean strings.

lee-yunseok:
Update missing Korean strings.


PalmerAL:
Thanks! I just added a couple of new strings for the 1password integration: https://github.com/minbrowser/min/pull/971/files#diff-33a5d5b4932353be86d9c603f8c25d25 , could you translate those as well?

PR #976, opened 2020-04-24
Missing translation

avazkhan2808:


PalmerAL:
Thanks! Could you translate the ones I just added here as well: https://github.com/minbrowser/min/pull/971/files#diff-33a5d5b4932353be86d9c603f8c25d25 ?

PR #977, opened 2020-04-24
Update pt-BR.json

idAlencar:


localization/languages/pt-BR.json
```diff
@@ -224,7 +224,7 @@
         "loginPromptTitle": null, //missing translation
```
idAlencar, line 224: I didn't quite understand that part, could you explain it to me

idAlencar:
hi @PalmerAL ,new translations and minor gender corrections

localization/languages/pt-BR.json
```diff
@@ -138,9 +138,9 @@
         "settingsHistoryButtonToggle": "Habilitar botão voltar",
         "settingsAdditionalFeaturesHeading": "Recursos adicionais",
         "settingsUserscriptsToggle": "Habilitar scripts do usuário",
-        "settingsSeparateTitlebarToggle": null, //missing translation
+        "settingsSeparateTitlebarToggle": "Habilitar modo compacto",
```
PalmerAL, line 141: I'm not sure this matches the meaning of the setting. When the setting is disabled, the buttons are all in one line:



When it's enabled, the buttons and tabs are on two separate lines:



This translation would mean the opposite if I'm understanding it correctly.

localization/languages/pt-BR.json
PalmerAL: The original string is this:
```
        "loginPromptTitle": "Sign in to %h (%r)", //%h is replaced with host, %r with realm (title of protected part of site)
```

It's displayed when signing into sites that use HTTP authentication:
Screen Shot 2020-04-24 at 6 33 17 PM

idAlencar: I still don't know how to make it friendly, for now it can be (log into or log in -> **Logar em**) , what do you think?, does it make sense to you?

localization/languages/pt-BR.json
idAlencar: Confused kkkk, **I still don't know how to solve this item, what do you think of** (Disable compact mode) for now?

**Second google translator**

localization/languages/pt-BR.json
PalmerAL: I don't know how to answer the grammar question, but I guess it would be the same as what other login forms do. If you go here: https://github.com/login, there's a heading that says "sign in to GitHub" - how is that translated?

localization/languages/pt-BR.json
PalmerAL: The meaning of "use separate titlebar" is that when it's enabled, the title bar is separated from the tab bar - is it possible to translate it like that?

idAlencar:
Looks good

PalmerAL:
Thanks!

PR #979, opened 2020-04-24
Implement auto-launching installer for password tool.

cog1to:
Tested on OSX with a privileged user. Need to check with a normal user before merging.


PalmerAL:
I like it! The installer doesn't need to be opened as an administrator, so it should work regardless of the account type.

I moved the timeout creation to when the dialog is opened so that everything will still work if you run the installer manually instead of dropping it.

cog1to:
Makes sense, thanks.

PR #983, opened 2020-04-25
Missing translations for Russian and Uzbek

avazkhan2808:
```git diff``` is showing a lot of changes for Russian even though they are identical. I only fixed missing translations for Russian. I guess this is an indent issue with my editor.


PalmerAL:
Your editor converted the space indents to tabs; since most of the files are using space indentation already, can you switch it back?

You and @KagamineP opened PR's for the Russian translation at the same time; it looks like your translations are pretty similar, but a couple of them are slightly different. Could you look over each other's changes and let me know which version to use?

avazkhan2808:
Yeah, the other person's translations are better than mine. Sorry for crappy PR...

PalmerAL:
OK, thanks for checking. I've merged the Uzbek change in 2400c92ec63c8a4108ab632506e8c3c75424e1f6.

I'm not sure if the Electron issue for supporting Uzbek is fixable or not, but if it's not, we can probably add our own language setting to the preferences page that would override the OS setting.

PR #994, opened 2020-05-01
Webpack, start of big refactoring

RealPeha:
[Google Translate]
I recently found out about your project and became very interested in it. I especially liked its minimalistic design.

I wanted to add some new features for tabs (for example, the ability to pin them), but after a quick study of the source code I realized that it would be very difficult to figure it out. So I decided to start with refactoring)

Worked on the main process, here is what is ready at the moment:
- added electron-webpack to build the project
- added eslint standard configuration
- rewrote functions on arrow functions (not 100%)
- removed the use of var (not 100%)
- ecmascript modules (import / export)

Package.json commands:
- beta: compileWin - build a project using electron-webpack
- beta: run - run the project

I think this should speed up the build, improve the quality and performance of the code and simplify its support. What about the idea of ​​implementing webpack?


PalmerAL:
Thanks for working on this!

* I like the idea of splitting the main process code into modules.
* I've been gradually converting the UI process code into modules using Browserify; is there a significant benefit to using Webpack? If there is, we should convert both; if there's not, it'd probably be easier to keep using Browserify.
* Adding a `run` command is a good idea!
* I don't like arrow functions very much; they're nice for 1-line functions, but otherwise I feel like they're harder to read. Unless there's a benefit to them I'm missing, I think I'd prefer to stick with the traditional syntax (I guess not affecting `this` is nice, but I don't think there's anywhere in the code where that would make a difference).
* var to let/const is fine.
* It might be nice to refactor the code so that `global.js` isn't necessary (although that's probably a larger project).

RealPeha:
I decided to use webpack for one most important reason: this is the use of modules and import / export. For me, the biggest problem in the current code is variables declared in one file, which are used in others without their explicit import. Therefore, if I saw a variable that was not declared in the current file, my IDE could not tell what was in it and I had to look for it manually. This increases the development process if you don’t know all the project code

I had never written in electron before, but I saw that there is an electron-webpack and electron-builder package and thought, why did this project have its own scripts for assembly?

I also did not like the use of global.js) But to remove its use, i will have to rewrite the current code a bit. At the moment, I did not do this, but just decided to add this temporary file

PalmerAL:
I definitely agree importing modules would be better than just combining all the files together. What I'm wondering is I've been working to do that in the UI code already using Browserify. For example here: https://github.com/minbrowser/min/blob/master/js/navbar/tabBar.js#L1-L8 (there are still some global variables because not all the files have been converted yet). Is there an advantage to using Webpack versus importing the modules that way?

RealPeha:
I can’t give an exact answer about what will be better. But in my opinion webpack can make a better build, for example using features such as [Tree Shaking](https://webpack.js.org/guides/tree-shaking/) and [Code Splitting](https://webpack.js.org/guides/code -splitting/)
But I really don't know how useful this is for electron

PalmerAL:
I doubt it would actually change that much, because I don't think we have a lot of large dependencies where things could be eliminated. File size doesn't matter quite as much for Electron either, although loading less code might reduce memory usage a bit, which is good.

I guess it might be interesting to do a comparison between webpack/browserify/rollup/parcel/some other bundlers, but browserify has been working well enough for now that I don't see a big reason to switch.

Milo123459:
@PalmerAL Is anything still happening here?

PalmerAL:
Not really. Some of this (switching to let/const, adding a `start` command) has been implemented separately since this was opened; the rest of it would need to be rebased if we wanted to continue with it.

Refactoring the main process code would still be good to do, although I'd probably stick with Browserify rather than switching to Webpack.

PR #1002, opened 2020-05-06
Switchable auto dark mode and new windows icon

labtorie:
Greetings! I do like using this browser but I've found some reasons to stay with Chrome. That's sad, so I decided to participate in the development. Trying my best :)

I've added the option to switch the automatic dark mode:
image

And also normalized the icon and now it looks way better 
image

Hope you'll like it!


PalmerAL:
Thanks for contributing, these are good ideas!

For the dark mode setting, it seems confusing to have a separate checkbox for the automatic mode when there's already an "enable dark mode" option. Instead, I think we could replace both of these with a single setting that has three options: "Enable dark mode: [always / at night / never]". This could either be a dropdown (like the search engine setting) or 3 radio buttons (like the content blocking setting.

I agree the icons should probably be bigger. Resizing the existing 256px icon makes it a bit blurry, so I'm going to re-render it from the original source file instead (and also add the source file to the repository).

PalmerAL:
Also, can you format everything using standard-format?

PalmerAL:
I made an alternate version of the icons in #1003. I left a little bit of space around the edges so the shadow wouldn't get clipped (and I think it's still good to include that), but they should still be larger than before.

labtorie:
Thanks for your answers! 
1) Dark mode. Yes, indeed, three RButtons makes more sense. I'll fix that. 
2) Icons. Windows renders it's own shadow around the icon, so we don't have to. Moreover, Windows guidelines tell us that the icon must be normalized to the regular icon size. I still like it
image
But I've met trouble with taskbar icons: they look pixelated after installing min. Looking for the solution.
3) Code formating. Sorry for that. I've accidentally reformated it with VSCode and fell asleep. Gonna fix it too.

labtorie:
Icon PSD
[min_ico.zip](https://github.com/minbrowser/min/files/4593490/min_ico.zip)

labtorie:
Hello again! Here's what changed:
1) Dark mode switches like this now 
image
I've decided to use radiobuttons. Thus you can change the option with one click instead of two in case we would use dropdown list.
2) Icon is great now, IMO. I've made different border / filler ratio to make it look good in any size:
image
Downscaling:
imageimageimage
Taskbar view:
Small:
image
Normal:
image

Hope you'll like it!

PalmerAL:
That is a better UI for the settings! I think it would make sense to have the settings code match that internally - instead of separate `darkMode` and `autoDarkMode` settings, we could have a single `darkMode` setting with 3 values - `-1` for never, `0` for auto, or `1` for always. That would require updating the existing code that uses this setting - I'd be willing to do that if you want.

The ICO size looks good, although the inside circle has a different color now in this version (compare [original colors](https://raw.githubusercontent.com/minbrowser/min/93eca2d0dc5f38c292320a8f0debf416e18d0017/icons/icon256.ico) to [this version](https://raw.githubusercontent.com/minbrowser/min/856c36bd458ba79db5d0d7b05ca31bac1277f78d/icons/icon256.ico)). The PNG file looks a bit pixelated around the edges, and it looks like the edge of the circle is cut off a little bit on the sides. If you don't have Pixelmator, it'll probably be easier for me to just export a new one that matches the size of this one, if you want me to do that. Also, we may not need to update the PNG file at all - it's only used on Linux, and I think it looks OK there already.

Do you have the icons with a different ratio to try? The images in the PR look like they're the same ratio as the old ones. It does seem like that might help at smaller sizes, although I'm not sure if there's a way to adjust the ICO file based on where it's being displayed or not.

labtorie:
Indeed, using just one `darkMode` variable would make more sense. Seems pretty simple, I will try.

Yes, probably I've messed up the colors. I've just repeated your gradients with starting and ending colors, so maybe that's why it doesn't match in the middle. I'll try again.

And yes, .ico file contains bmp's or whatever with different resolutions. Windows automatically picks one based on the size it needs.
image
Here is one sample .ico file and you can see that it contains different .bmp's. If you try to press `ctrl`+`alt`+`+` or `-` on your windows desktop you'll see that icons changes according to the container size. So, I think different ratios make sense.

labtorie:
Well, I've restored the original icons and I'm going to open a new pull request to offer mine

labtorie:
Here is PSD of the new icon.
[min_logos.zip](https://github.com/minbrowser/min/files/4606702/min_logos.zip)

PalmerAL:
It's interesting that ICOs can do that, I didn't know that. I do agree a variable ratio would be nice. I noticed a couple minor issues (the ratio seems slightly too large at small resolutions, and there's some pixelation around the edges at larger sizes), and I'd like to keep the file in Pixelmator since I don't have Photoshop, so I'm going to try making another version of this and see if I can fix it. MacOS does something similar with its icon files, so once we have the images for each size, we should be able to update the .icns file as well.

The dark mode code is looking a lot better! I did notice a couple of issues with the settings checks. For example, this line in tabColor.js to check whether the colors should darken:
```
     isAuto = (value !== undefined && value === 0) 
```

will only be true if the setting was modified after this PR, and set to "auto". We actually want it to be true if:

1. The setting was never modified (since "auto" is the default) --> it will have a value of "undefined".
2. The setting was modified before this PR --> it will have a value of "true".
3. The setting was modified after this PR, and set to either "auto" or "always".

So I think it should be this:
```
     isAuto = (value === undefined || value === true || value >= 0)
```

The other checks would need similar changes.

js/navbar/tabColor.js
```diff
@@ -125,10 +125,18 @@ function isLowContrast (color) {
 function adjustColorForTheme (color) {
   // dim the colors late at night or early in the morning, or when dark mode is enabled
   let colorChange = 1
-  if (hours > 20) {
-    colorChange = 1.01 / (1 + 0.9 * Math.pow(Math.E, 1.5 * (hours - 22.75)))
-  } else if (hours < 6.5) {
-    colorChange = 1.04 / (1 + 0.9 * Math.pow(Math.E, -2 * (hours - 5)))
+  let isAuto = false
+
+  settings.listen('darkMode', function(value){
```
PalmerAL, line 130: Since this code only cares about the current value of the setting, and doesn't need to listen for changes, this can just be `settings.get('darkMode')`.

PalmerAL:
Also, what tool did you use to make the ICO file?

js/navbar/tabColor.js
labtorie: Oh, right. Now I've got the idea how it works :) Thanks. Going to change it ASAP

labtorie:
> Also, what tool did you use to make the ICO file?

I use [this](https://redketchup.io/icon-editor) online tool for that.

labtorie:
By the way, I also had a problem with the icon being pixelated. I still don't have the idea why it happens. The installator icon was smooth, unlike the app's one:
image
Some kind of mystery. 

I tried to catch the problem and build the app with this icon:
image
The number shows which size of bmp is used right know. And the most interesting thing is that there still were min's icons. I think maybe windows has some kind of cache for them. And I have no idea why did it ignore the new one. I don't remember what exaxtly did I do, but looks like somehow I managed to wipe that cache, idk. Long story short, now I have the icon smooth. 

Moreover, I tried to install my build to another PC which has never had the app installed on it. And the icon was not pixelated. So, maybe windows really has something like a cache for icons. I need to find it out.

Sorry if my bad English made the story unreadable :) I've tried my best

PalmerAL:
Windows definitely caches icons; I remember having the same problem when I made the icon originally. I think I ended up putting the app inside a folder (changing the path makes windows load the icons again), or it might have just fixed itself after a while.

> Sorry if my bad English made the story unreadable :) I've tried my best

It's fine!

PalmerAL:
This works for clearing the icon cache too: https://www.windowscentral.com/how-reset-icon-cache-database-windows-10

labtorie:
Since I've made the .ico with all the resolutions that the OS uses I think we can't do anything about the temporary pixelization. There's an [article](https://www.howtogeek.com/232779/how-to-rebuild-a-broken-icon-cache-in-windows-10/) about the cache. I've followed the guide and now I have this:
imageimage

Now I see that the ratio is too big indeed. But there's no pixelization at all. So, I guess, the final thing to do is to recreate the icon choosing the ratio wisely. Yes?

PalmerAL:
Here's my attempt at making an icon: 
[icon256.ico.zip](https://github.com/minbrowser/min/files/4611982/icon256.ico.zip)

I think it's pretty close to being right. Some screenshots:

Screen Shot 2020-05-11 at 3 46 14 PM
Screen Shot 2020-05-11 at 3 46 21 PM
Screen Shot 2020-05-11 at 3 47 45 PM

Virtualbox seems to make everything pretty fuzzy for some reason (even though it's configured to use high-dpi), but hopefully they look OK on a high-resolution screen as well.

labtorie:
The one in taskbar still are not normalized and is noticeably smaller than chrome's one

labtorie:
Or wait, they have the same size. Looks like an optical illusion. Maybe that's because of outline proportion. Weird thing

labtorie:
Seems like our brains perceive the outline as an icon's border, limiter, idk. And we don't see it's whole shape, but it's guideline (I don't know how to explain it correctly even in my native language). But here's my idea:
image

I think we see the black line as the icons' edge.

PalmerAL:
I think you're right about the border not being visible, and that increasing the border width might help. At least on macOS, most icons do have some kind of border (like Itunes for example), and so it seems like it makes more sense for the blue part to be the focus. A version with a thicker border (I don't think this is 70/30, but it's close) seems kind of odd because it doesn't match the other icons. However, it does seem to stand out more in the dock, so I guess maybe it is better.

Screen Shot 2020-05-12 at 3 28 02 PM
Screen Shot 2020-05-12 at 3 27 53 PM

There's also a bit of empty space left around the ICO edges; we could try making the icon extend all the way to the edge (so that the size of the blue circle alone is closer to the size of the other icons).

js/navbar/tabColor.js
```diff
@@ -125,10 +125,21 @@ function isLowContrast (color) {
 function adjustColorForTheme (color) {
   // dim the colors late at night or early in the morning, or when dark mode is enabled
   let colorChange = 1
-  if (hours > 20) {
-    colorChange = 1.01 / (1 + 0.9 * Math.pow(Math.E, 1.5 * (hours - 22.75)))
-  } else if (hours < 6.5) {
-    colorChange = 1.04 / (1 + 0.9 * Math.pow(Math.E, -2 * (hours - 5)))
+  let isAuto = false
+
+  settings.get('darkMode', function(value){
```
PalmerAL, line 130: `settings.get ` is synchronous when called from the browser UI, so it won't have a callback.

js/util/theme.js
```diff
@@ -4,7 +4,13 @@ if (typeof require !== 'undefined') {
 
 function shouldEnableDarkMode () {
   var hours = new Date().getHours()
-  return hours > 21 || hours < 6
+  let isAuto = false
+
+  settings.listen('darkMode', function (value){
+    isAuto = (value === undefined || value === true || value >= 0)
+  })
```
PalmerAL, line 11: The setting is already checked in initialize(), which is the only place that calls this function, so it should be fine to remove this. It would also probably make sense to rename this function to something like `isNight()` if you want to do that.

pages/settings/settings.js
```diff
@@ -150,13 +149,35 @@ for (var contentType in contentTypes) {
...
+settings.get('darkMode', function(value) {
+  darkMode = value;
 })
 
-darkModeCheckbox.addEventListener('change', function (e) {
-  settings.set('darkMode', this.checked)
+  /* Setting up the RB's */
+    darkModeNever.checked = (darkMode === -1)
+    darkModeNight.checked = (darkMode === 0 || darkMode === undefined || darkMode === false) 
+    darkModeAlways.checked = (darkMode === 1 || darkMode === true)
```
PalmerAL, line 165: `settings.get` is asynchronous when called from the preferences page, so these should go inside the callback function.

labtorie:
> on macOS, most icons do have some kind of border

they do, but they also have their main parts (a compass needle, a music note, etc.). Min's icon have nothing but empty space, basically. Look how do icons look without their borders:
image image

They still have sense
And there is min's one:
image

Thus we see that the border is not just a border in our case. It's as important as the blue part. So I still think we should have them with the same value

PalmerAL:
It's definitely true that my graphic design skills are somewhat lacking :) After leaving them both side-by-side in the dock for a bit, I'm starting to like the larger-border version better; I'll put together an ICO and ICNS version to test.

labtorie:
What do you think about this ratio? : 
image

I can render it antialiased in PS and make .ico and .icns files with this one, if you want me to. Also we could update the linux icon and the installator gif. _(btw, if this gif can be animated, I could animate it as I know something about the motion design)_

Also, I've fixed those sync issues. Thanks for giving me a chance to contibute the project despite my JS knowledge is poor so far :)

PalmerAL:
I was just working on this actually. Here's my attempt at an ICO:
[icon256.ico.zip](https://github.com/minbrowser/min/files/4619351/icon256.ico.zip)

This has a ratio of 26% at 48px and larger, and 30% at <=40px.

Screenshots:
Screen Shot 2020-05-12 at 8 46 32 PM
Screen Shot 2020-05-12 at 8 50 31 PM

PalmerAL:
I'd definitely be open to animating the install gif - what kind of motion were you thinking about?

labtorie:
> This has a ratio of 26% at 48px and larger, and 30% at <=40px.

I do like this. Looks pretty and solid. Shall I paste them to my PR?

> what kind of motion were you thinking about?

uhh, maybe some kind of smooth accelerating/slowing down spinning (like the windows spinner does) for the inner circle. The outline will act like it's attached to the inner circle with springs (i'm bad at explanations): it'll follow the blue one with a delay, maybe  bounce a little. I imagine it and it looks modern

PalmerAL:
> I do like this. Looks pretty and solid. Shall I paste them to my PR?

Sounds good! I'll generate some files for mac and linux as well.

I'm having a hard time imagining the motion, but I'm interested in seeing it when you get a chance.

localization/languages/de.json
```diff
@@ -135,7 +135,10 @@
             "unsafeHTML": null //missing translation
         },
         "settingsAppearanceHeading": "Aussehen",
-        "settingsDarkModeToggle": "Dark-Mode einschalten",
+        "settingsEnableDarkMode": "Dunkelmodus einschalten:", 
+        "settingsDarkModeNever": "Nein", 
+        "settingsDarkModeNight": "In der Nacht", 
+        "settingsDarkModeAlways": null, //"Always" missing translation
```
PalmerAL, line 141: Would it be possible to translate this as well?

PalmerAL:
The dark mode changes look good; I've tested all the combinations of true/false/-1/0/1/day/night (I think), and everything seems to work.

localization/languages/de.json
labtorie: Oh, yes. Didn't notice

labtorie:
Great! So, the last step is to finish the icon, I suppose

PalmerAL:
All the icon files should be updated now.

labtorie:
splash
Here's my idea

PalmerAL:
The icons and the dark mode setting look good to me at this point, is there anything else you think should be changed?

I like the animation! There's just a couple of small things I might do differently. However, one interesting effect I noticed when trying this out in the installer is that on my (very slow) windows VM, having an animated image seems to make the installer significantly slower - installs are taking around 0:50 with a still image, and 2:30 with an animated image. I'm not sure if this still happens on a faster system or not, but we should probably make sure that isn't an issue before continuing on the animation.

labtorie:
> is there anything else you think should be changed?

I think it's allright. We did our best

> installs are taking around 0:50 with a still image, and 2:30 with an animated image

Weird thing. I can lower the gif's framerate, now it is 29,97 fps. But I'm afraid that it will look stuttering even with motion blurring. 

a newer version, btw:
animation

PalmerAL:
I'm going to merge this PR for now, but I'll try out the newer gif and see if that makes a difference.

PalmerAL:
With that gif, I get 2:05 (versus 0:45 when I ran the original installer again). Is there any way that you know of to have a variable framerate?

labtorie:
the newer one has the same framerate as the previous does. 

Indeed, my AMD A6 4455M laptop has:
1:11 for animated installer
0:52 for still one

>Is there any way that you know of to have a variable framerate?
I think I could render it with something around 15fps or/and compress the images. I have no idea why does it happen

labtorie:
This one is 15 fps:
animation
The installation took 1:08. Seems like lowering the framerate almost does not affect

PalmerAL:
If there isn't a workaround, it's probably better to just leave the still image; it's kind of strange that this is happening though.

labtorie:
That is kinda sad to leave this animation unused

labtorie:
What about this one? It has no loop and I guess it should work faster. 25 frames times 2 seconds.
windows-installer

labtorie:
windows-installer

15 fps, 20 frames total. Gonna test it as soon as the build finishes

PalmerAL:
I'm also seeing identical installation times with the last one compared to a still image, and I like the animation! Maybe a little less space between the text and the logo? The border color also seems a bit too dark; I think it should match the old one (although that may not be right either).

labtorie:
> Maybe a little less space between the text and the logo?

How about this? 
image

> The border color also seems a bit too dark; I think it should match the old one (although that may not be right either).

The border color is the same as the old one has:
image

Maybe you see it darker because of the white background. I could make it brighter but thus it won't match the icon.

By the way, if you want to, we could discuss it somewhere else like Telegram

labtorie:
image
This one is quite brighter

PalmerAL:
Sorry, I meant the blue border around the entire image. The old one is 1px rgb(152, 192, 236); the new one is rgb(56, 98,195), and it looks like it's wider as well. The original logo colors are fine!

The new spacing looks good, and I like the motion!

I'm fine with Github, but my email is listed in my profile if you'd rather use that.

labtorie:
81840412-a3219c80-9551-11ea-9ada-830a41bff155
Here's what I've got. Should I change anything else?

PalmerAL:
It looks great, thank you! Merged in 52a12fa61494624faac7eeac371a811c390f2774.

PR #1009, opened 2020-05-12
Chrome-like search suggestions behaviour

labtorie:
Hello again! 
I think the ability to complete search suggestions is really important in browsing experience. Sadly Min does not have this feature yet. Let's say I want to search "Ryan Gosling movies" in Min. I have to type in the whole string as DuckDuckGo suggests only these items with "Ryan":
image
Not really handy. 
I've decided to fix that. Here's how it works.
As I navigate through the items, search bar updates automatically:
image

When I return to the bar, it restores the initial request, which is 'Ryan'.

And when I hit the `space` key, the focus returns to the search bar and now I have ability to complete my request.
(фильмы = movies)
image
I think it makes the experiense better. 

TODO: I think hitting the arrow up key in the searchbar should make the last suggestion in focus.

### DIFF
--- a/js/searchbar/searchbar.js
+++ b/js/searchbar/searchbar.js
@@ -12,6 +12,20 @@ function openURLInBackground (url) { // used to open a url in the background, wi
   }
 }
 
+var inputUpdater = {
+  input: null,
+  updateInput: function () {    
+    var currentText = searchbar.el.querySelector('.searchbar-item:focus .title') 
+      searchbar.associatedInput.value = currentText.innerHTML
+  },
+  saveInitialInput: function() {
+     this.input =  searchbar.associatedInput.value
+  },
+  restoreInitialInput: function() {
+    searchbar.associatedInput.value = this.input
+  }
+}
+
 var searchbar = {
   el: document.getElementById('searchbar'),
   associatedInput: null,
@@ -48,7 +62,7 @@ var searchbar = {
 
     var allItems = [].slice.call(searchbar.el.querySelectorAll('.searchbar-item:not(.unfocusable)'))
     var currentItem = searchbar.el.querySelector('.searchbar-item:focus, .searchbar-item.fakefocus')
-
+   
     var index = allItems.indexOf(currentItem)
     var logicalNextItem = allItems[(previous) ? index - 1 : index + 1]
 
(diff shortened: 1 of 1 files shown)


PalmerAL:
I've run into this issue a bunch myself, so thanks for fixing it!

I think it's kind of distracting to have the input value update as you tab through the results. Usually, I tab through the results as I'm reading them to save time, and so I'm looking at the items in the list 1 or 2 items ahead of what's selected. With this change, every time I select a new item, my focus jumps back up to the input, which makes reading through the items slower.

I think it might be better avoid this by only updating the input value after you press space. The downside would be that it might be harder to figure out that there's a shortcut to do this.

Some other ideas:
* The right arrow key should probably do the same thing as space.
* For some of the items (wikipedia results, items for currently-open tabs), it doesn't make sense to put the title text in the URL bar. Instead of just using the item title, we could add an `inputValue` property to `searchbarUtils.createItem` that would set a `data-input-value` attribute on the list item, and then updateInput could check whether this property exists and use the value of it.

PalmerAL:
Patch for fixing the arrow up behavior:

```diff
diff --git a/js/navbar/tabEditor.js b/js/navbar/tabEditor.js
index 47d3e02..ce2a935 100644
--- a/js/navbar/tabEditor.js
+++ b/js/navbar/tabEditor.js
@@ -85,6 +85,8 @@ const tabEditor = {
       if (e.keyCode === 9 || e.keyCode === 40) { // if the tab or arrow down key was pressed
         searchbar.focusItem()
         e.preventDefault()
+      } else if (e.keyCode === 38) { // arrow up key
+        searchbar.focusItem({focusPrevious: true})
       }
     })
 
diff --git a/js/searchbar/searchbar.js b/js/searchbar/searchbar.js
index 6d87797..7a9f40d 100644
--- a/js/searchbar/searchbar.js
+++ b/js/searchbar/searchbar.js
@@ -64,7 +64,11 @@ var searchbar = {
       searchbar.associatedInput.focus()
       return
     } else if (allItems[0]) { // no item is focused.
-      allItems[0].focus()
+      if (previous) {
+        allItems[allItems.length - 1].focus()
+      } else {
+        allItems[0].focus()
+      }
     }
   },
   openURL: function (url, event) {
```

labtorie:
Indeed, on-screen updates usually get our attention, but I didn't even think that this one could be distractive. That's how every browser that I know behaves in order to make it's UI intuitive. But I wouldn't say I dislike the idea of keeping searchbar steady 'till we press `space`. Maybe we could also add buttons like these:
image

It would be useful for touchscreens and would hint that suggestions are modifiable. 

About WIKI Titles. I bet there are lots of cases like this:
image

So, maybe it makes more sense to make it work for wiki's too

PalmerAL:
I like the idea of having buttons on the right; we have some existing code for that we could use as well ([example](https://github.com/minbrowser/min/blob/master/js/searchbar/bookmarkManager.js#L66-L70)).

That's a good point regarding wiki articles, I'm fine with supporting those as well if you want to.

Milo123459:
I really like the idea of this:
I think that merging this (once conflicts are solved) would be a great idea since I'm moving from chrome to min.
It'd just be nice.

Milo123459:
Is this PR still active @PalmerAL?

PalmerAL:
I think not; @labtorie feel free to make a new PR if you want to work on this again.

PR #1022, opened 2020-05-21
Changes in Reading mode

idAlencar:
Reading mode
*Adding translation
*Other small adjustments

### DIFF
--- a/reader/index.html
+++ b/reader/index.html
@@ -2,6 +2,7 @@
 <html>
 
 <head>
+	<meta charset="UTF-8" />
 	<meta name="theme-color" id="meta-theme"/>
 	<script src="../ext/readability-master/Readability.js"></script>
 	<link rel="stylesheet" href="../css/base.css"/>
@@ -13,6 +14,7 @@
 
 	<button id="settings-button" class="fa fa-gear" data-label="buttonReaderSettings"></button>
 	<div id="settings-dropdown" hidden>
+		<label data-string="themeReader"></label>
 		<div class="setting-section">
 				<button class="theme-circle selected" data-theme="light" data-label="buttonReaderLightTheme"></button>
 				<button class="theme-circle" data-theme="sepia" data-label="buttonReaderSepiaTheme"></button>
@@ -23,14 +25,17 @@
 		</div>
 	</div>
 	<div id="site-nav-links"></div>
-	<div id="auto-redirect-banner">
+	<div id="auto-redirect-banner" hidden>
 		<label data-string="autoRedirectBannerReader"></label>	
 		<span id="auto-redirect-button-area">
 			<button id="auto-redirect-yes" data-string="buttonReaderRedirectYes"></button>
 			<button id="auto-redirect-no" data-string="buttonReaderRedirectNo"></button>
 		</span>
 	</div>
-	<div id="backtoarticle"><i class="fa fa-angle-left"></i><label class="fa" data-string="articleReaderView"></label></div>
+	<div id="containerbacktoarticle">
+	<div id="backtoarticle"><i class="fa fa-angle-left"><label id="lavelarticlereader" class="fa" data-string="articleReaderView"></label></i></div>
+	</div>
+	
 	<!-- DOM inserted in reader.js -->
 
 	<script src="../ext/Dexie.min.js"></script>
--- a/reader/index.html
+++ b/reader/index.html
@@ -32,8 +32,11 @@
 			<button id="auto-redirect-no" data-string="buttonReaderRedirectNo"></button>
 		</span>
 	</div>
-	<div id="containerbacktoarticle">
-	<div id="backtoarticle"><i class="fa fa-angle-left"><label id="lavelarticlereader" class="fa" data-string="articleReaderView"></label></i></div>
+	<div id="backtoarticle">
+		<a id="backtoarticle-link">
+			<i class="fa fa-angle-left"></i>
+				<span data-string="articleReaderView"></span>
+		</a>
 	</div>
 	
 	<!-- DOM inserted in reader.js -->
--- a/reader/index.html
+++ b/reader/index.html
@@ -14,7 +14,6 @@
 
 	<button id="settings-button" class="fa fa-gear" data-label="buttonReaderSettings"></button>
 	<div id="settings-dropdown" hidden>
-		<label data-string="themeReader"></label>
 		<div class="setting-section">
 				<button class="theme-circle selected" data-theme="light" data-label="buttonReaderLightTheme"></button>
 				<button class="theme-circle" data-theme="sepia" data-label="buttonReaderSepiaTheme"></button>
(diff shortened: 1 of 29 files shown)


reader/index.html
```diff
@@ -11,35 +12,41 @@
 
 <body>
 
-	<button id="settings-button" class="fa fa-gear" title="Reader Settings"></button>
+	<button id="settings-button" class="fa fa-gear" data-label="buttonReaderSettings"></button>
 	<div id="settings-dropdown" hidden>
+		<label data-string="themeReader"></label>
```
PalmerAL, line 17: I'm not sure this needs a label; I think having the colored circles is enough to figure out what they do.

PalmerAL:
It looks like  the CSS for the "original article" link may not be working correctly:

reader/readerContent.css
```diff
@@ -4,6 +4,7 @@ html {
 	margin: 0;
 	font-size: 19.5px;
 	color: rgb(15, 15, 15);
+	overflow: hidden;
```
PalmerAL, line 7: Why is this necessary?

reader/readerContent.css
```diff
@@ -98,13 +99,14 @@ ol li {
...
 }
 a {
-	color: #1b74bb;
+	color: #1e90ff;
 	font-size: 1em;
 	text-decoration: none;
 }
 body[theme="dark"] a {
-	color: #65a9ed;
+	color: #1e90ff;
```
PalmerAL, line 107: Can you switch back to the original colors? In a paragraph with a lot of links, the old colors made it easier to read the paragraph without looking at the links:



But with these colors, the links stand out more than the rest of the text, and it's hard to read:

reader/readerUI.css
```diff
@@ -2,6 +2,7 @@ body, html {
 	height: auto;
 	overflow: auto;
 	min-height: 100vh;
+	background-color: transparent;
```
PalmerAL, line 5: Why is this necessary?

idAlencar:
When zooming, I see a white background and two scroll bars

telafundo

barra

reader/index.html
idAlencar: the intention is more in the future to add the option to change the font color, I am already introducing the theme indicator

reader/readerContent.css
idAlencar: adjusted

idAlencar:
adjusted css - original article
t

PalmerAL:
I'm not sure what's causing the scrollbar issue; I've seen that occasionally before as well, so I don't think it's caused by this PR.

I think ideally the arrow icon next to "original article" should line up with the heading - the original is closer to that, although it's not perfect either.

idAlencar:
ti

reader/index.html
PalmerAL: What are some examples of situations where you would want to change the font color? I can't think of any combinations that work well except white-on-black or black-on-white (or almost-black on sepia), and we already have those.

PalmerAL:
Reducing the click area sounds good! I think it would be good to leave a little bit of space around the link that's clickable though.

The font on the link seems messed up. Before:
Screen Shot 2020-05-21 at 9 49 04 PM
After:
Screen Shot 2020-05-21 at 9 49 07 PM

I'll see if I can fix it.

PalmerAL:
I made some changes the link. Screenshots:

Click area:
click-area

With navigation links:
navlinks

With navigation links and banner:
navlinks_banner

With nothing:
nothing

PalmerAL:
Also, these screenshots actually show an issue with this change: https://github.com/minbrowser/min/pull/1022/files#diff-85209f196d84c2c5daa8572c49d942ecR5. If you set the page background to transparent, the `!screenshot` command produces a transparent screenshot, which isn't right. Unless there's a reason to make it transparent, I think this should be removed.

idAlencar:
In the original version, there is a white strip being generated on the side, you can see it by sliding the scroll bar to the right


telafundo

idAlencar:
I believe this solves the problem of the occurrence of this white band and we can remove the 

`background-color: transparent;`

reader/readerUI.css
idAlencar: I already made a correction, believe it solved it, so it is no longer necessary `background-color: transparent;`

PalmerAL:
That works for me as well!

reader/index.html
idAlencar: I think the choice of color is a lot of the person’s taste, particularly use links in the same color as the text, to avoid distractions, there may be other people who like something like that

PalmerAL: I'm not sure I agree; I think what we have is a pretty good balance between having the links not stand out too much, and still being able to tell they're there. In most articles that I've seen with links, the links are typically useful and an important part of the article, so I can't think of a lot of cases where I'd want the links to be hidden completely.

idAlencar: ok, it can be removed, i make a path for my use, do you know any automatic way to remove this item from translations? I put file by file, it's sad to think about removing one at a time kkkkkkk

PalmerAL: VS Code supports global find-and-replace with regexes, so I removed it with this: `\t"themeReader":.*\n`.

idAlencar:
hi, it prevents the appearance of a second scroll bar when zooming, I comment here

https://github.com/minbrowser/min/pull/1022#issuecomment-632359381

PalmerAL:
Are you still seeing the second scrollbar in the latest version? I saw that in the original version, but after the fix you made for the settings button, I don't see it anymore.

idAlencar:
I still see in the third zoom, I'm looking for what causes this bar

Peek 26-05-2020 18-03

idAlencar:
I see that the `Div class="reader-main` doesn't fit horizontally and vestually
when we have the zoom, the (div) has no maximum size and the files are very large, generating an extra scroll bar.
I think we can do something like this: 
1-we can keep `overflow: hidden;`; or

2 - we disable the scroll bar of the entire page and activate only in the `#document` area; or

3- we limit the size of the div

what do you think about?

Google translator

Captura de tela de 2020-05-28 07-46-30

Captura de tela de 2020-05-28 07-47-42

PalmerAL:
OK, I see the issue. When the article loads, the height of the `<iframe>` is set to the height of its contents, but if you zoom or resize the window, the content height changes, and then the frame itself scrolls, creating a second scrollbar.

Resizing the frame when the window size changes should fix this.

PalmerAL:
Can you try it with the commit I just added?

idAlencar:
ok, i will test tomorrow

idAlencar:
This works for me!, But I realized that the shortcut (ctrl + scroll) to zoom only works when the arrow is out of `#document`, did you notice that?

PalmerAL:
I see that too; I think it was an issue even before this PR. I created a new issue about it: #1038.

idAlencar:
good adjustments have been made, everything seems to work fine, I am satisfied

PalmerAL:
Great! Thanks for your work on this!

PR #1037, opened 2020-05-29
Follow the changed OS name "Raspberry Pi OS".

lee-yunseok:
With 8GB RAM released, Raspbian has been renamed to Raspberry Pi OS.

https://www.raspberrypi.org/downloads/raspbian/


PalmerAL:
Thanks.

PR #1061, opened 2020-06-12
Update en-US.json

JB1905:


PalmerAL:
Oops, thanks for fixing that.

PR #1062, opened 2020-06-12
fix menu-button closing tag

chimnz:


PalmerAL:
Thanks for pointing that out. With this change, the other buttons are placed inside the menu button, so I went ahead and made the correct fix in c43bed97cd77059bcb68c90093c3d37d265654bc.

Issue #1072, opened 2020-06-18
Please add a preference to switch to new tab immediately


BillDietrich:
When I control-click on a link to open it in a new tab, the tab is created but focus remains on the original tab.  Please add a preference to switch focus to the new tab immediately.  Thanks.

I'm using Min 1.14.1 on Ubuntu 20.04 GNOME.

BenStigsen:
I agree with this.

BenStigsen:
This option has now been merged into the master branch. PR https://github.com/minbrowser/min/pull/1075

PR #1075, opened 2020-06-18
Add setting for opening tabs in foreground 

BenStigsen:
From issue https://github.com/minbrowser/min/issues/1072

Allows the user to open a new tab and switch to it with CTRL+left.
I'm not quite sure what `disposition` refers to. And this should probably be a user preference instead of a direct change (which I'm not sure how to do).

### DIFF
--- a/pages/settings/index.html
+++ b/pages/settings/index.html
@@ -85,7 +85,7 @@
 			<label for="checkbox-separate-titlebar" data-string="settingsSeparateTitlebarToggle"></label>
 		</div>
 
-		<div class="setting-section" hidden id='section-open-tabs-in-foreground'>
+		<div class="setting-section" id='section-open-tabs-in-foreground'>
 			<input type="checkbox" id="checkbox-open-tabs-in-foreground" />
 			<label for="checkbox-open-tabs-in-foreground" data-string="settingsOpenTabsInForegroundToggle"></label>
 		</div>
--- a/pages/settings/index.html
+++ b/pages/settings/index.html
@@ -85,6 +85,11 @@
 			<label for="checkbox-separate-titlebar" data-string="settingsSeparateTitlebarToggle"></label>
 		</div>
 
+		<div class="setting-section" hidden id='section-open-tabs-in-foreground'>
+			<input type="checkbox" id="checkbox-open-tabs-in-foreground" />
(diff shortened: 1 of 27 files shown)


BenStigsen:
I'm still trying to get a hang of this file system, not to mention it's my first time messing around with Node.
I'm not sure why the setting isn't showing up in the settings / preferences page.

BenStigsen:
Hmm, is there some extra file I don't know about? I feel like I've looked in most of them now.

pages/settings/index.html
```diff
@@ -85,6 +85,11 @@ <h3 data-string="settingsAdditionalFeaturesHeading"></h3>
 			<label for="checkbox-separate-titlebar" data-string="settingsSeparateTitlebarToggle"></label>
 		</div>
 
+		<div class="setting-section" hidden id='section-open-tabs-in-foreground'>
```
PalmerAL, line 88: I think the `hidden` attribute is the problem :)

BenStigsen: Yep, that made the preference show up. Not sure why though, I literally just copied `section-separate-titlebar` which contains `hidden`

BenStigsen:
Am I right to assume that browserUI.js does not have access to `pages/settings/settings.js`?

Because this doesn't create a tab until the browser has been restarted:
```js
openInBackground: disposition === 'background-tab' && !settings.get('openTabsInForeground')
``` 

It's also not possible to do `var settings = require(../pages/settings/settings.js)` since that creates a view within the view.

BenStigsen:
Ah, requiring `js/util/settings/settings.js` fixed it.

BenStigsen:
Now when holding CTRL while clicking with the left mouse button, it opens a new tab in foreground.
This seems to work as intended and only minimal additions have been made to code. It should be ready for merging.

Min - Tab Open In Foreground

BenStigsen:
Oh wait, this should probably be merged into the 1.15.x branch instead right?

pages/settings/index.html
PalmerAL: The separate titlebar pref is only supposed to be visible on Linux, so it starts out hidden and there's some JS that checks the operating system and then un-hides it.

PalmerAL:
LGTM as well, thanks!

The master branch is the right one; the 1.15 branch only gets certain bug fixes (so that you can make a 1.15.1 release with just those fixes while continuing to merge other changes).

PR #1081, opened 2020-06-20
nl.json - Dutch translation

Firstl4rs:
Completed Dutch translation

### DIFF
--- a/localization/languages/nl.json
+++ b/localization/languages/nl.json
@@ -28,7 +28,6 @@
         "suggestedSite": "Voorgestelde website", //this is used to label suggested websites from the DuckDuckGo API,
         "resultsFromDDG": "Resultaten van DuckDuckGo", //this is used as a label to indicate which results come from DuckDuckGo's API
         "taskN": "Taak %n", //this is used as a way to identify which tab a task is in "task 1", "task 2", ...
-        "hostsFileEntry": "Uit hosts bestand", //this is a label used to indicate that a result is coming from the user's host file
         /* custom !bangs
         these are some of the items that show up when you press ! in the searchbar.
         Each one of these strings describes what the command will do when you run it. */
@@ -80,7 +79,7 @@
         "isFocusMode": "U bent in focusmodus.",
         "closeDialog": "OK", //used as a label for the button that closes the dialog
         "focusModeExplanation1": "In de focusmodus kan u geen nieuwe tabbladen openen of van taak wisselen.",
-        "focusModeExplanation2": "U kan de focusmodus weer uitschakelen in het menu 'beeld'. ,
+        "focusModeExplanation2": "U kan de focusmodus weer uitschakelen in het menu 'beeld'.",
         /* relative dates */
         "timeRangeJustNow": "Daarnet",
         "timeRangeMinutes": "Enkele minuten geleden",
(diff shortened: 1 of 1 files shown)


localization/languages/nl.json
```diff
@@ -0,0 +1,251 @@
+{
+    "name": "Dutch)",
```
PalmerAL, line 2:
```suggestion
    "name": "Dutch",
```

localization/languages/nl.json
```diff
@@ -0,0 +1,251 @@
...
+    "identifier": "nl-BE",
```
PalmerAL, line 3: How similar is Belgian Dutch to Dutch in the Netherlands? Should this set of translations be used in the Netherlands as well?

Firstl4rs:
No difference at all, just proud to be Belgian, hence the nl-be 😁. Feel
free to change to nl-nl.

Op zo 21 jun. 2020 21:09 schreef PalmerAL <notifications@github.com>:

PalmerAL:
Based on this [list](http://www.lingoes.net/en/translator/langcode.htm), the locale can be any of nl, nl-NL, or nl-BE, so I'm guessing we should use `nl` in the translation file, and then change the localization system so it will find a translation based on the first part of the code if there isn't an exact match.

localization/languages/nl.json
```diff
@@ -0,0 +1,251 @@
...
+        },
+        "settingsAppearanceHeading": "Uitzicht",
+        "settingsEnableDarkMode": "Donkere modus inschakelen:",
+        "settingsDarkModeNever": "Nooit",
+        "settingsDarkModeNight": "'s Nachts",
+        "settingsDarkModeAlways": "Altijd",
+        "settingsSiteThemeToggle": "Kleurthema van site inschakelen",
+        "settingsAdditionalFeaturesHeading": "Aanvullende opties",
+        "settingsUserscriptsToggle": "Gebruikerscripts inschakelen",
+        "settingsSeparateTitlebarToggle": "Toon de titelbalk",
```
PalmerAL, line 146: This is a pretty minor issue, and I'm reviewing this by using Google Translate and looking at a dictionary (so I could be totally wrong), but if I understand correctly this is "Show the titlebar" - it might make more sense to include "separate" here? There's always a titlebar shown regardless of this setting; the setting just controls whether the titlebar is combined in the same line as the other controls or is on a separate line ([screenshots](https://github.com/minbrowser/min/pull/977#discussion_r414918201)).

localization/languages/nl.json
```diff
@@ -0,0 +1,251 @@
...
+        "appMenuRedo": "Opnieuw doen",
+        "appMenuCut": "Knippen",
+        "appMenuCopy": "Kopiëren",
+        "appMenuPaste": "Plakken",
+        "appMenuSelectAll": "Alles selecteren",
+        "appMenuFind": "Zoeken",
+        "appMenuView": "Beeld",
+        "appMenuZoomIn": "Inzoomen",
+        "appMenuZoomOut": "Uitzoomen",
+        "appMenuActualSize": "100%",
```
PalmerAL, line 175: Is there a way to keep this in words so that it's consistent with the other menu items? Safari uses "werkelijke grootte" and Chrome uses "normaal"; would one of those work?

localization/languages/nl.json
Firstl4rs: Correct, I'll change that to 'Toon aparte titelbalk'

localization/languages/nl.json
Firstl4rs: In that case I'd go with 'Werkelijke grootte', although in my opinion 100% will do just fine.

localization/languages/nl.json
```diff
@@ -0,0 +1,251 @@
...
+        "taskDescriptionThree": "%t, %t, en %n meer", //used to describe a task that has three or more tabs
+        /* find in page toolbar */
+        "searchInPage": "Zoeken op deze pagina", //this is used as the placeholder text for the textbox in the find in page toolbar
+        "findMatchesSingular": "%i van %t overeenkomst", //this and the next label are used to indicate which match is currently highlighted
+        "findMatchesPlural": "%i van %t overeenkomsten",
+        /* Focus mode */
+        "isFocusMode": "U bent in focusmodus.",
+        "closeDialog": "OK", //used as a label for the button that closes the dialog
+        "focusModeExplanation1": "In de focusmodus kan u geen nieuwe tabbladen openen of van taak wisselen.",
+        "focusModeExplanation2": "U kan de focusmodus weer uitschakelen in het menu 'beeld'. ,
```
Sestowner, line 83: Missing closing double quote.

PR #1096, opened 2020-06-25
Update pt-PT.json

smarquespt:


PalmerAL:
Thanks for noticing that! There's another PR open (#1097) that fixes that and another error, so I merged that one instead.

smarquespt:
Ok. thanks

PR #1100, opened 2020-06-29
Fix postinst

lee-yunseok:
Just added newline at end of file.


PalmerAL:
Thanks. Just out of curiosity, what errors does this fix?

lee-yunseok:
I don't know what is the possible problem. However, I think that this can prevent the problem from happening because the last line ends with LF.

PR #1109, opened 2020-07-04
Add a light divider between tabs to increase contrast

eertmanhidde:
It was quite hard for me to see how many tabs I had open so I added a light divider between the tabs to improve that contrast.

**Before**
Screenshot 2020-07-04 at 13 32 06

**After**
Screenshot 2020-07-04 at 13 28 07

If this doesn't suit the idea of having a minimalistic browser I could probaly make this an optional feature in the users preferences.
Let me know what you think 😁

### DIFF
--- a/localization/languages/uk.json
+++ b/localization/languages/uk.json
@@ -150,7 +150,7 @@
         "settingsSiteThemeToggle": "Увімкнути тему сайту",
         "settingsAdditionalFeaturesHeading": "Додаткові можливості",
         "settingsUserscriptsToggle": "Увімкнути скрипти користувача",
-        "settingsShowDividerToggle": "Показати роздільник між вкладками",
+        "settingsShowDividerToggle": null, // Missing translation
         "settingsSeparateTitlebarToggle": "Використовувати окремий рядок заголовка",
         "settingsOpenTabsInForegroundToggle": "Відкриватии нові вкладки на передньому плані",
         "settingsUserscriptsExplanation": {
--- a/localization/languages/uk.json
+++ b/localization/languages/uk.json
@@ -18,7 +18,7 @@
         "copyLink": "Копіювати посилання",
         "copyEmailAddress": "Копіювати адресу електронної пошти",
         "selectAll": "Вибрати все",
-        "undo": "Скасувати",
+        "undo": "   ",
         "redo": "Повторити",
         "cut": "Вирізати",
         "copy": "Копіювати", //this is a verb (as in "copy the currently-selected text")
@@ -150,6 +150,7 @@
         "settingsSiteThemeToggle": "Увімкнути тему сайту",
         "settingsAdditionalFeaturesHeading": "Додаткові можливості",
         "settingsUserscriptsToggle": "Увімкнути скрипти користувача",
+        "settingsShowDividerToggle": "Показати роздільник між вкладками",
         "settingsSeparateTitlebarToggle": "Використовувати окремий рядок заголовка",
         "settingsOpenTabsInForegroundToggle": "Відкриватии нові вкладки на передньому плані",
         "settingsUserscriptsExplanation": {
--- a/localization/languages/uk.json
+++ b/localization/languages/uk.json
@@ -18,7 +18,7 @@
         "copyLink": "Копіювати посилання",
         "copyEmailAddress": "Копіювати адресу електронної пошти",
         "selectAll": "Вибрати все",
-        "undo": "   ",
+        "undo": "Скасувати",
         "redo": "Повторити",
         "cut": "Вирізати",
         "copy": "Копіювати", //this is a verb (as in "copy the currently-selected text")
--- a/css/tabBar.css
+++ b/css/tabBar.css
@@ -92,6 +92,9 @@ body.linux #menu-button {
 	top: var(--control-space-top) !important;
 	background: rgba(255, 255, 255, 0.8) !important;
 }
+.tab-item:not(:last-child) {
+	border-right: 1px solid rgba(255, 255, 255, 0.4);
+}
 .tab-item:not(.active):hover {
 	background-color: rgba(0, 0, 0, 0.03);
 }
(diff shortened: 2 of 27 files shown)


PalmerAL:
I do think the dividers are a bit distracting when you're not looking at the tabstrip, which I'd like to avoid. One possibility I mentioned in the last PR is only showing the dividers when hovering over the tabstrip, but I'm not sure if that actually solves the issue or not:

* If the issue is not knowing where the clickable area for each tab is, only showing the dividers on hover would probably be fine.
* If the issue is wanting to actually count the tabs and know where you are, it probably wouldn't be. This is actually probably related to what you were mentioning about the keyboard shortcuts to switch tabs - if you have a very specific order for your tabs and think of them based on position, it probably is difficult to tell them apart.
  * (And the traditional model in most browsers is that you look at the tab shapes, count them, and pick one to switch to - like "That's the third tab shape I see, so I'm going to press cmd+3 to switch to it". But what I find I do in Min, since the keyboard shortcuts to switch tabs are relative, is just sort of glance at the tabs and guess how far away the tab I want is, and then pick a number to switch to it that way. And in that case, the exact position of the tab doesn't really matter).

I also wonder if the issue depends on the number of tabs. In your screenshot above, it seems pretty easy to tell that there are 5 tabs even without the dividers, just based on the position of the text. But when there are more tabs, it becomes a lot harder to tell:
Screen Shot 2020-07-04 at 3 07 39 PM

I wonder if having a better way to truncate the tab titles would solve this issue by making it clearer where each title ends:
Screen Shot 2020-07-04 at 3 07 22 PM
Screen Shot 2020-07-04 at 3 06 04 PM

But I feel like I'm thinking like the person in [this movie](https://youtu.be/fXXNrHpaQ3k?t=7) except with browser tabs, so I guess if you feel strongly about it one way or another I'm fine with doing that.

eertmanhidde:
Yeah I do get that it's kinda a step away from the idea of Min being a truly minimalistic browser. The issue for me wasn't really not knowing where to click to select a tab. I like to be aware of how many tabs I have open so I check from time to time which tabs are open and how many I have open. It was just a little bit harder for me to 'read' how many tabs I have open.

At the end of the day I think it will come down to personal preference, so maybe having a checkbox in your preferences with something like:
- [ ] Show dividers between tabs

By default you could have it off and for the users who would like to have dividers could turn it on.
Or maybe even have 3 radio buttons with:

O No dividers between tabs
O Always show dividers between tabs
O Show dividers between tabs on hover

This would give the user more freedom to change it to their prefered way.
What is your opinion on this approach @PalmerAL ?

I think the truncating of text looks nice in your screenshots, it will let the user know that the title of a tab is cut off.

PalmerAL:
Yeah, that does make sense (although I'm a little curious about the "being aware of how many tabs I have open" comment - is the idea that you're trying to avoid having too many tabs, so you have a cut-off at which you start closing them?)

I made another version of this where I adjusted the colors a bit, and added a horizontal divider line to the dark theme:
```
.dark-theme #navbar:before {
	background: rgba(255, 255, 255, 0.2);
}
.tab-item:not(:first-child)::before {
	content: "";
	position: absolute;
	top: 0;
	left: 0;
	width: 1px;
	height: calc(100% - 1px);
	background-color: transparent;
	transition: 0.15s background-color;
}

#tabs-inner .tab-item:not(.active) + .tab-item:not(.active):before {
	background-color: rgba(0, 0, 0, 0.1);
}

.dark-theme #tabs-inner .tab-item:not(.active) + .tab-item:not(.active):before {
	background-color: rgba(255, 255, 255, 0.3);
}
```

I think it helps the divider lines fit in a bit better:
Screen Shot 2020-07-05 at 2 52 58 PM
Screen Shot 2020-07-05 at 2 52 47 PM

(The light theme might need a little more contrast, I'm not sure).

I do agree some kind of preference would be a good idea. I really have no idea what the majority opinion on this would be, but I think I'm fine with having a checkbox that's enabled by default; if we get too many complaints about it, we can switch it to off-by-default later.

Let me know if you have any questions about the settings code. The architecture document has a little bit of information about it; I think it's mostly possible to figure out by looking at the existing settings, but maybe it's not as clear as I think.

eertmanhidde:
@PalmerAL I've added it to the user preferences page to show dividers based on user preferences.

I added a grey color for both light and dark theme since it fits both quite well, there are edge cases when some sites have that same grey color as a theme-color but we can't avoid that probably. Let me know what you think 👍

PalmerAL:
The settings changes look good!

For the colors, it would probably work to have two different colors that are based on the text color:
```
#navbar.show-dividers .tab-item:not(:last-child) {
	border-right: 1px solid rgba(0, 0, 0, 0.15);
}
.dark-theme #navbar.show-dividers .tab-item:not(:last-child) {
	border-right-color: rgba(255, 255, 255, 0.4);
}
```

On a white or black background, these will look almost the same as the gray, but they'll also stand out against any other color (since the `dark-theme` class gets added and removed depending on whether it's needed to maintain contrast).

Example on a site with a gray background:

Screen Shot 2020-07-11 at 2 11 29 PM
Screen Shot 2020-07-11 at 2 12 41 PM

Also, where did you get the translations for the option from? If you know all those languages, or you have a reliable source for them, that's great, but I'd like to avoid automated translations unless we can confirm they're accurate.

localization/languages/uk.json
```diff
@@ -15,7 +15,7 @@
         "searchWith": "Шукати з %s", //%s will be replaced with the name of the current search engine
         "copyLink": "Копіювати посилання",
         "selectAll": "Вибрати все",
-        "undo": "Скасувати",
+        "undo": "   ",
```
PalmerAL, line 18: Was this intentional?

PalmerAL:
Everything looks good to me, is there anything else you wanted to change?

PalmerAL:
I'm going to merge this before there are more conflicts, since everything looks good to me. Thanks for working on this!

PR #1110, opened 2020-07-04
fix for buttons and navbar icons

idAlencar:
hi @PalmerAL, I made some corrections to the buttons and icons in the navigation bar, highlighted result, what do you see?

* size adjustment
* Icon declaration (index.html) to improve your ownership settings
* adding a Margin to decrease accidental clicks

Captura de tela de 2020-07-04 08 10 41

### DIFF
--- a/css/tabBar.css
+++ b/css/tabBar.css
@@ -35,13 +35,13 @@
 
 /* navbar buttons */
 
-#back-button {
+#icon-back-button {
 	opacity: 1;
 	transition: none;
 	font-size: 1.5em;
-	padding: 0 0.6em 0 0.4em;
+    	margin-top: 5px;
 }
-#back-button:disabled {
+#icon-back-button:disabled {
 	opacity: 0.2;
 }
 
@@ -50,11 +50,13 @@ body.is-edit-mode #back-button {
 }
 
 #menu-button {
-	margin-top: 1px;
+	display: none;
+	margin: 0;
 }
 
-#menu-button {
-	display: none;
+#icon-menu-button{
+	margin-top: 11px;
+    	font-size: 1.1em;
 }
 
 body.windows #menu-button,
@@ -192,15 +194,13 @@ body.dark-mode #overlay {
 /* buttons */
 
 .navbar-action-button {
-	padding: 0 1em;
 	height: 36px;
+	width: 36px;
 	line-height: 36px !important;
+	margin-left: 2px;
 }
 .navbar-action-button.invisible {
 	visibility: hidden;
-}
-.navbar-action-button + .navbar-action-button {
-	padding-left: 0.5em;
 }
 
 /* reader button */
--- a/css/tabBar.css
+++ b/css/tabBar.css
@@ -39,9 +39,17 @@
 	opacity: 1;
 	transition: none;
 	font-size: 1.5em;
-    	margin-top: 5px;
+	margin-top: 5px;
 }
-#icon-back-button:disabled {
+/* The mac window controls add additional space, so the button doesn't need as much space */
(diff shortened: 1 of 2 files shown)


PalmerAL:
Just to make it easier to see what's changed, I made a set of images to compare:

Screen Shot 2020-07-05 at 3 05 32 PM
Screen Shot 2020-07-05 at 3 06 07 PM

* The changes to the new tab and task buttons are definitely an improvement; it's nice that they're the same size now, and that the new tab button is aligned correctly.
* I think the back button is too small now; can you change it back to the previous size?
  * And the `margin-top: 1px` on the menu button was added to align it with the back button, so that should probably be kept.
* The margins are a good idea!
* I don't understand why the icons need to be a separate element - I think adjusting the font-size and padding of the button element is enough. But if you want to leave them this way that's fine as well.

idAlencar:
1 - I chose to use the icons separated by settings that were not legal when using the padding (the menu and back icons are not aligned)
2- It looks like the Back button is getting different for macO, as you can see in my screenshot, it has the same size as before 1.5 em”.
Google translator

idAlencar:
What do you think?
Peek 06-07-2020 13-45

PalmerAL:
Sorry, I forgot to use the updated HTML file when testing this before, which is why the back button wasn't rendering correctly.

I made some changes to the back button CSS to make it fit better on mac; other than that, I think everything looks good if you're OK with it!

PalmerAL:
I'll look at the hover effects tomorrow.

Sestowner:
Color in the address bar is too dim.
image

idAlencar:
in Patch-4, it was changed,
* adding effect
* changes the size of the buttons (24px)
* Removal of the Task button (invisible)
* removes the relative position property from the navigation bar (for the task button to be corrected)
* task button is now declared in tabbar.css.

for me it works well, how can it work for MacOs and Windows?

idAlencar:
hi @PalmerAL,  it seems to me that [patch-4](https://github.com/idAlencar/min-dev-test/tree/patch-4) is more suitable to be a candidate for PR

hi @Sestowner, I just made some corrections in [patch-4](https://github.com/idAlencar/min-dev-test/tree/patch-4) you can check if the problem continues?

Sestowner:
> hi @Sestowner, I just made some corrections in [patch-4](https://github.com/idAlencar/min-dev-test/tree/patch-4) you can check if the problem continues?

The problem is solved.

idAlencar:
Ok, good, thank! @Sestowner 

I organized a possible PR in [patch-5 ](https://github.com/idAlencar/min-dev-test/tree/patch-5), has the same modifications, but is organized

PalmerAL:
@idAlencar If you want to use that branch instead, can you open a PR for it?

From looking at it:

* The hover effects themselves are nice!
* I think the clickable area of the buttons should stay the same size as they are now; the best way to fix that is probably to leave the buttons unchanged, and use a css `:before` element for the background.
* The menu button needs to be hidden on macOS (as it was before).
* There isn't enough space for the task overlay button now:
Screen Shot 2020-07-07 at 4 42 10 PM
The reason the invisible button is there is so the available space adjusts automatically, so it's probably easiest just to leave it.

idAlencar:
ok i will create a PR for Patch-5

PalmerAL:
@idAlencar Does the other PR replace this one? Or do I need to merge this one, and then the other one as well?

idAlencar:
@PalmerAL, each PR is independent, I will probably discard both and do another branch with all the suggested changes, I think it is more organized

idAlencar:
After #1134 this PR needs to be modified, I will close and update

PR #1116, opened 2020-07-08
 Second proposal to correct buttons on the NavBar starting from (#1110)

idAlencar:
@PalmerAL , I have already redid the invisible button statement, as soon as you have some time, can you check? the size I'm still not sure how I will do it

### DIFF
--- a/css/tabBar.css
+++ b/css/tabBar.css
@@ -40,7 +40,7 @@
 	opacity: 1;
 	transition: none;
 	font-size: 1.5em;
-	padding: 0 0.1em 0.2em 0;
+	padding: 0 0.1em 0.1em 0;
 	line-height: 18px;  /* corrects unexpected vertical movement*/
 }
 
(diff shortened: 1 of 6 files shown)


PalmerAL:
The button is still off center:



I think the issue is that the navbar buttons aren't big enough. The old invisible button is this size:


But the new one is smaller:

And doesn't have much space on the right.

I think the change I mentioned in the other PR (keeping the buttons the same and adding the background with a separate `:after` element) should fix this as well.

Example (based on the current master branch, it would be a bit different with the changes from your other PR):
```
.navbar-action-button {
	position: relative;
}

.navbar-action-button:hover:after {
	content: "";
	width: 2em;
	height: 2em;
	background-color: red;
	position: absolute;
	left: 4px;
	z-index: -1;
	border-radius: 50%;
}
```

idAlencar:
I understand, so it will be easier to continue with the idea of ​​the previous PR. I will organize the adjustments

idAlencar:
After #1134 this PR needs to be modified, I will close and update

PR #1127, opened 2020-07-13
New Iconset!

antfu:
This PR replaced the Font Awesome icons with the [Carbon](https://github.com/carbon-design-system/carbon/tree/master/packages/icons) (you can browse them in [this page](https://iconify.antfu.me/collection/carbon)).

---

Font Awesome it great but it seems to be too bold for the minimalist style of Min. Carbon offered some more thin-lined and modern style icons. In this PR, I also replaced the Font based icons to the unified js-based library [Iconify](https://iconify.design/). It's lighter, faster and can be replaced easily with other icon sets.

<details>
<summary>Some screenshots:</summary>

Screen Shot 2020-07-14 at 12 47 26 AM

Screen Shot 2020-07-14 at 12 47 20 AM

Screen Shot 2020-07-14 at 12 49 38 AM

Screen Shot 2020-07-14 at 12 49 05 AM

Screen Shot 2020-07-14 at 12 51 12 AM

Screen Shot 2020-07-14 at 12 53 30 AM

</details>


PalmerAL:
The icons look nice! And at least at first glance, they do seem to be an improvement over what we have.

I've only tested this for a few minutes now, and I haven't read through all the changes yet, but one thing that initially stands out to me is the approach that iconify is using to create the icons. It looks like the library is using a mutation observer to watch for newly-created DOM elements, and then inserting the SVG's. The issue with this is that this means the icons aren't visible when they're initially created, and then appear a little bit later, which causes flickering. Here's a video of this happening with the commands dropdown: [icon-bug.mp4.zip](https://github.com/minbrowser/min/files/4914720/icon-bug.mp4.zip).

The place where this is really visible is the bookmarks manager, where I have about ~100 bookmarks that appear in the manager list. When the list is first opened, there are no icons, and then the iconify library runs some JS to update them all (which alone takes ~120ms), and then the items all shift around as the icons appear. I haven't measured it, but I would guess that the mutation observer adds some overhead to other DOM updates as well.

I'm not sure if there's a solution to this or not, but I'm guessing that going back to an icon font would avoid these issues; how difficult would it be to create a font based on these icons?

antfu:
Oh, I didn't notice that. Just make the creation of the icons in sync instead of waiting for the observer to update. The flickering should be gone (at least for me). Please help to check it out :)

Switching to the font is possible, but it will lose the ability to add/change icons. Would happy to do it if the Iconify solution is not ideal for you.

PalmerAL:
That's definitely an improvement, and the flickering is gone for me now. However, there still seems to be a performance regression. With this branch, the time to open the bookmarks list (which contains ~100 star icons) goes from ~100ms to ~250ms:

Before:
icons before
After:
icons after

It looks like the additional time is spent roughtly equally in `getSVGObject`, and parsing the HTML for each icon.

It seems like the icon font would solve this - as far as I can tell, the existing icon font has basically no performance impact. In order to keep the ability to add icons, we could just include the entire Carbon set in the font - that would increase the file size a bit, but I think that's probably fine. But I'd be open to a different solution as well, if you have any other ideas.

antfu:
Thanks for the pref tests. You are right, we should go for using font instead, I will make the changes soon.

antfu:
Emm, Github seems to be broken, I have pushed the commits but the PR doesn't get updated. I will close this to create another one.

PR #1134, opened 2020-07-15
New Iconset! Re

antfu:
This continues #1127.

TL;DR: I have changed the icons back to font-based. It works for most of the icons, but for some icons, I didn't know how to make them show up. It would be great if you can help to check them out.

---

Actually Carbon doesn't offer the font version, and I am also planning to have the icons batch export functionality for my icon explore. I took something to investigate the conversion, it ends up with me making [a library](https://github.com/antfu/svg-packer) for that 😂.  (I am really enjoying making it, thanks for the inspiration!)

And so I selected some of the icons that I think would be possibly useful for Min and generated the iconfont. You can have a look at the link below.

[iconfont-min.zip](https://github.com/minbrowser/min/files/4926507/iconfont-min.zip)

In the PR, I remove all the fonts except for `woff2` as we don't need backward compatibility for electron. The font size is about 5KB, compares to FontAwesome's 66KB (or 819KB with all font formats), I think it's a huge improvement :)

---

BTW, the only thing left is the icons in `passwordFill.js` which I am not very sure what is happened, I will have a deeper look at it later.


PalmerAL:
That's a lot better! (And nice library!) I'll start going through this and giving it a more detailed review.

If it's only 50-60kb to bundle all the icons into the font, I think it's probably worth it; otherwise, we'd potentially have to regenerate the iconset every time we added a new feature in the future, which would be nice to avoid.

css/findinpage.css
```diff
@@ -37,12 +37,6 @@ body.linux #findinpage-bar {
...
-	padding: 0 0.5em;
-	margin-top: -2px;
-}
-#findinpage-bar .icon.fa-angle-up,
-#findinpage-bar .icon.fa-angle-down {
-	font-size: 1.2em;
-}
+#findinpage-bar button {
+	padding: 0 0.25em;
+}
```
PalmerAL, line 42: It would probably be nice if these were aligned better:

Screen Shot 2020-07-16 at 10 59 57 AM

However, the old icons weren't aligned very well either, so it's fine if you want to leave it as-is.

css/listItem.css
```diff
@@ -43,20 +43,18 @@
 }
 
 .searchbar-item i,
+.searchbar-item > .i:first-child,
```
PalmerAL, line 46: This can be removed now that we're using the icon font.

css/listItem.css
```diff
@@ -107,11 +105,9 @@ body.dark-mode.touch .searchbar-item .action-button {
 }
 .searchbar-item .action-button:hover {
     opacity: 1;
-    background: rgba(0, 0, 0, 0.05);
 }
 .dark-mode .searchbar-item .action-button:hover {
     opacity: 1;
-    background: rgba(0, 0, 0, 0.15);
```
PalmerAL, line 114: Can you leave this as-is?

css/searchbar.css
```diff
@@ -11,6 +11,7 @@
 	-webkit-user-select: none;
 	max-height: calc(100% - 125px);
 	overflow-y: auto;
+	padding-bottom: 0.5em;
```
PalmerAL, line 14: This should probably only be applied when the searchbar contains > 0 results, otherwise it looks strange:

Screen Shot 2020-07-16 at 11 09 39 AM

I can't think of an easy way to do this at the moment.

css/tabBar.css
```diff
@@ -38,7 +38,6 @@
 #back-button {
 	opacity: 1;
 	transition: none;
-	font-size: 1.5em;
 	padding: 0 0.6em 0 0.4em;
```
PalmerAL, line 41: The back button isn't aligned correctly anymore:

Screen Shot 2020-07-16 at 11 11 30 AM

css/tabBar.css
```diff
@@ -102,12 +101,12 @@ body.linux #menu-button {
...
+	margin-right: -0.7em;
 }
 .tab-item .tab-icon {
-	font-size: 0.8em;
-	margin-top: -0.1em;
-	padding: 0.66em;
-	vertical-align: middle;
+	margin-top: -2px;
+	font-size: 1em;
+	padding: 0.66em 0.5em;
```
PalmerAL, line 109: It seems like the sizes a bit off; here's a comparison:

Screen Shot 2020-07-16 at 11 18 17 AM

It feels like the private tab and insecure icons should be a bit smaller than they are, and the close button should be a bit larger.

css/tabBar.css
```diff
@@ -125,6 +124,9 @@ body.linux #menu-button {
 .tab-item.fade:hover {
 	opacity: 1;
 }
+.tab-item .title {
+	text-overflow: ellipsis;
+}
```
PalmerAL, line 129: Can you remove this for now? it doesn't actually work without switching the tab contents to a flexbox-based layout, and I think I'd rather use a fade-out effect like in https://github.com/minbrowser/min/pull/1109#issuecomment-653810851.

css/tabBar.css
```diff
@@ -192,32 +194,28 @@ body.dark-mode #overlay {
 /* buttons */
 
 .navbar-action-button {
-	padding: 0 1em;
+	padding: 0 0.5em;
 	height: 36px;
-	line-height: 36px !important;
+	margin-right: 0.5em;
+	margin-top: -1px;
```
PalmerAL, line 200: The new tab and task buttons seem a bit small to me; would it be possible to make them larger?

css/tabBar.css
```diff
@@ -192,32 +194,28 @@ body.dark-mode #overlay {
...
 	padding-left: 0.5em;
 }
+.navbar-right-actions {
+	margin-left: 0.5em;
+	margin-right: 2em;
+}
 
 /* reader button */
 
 .tab-item .reader-button {
```
PalmerAL, line 215: The reader button seems to be misaligned now:

Screen Shot 2020-07-16 at 11 22 35 AM

css/tabBar.css
```diff
@@ -230,14 +228,14 @@ body.dark-mode #overlay {
...
+	margin-right: 0.66em;
+	margin-top: -0.2em;
+	margin-left: -0.3em;
+	border-radius: 0.75em;
 }
-.tab-item .tab-icon.permission-request-icon i {
-	margin: 0 0.15em;
-	vertical-align: bottom;
+.tab-item .tab-icon.permission-request-icon .i {
+	vertical-align: middle;
```
PalmerAL, line 238: This button is probably a little too large now, or it needs to be moved upwards:

Screen Shot 2020-07-16 at 11 25 55 AM

(It looks fine in your original screenshot, I guess it regressed at some point)

css/tabBar.css
```diff
@@ -288,7 +286,6 @@ body.is-focus-mode .navbar-action-button:not(#menu-button) {
 	position: absolute;
 	top: 0;
 	left: 0;
-	margin: 0 0.8em;
```
PalmerAL, line 291: Can you revert this? I think it looks better with the bar aligned to the tab title (as it was before).

css/tabEditor.css
```diff
@@ -12,9 +12,7 @@
 }
 
 #tab-editor-input {
-    background: none;
-	line-height: 1.15em;
-	font-size: 1.15em;
```
PalmerAL, line 17: Maybe `1.1em`?

css/tabEditor.css
```diff
@@ -31,7 +29,8 @@
 }
 
 #tab-editor .tab-editor-button {
-	padding: 0.5em;
+	padding: 0.5em 0.3em;
+	font-size: 0.9em;
```
PalmerAL, line 33: I think they look better at `1em`.

css/taskOverlay.css
```diff
@@ -1,9 +1,13 @@
...
+	display: flex;
 	width: 100%;
 	height: calc(36px + var(--control-space-top));
 	background-color: inherit;
 	z-index: 1;
-	border-bottom: 1px rgba(0, 0, 0, 0.1) solid;
+}
+
+.simulated-navbar .spacer {
+	flex: auto;
```
PalmerAL, line 10: This doesn't look like it's actually used.

css/taskOverlay.css
```diff
@@ -1,9 +1,13 @@
 .simulated-navbar {
+	display: flex;
 	width: 100%;
 	height: calc(36px + var(--control-space-top));
 	background-color: inherit;
 	z-index: 1;
-	border-bottom: 1px rgba(0, 0, 0, 0.1) solid;
```
PalmerAL, line 6: Can you keep the border so it's consistent with the tab bar?

index.html
```diff
@@ -86,23 +86,27 @@
...
-		<i class="fa fa-bars navbar-action-button invisible"></i>
+		<div class="navbar-right-actions">
+			<button id="add-tab-button" class="navbar-action-button i carbon:add" data-label="newTabAction" tabindex="-1">
+			</button>
+		</div>
 	</div>
 
-	<button id="switch-task-button" class="fa fa-bars navbar-action-button theme-text-color windowDragHandle"
-		data-label="viewTasks" tabindex="-1"></button>
+	<button id="switch-task-button" class="navbar-action-button theme-text-color windowDragHandle i carbon:chevron-down"
```
PalmerAL, line 107: I feel like a horizontal-menu button would be better, but the four-line one that Carbon has doesn't fit very well. I think I might try drawing one (like how the old reader button worked).

js/searchbar/bangsPlugin.js
```diff
@@ -170,7 +170,7 @@ function getBangSearchResults (text, input, event) {
       showBangSearchResults(text, results, input, event, 4)
       searchbarPlugins.addResult('bangs', {
         title: l('showMoreBangs'),
-        icon: 'fa-angle-down',
+        icon: 'carbon:chervon-down',
```
PalmerAL, line 173:
```suggestion
        icon: 'carbon:chevron-down',
```

js/searchbar/searchbarUtils.js
```diff
@@ -55,10 +56,15 @@ function createItem (data) {
   }
 
   if (data.icon) {
-    var i = document.createElement('i')
-    i.className = 'fa' + ' ' + data.icon
+    if (data.icon.startsWith('fa')) {
```
PalmerAL, line 59: This can probably be removed.

js/util/createIcon.js
```diff
@@ -0,0 +1,5 @@
+module.exports = function (icon) {
```
PalmerAL, line 1: This function is small enough now that it might be easier to remove it and just create the icons directly.

reader/index.html
```diff
@@ -7,12 +7,12 @@
...
 	<link rel="stylesheet" href="../css/base.css"/>
 	<link rel="stylesheet" href="readerUI.css" />
-	<link rel="stylesheet" href="../ext/font-awesome-4.4.0/css/font-awesome.min.css" />
+	<link rel="stylesheet" href="../ext/icons/iconfont.css" />
 </head>
 
 <body>
-
-	<button id="settings-button" class="fa fa-gear" data-label="buttonReaderSettings"></button>
+	<button id="settings-button i carbon:settings-adjust" data-label="buttonReaderSettings">
```
PalmerAL, line 14:
```suggestion
	<button id="settings-button" class="i carbon:settings-adjust" data-label="buttonReaderSettings">
```

PalmerAL:
passwordFill uses a separate SVG to avoid having to load the entire icon font on every page; you should be able to just update `keyIcon` with one of the new SVG's.

css/tabBar.css
antfu: Yeah, sorry it was just a expirement that I forget to remove.

antfu:
Thank you so much for your detailed review! However, I didn't get my laptop with me, I will do that tmr

css/tabBar.css
antfu: It's because that `add` and `close` are designed smaller than others. Guess I might need to enlarge them in order to make them look the visually same size.
image

css/searchbar.css
antfu: Emm, maybe `.item:last-child { margin-bottom: 0.5em; }` ?

PalmerAL: The items are nested inside a set of containers depending on their type:

```
<div id="searchbar">
    <div class="searchbar-plugin-container">
        <div class="searchbar-item">
    <div class="searchbar-plugin-container">
        <div class="searchbar-item">
```

So it would need to be something like `#searchbar .searchbar-plugin-container:not(:empty):last-of-type .item:last-child`, but I think :last-of-type only matches based on the tag name.

css/listItem.css
PalmerAL: Actually, some of the icons are `<i>` elements, and others are `<span class="i">` still, so removing this broke the commands dropdown:

Screen Shot 2020-07-16 at 10 18 22 PM

It'd probably be best to pick one of the two methods (I don't think there's any real difference) and use it for all the icons.

antfu:
> If it's only 50-60kb to bundle all the icons into the font, I think it's probably worth it; otherwise, we'd potentially have to regenerate the iconset every time we added a new feature in the future, which would be nice to avoid.

Here is the full set. 66kb for css, and 120kb for the woff2. If you are ok with it, i can ship it with the PR.

[carbon.zip](https://github.com/minbrowser/min/files/4939005/carbon.zip)

PalmerAL:
Looks fine to me!

Minifying the CSS would probably cut the size a little bit, but I'm fine leaving it as-is if you don't want to do that.

css/tabBar.css
antfu: This is a bit weird though, it looks like the mac control buttons are misaligned instead of the back button. Or is everything in the menu bar should be a bit offset to the bottom?

PalmerAL: Hmm, that's a good point. We could change the position of the window controls, and make everything actually centered:

Screen Shot 2020-07-21 at 2 48 15 PM

It looks kind of odd to me, but I think that's just because I've spent so much time looking at the old UI, so I'd be fine with doing that if you want.

(To do that, I set `titleBarstyle: 'hidden', trafficLightPosition: {x: 12, y: 19}` on the window, removed the line-height from .tab-item, and repositioned the items using margins).

idAlencar:
hi @antfu, i'm following your work, do you have a branch that i can test?

idAlencar:
hi @antfu, i'm following your work, do you have a branch that i can test?

PalmerAL:
@idAlencar You should be able to do `git clone https://github.com/antfu/min.git`, `git checkout ui/newIcons`.

antfu:
Sorry for a bit delay, I will work on it now. @idAlencar thanks for getting interested in it, let me know what you have covered so we won't do duplicated works. If you need to push commits to this PR, I can give you the access to it as well :)

css/tabBar.css
antfu: Done :)

css/searchbar.css
antfu: Removed

css/listItem.css
antfu: Made `<span class="i">` -> `<i class="i">`, hope I have covered them all XD

css/tabBar.css
antfu: image

Changed to this, is that looks good to you?

antfu: BTW is there any website that I can use to test mic and camera at the same time?

index.html
antfu: Yeah, that's why I switched to chevrons. Guess I should leave it as-is for now?

Another options might be `carbon:overflow-menu-vertical`
image

or `carbon:overflow-menu-horizontal`
image

What do you think?

antfu:
I think I have covered the most of your suggestions, besides the password fill. I didn't manage to make it shown, is there any tips for that? And after that, I guess we are safe to remove font awesome?

css/tabBar.css
PalmerAL: Yeah, it looks fine.

Google meet will make a single request for camera and mic, which results in this:
Screen Shot 2020-07-25 at 5 00 49 PM

I think it's possible to have separate requests for camera and microphone (which would create two separate bubbles), but I haven't found a way to do it yet.

You can also do this:

```
window.addEventListener("click", function() {
navigator.mediaDevices.getUserMedia({video: true, audio: true})

Notification.requestPermission();

});
```

Screen Shot 2020-07-25 at 5 06 21 PM

PalmerAL:
In order to see the password fill icon you'll need to connect a password manager and go to a site you have a saved password for - maybe that's the issue?

I replaced it with "carbon:password" - old:
Screen Shot 2020-07-25 at 5 16 08 PM
vs new:
Screen Shot 2020-07-25 at 5 19 47 PM

index.html
PalmerAL: `overflow-menu-vertical` is already used on the left side for the app menu on Windows and Linux; having two different buttons with the same icon would probably be too confusing. I also think it helps a bit to have the icon look like a "list", since it shows a vertical list when pressed.

I'll try making a three-bar menu SVG when I have some time tomorrow or Monday.

PalmerAL:
It's looking pretty good! I don't see any usages of FA left, so I think it's fine to remove (and if something breaks, we can always fix it).

I think the navbar button spacing need to be adjusted a bit more - it seems like there's too much space to the right of the back button, and the task button needs to be moved to the right a bit, but I guess we should figure out what to do with the task overlay button before adjusting the spacing; I'll try to work on it soon.

PalmerAL:
Here's a simple three-bar icon for the task overlay:
```
		<svg style="width: 1em; height: 1em; filter:invert(1)" version="1.1" viewBox="0.0 0.0 32.0 32.0" fill="none" stroke="none" stroke-linecap="square"
			stroke-miterlimit="10" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns="http://www.w3.org/2000/svg">
			<g clip-path="url(#p.0)">
				<path fill="#000000" d="m0 0l32 0l0 5l-32 0z" fill-rule="evenodd" />
				<path fill="#000000" d="m0 13l32 0l0 5l-32 0z" fill-rule="evenodd" />
				<path fill="#000000" d="m0 26l32 0l0 5l-32 0z" fill-rule="evenodd" />
			</g>
		</svg>
```

I haven't been able to get the spacing right very easily, but with a bunch of CSS hacks applied, it looks something like this:

Screen Shot 2020-07-29 at 12 53 37 AM

antfu:
Wow looks great! Good work

PalmerAL:
Everything looks good to me, do you see anything else to do?

antfu:
Looks good to me too. Thanks for your detailed review :)

PalmerAL:
Thank you for working on this!

PR #1141, opened 2020-07-18
Add media autoplay setting

Sestowner:


PalmerAL:
What do you think about just making `user-gesture-required` the default (and not having a preference)?

The history of this is sort of confusing; autoplay was allowed up until at least v1.9, and then in 1.9.1, it was disabled (in 33a364a71e4511de99acbd9fb233dc7dcc57da58). But it looks like Electron changed their default at a later point (in https://github.com/electron/electron/pull/16396), which made it be enabled again.

It generally makes sense to me to disable it, and I don't recall any complaints from the time period when it was disabled, so I'd be fine doing that if it makes sense to you.

Sestowner:
> What do you think about just making `user-gesture-required` the default (and not having a preference)?

> It generally makes sense to me to disable it, and I don't recall any complaints from the time period when it was disabled, so I'd be fine doing that if it makes sense to you.

I don't mind because I prefer media autoplay to be disabled. Go for it.

PalmerAL:
Done in 83357fdfe93cb25aa50c458179ba8a185fb1e962.

PalmerAL:
It turns out we actually do need a preference for this because of #1413; I'm going to make a new PR based on this one.

PalmerAL:
New PR: #1550

PR #1144, opened 2020-07-19
Add Copy Email Address to context menu

### LINKED ISSUE #1136: feature request: Min does not recognize mailto links - missing "Copy email" choice under right click menu
The next most disturbing missing feature is that Min does not recognize mailto link when to right click on email address. There is only a choice "Copy link" (this will copy email with mailto: which is not needed) but missing "Copy email" choice under right click menu which will copy only email address.

Sestowner:
fixes #1136

### DIFF
--- a/js/webviewMenu.js
+++ b/js/webviewMenu.js
@@ -194,12 +194,24 @@ const webviewMenu = {
     }
 
     if (link || (mediaURL && !mediaURL.startsWith('blob:'))) {
-      clipboardActions.push({
-        label: l('copyLink'),
-        click: function () {
-          clipboard.writeText(link || mediaURL)
+      if (link.startsWith('mailto:')) {
+        var ematch = link.match(/(?<=mailto:)[^\?]+/)
+        if (ematch) {
+          clipboardActions.push({
+            label: l('copyEmailAddress'),
+            click: function () {
+                clipboard.writeText(ematch[0])
+            }
+          })
         }
-      })
+      } else {
+        clipboardActions.push({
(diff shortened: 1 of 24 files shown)


js/webviewMenu.js
```diff
@@ -194,6 +194,18 @@ const webviewMenu = {
...
+          click: function () {
+            var ematch = link.match(/(?<=mailto:)[^\?]+/)
+            if (ematch) {
+              clipboard.writeText(ematch[0])
+            }
+          }
+        })
+      }
+
       clipboardActions.push({
```
PalmerAL, line 209: Maybe put this inside an `else` block, so that "copy link" doesn't show up for email addresses?

js/webviewMenu.js
```diff
@@ -194,6 +194,18 @@ const webviewMenu = {
     }
 
     if (link || (mediaURL && !mediaURL.startsWith('blob:'))) {
+      if (link.startsWith('mailto:')) {
+        clipboardActions.push({
+          label: l('copyEmailAddress'),
+          click: function () {
+            var ematch = link.match(/(?<=mailto:)[^\?]+/)
+            if (ematch) {
```
PalmerAL, line 202: Minor issue: `ematch` should probably be found before the item is created, and then if there isn't a match, the menu item shouldn't be shown at all.

PalmerAL:
LGTM aside from comments.

PalmerAL:
Sorry, Github didn't send me a notification the first time.

Everything looks good!

PR #1160, opened 2020-07-23
Use openTabsInForeground setting for addTab option openInBackground

Sestowner:
@PalmerAL Looks like you conceived something in #1153.

What do you think about this? I made these changes a couple of days ago, and this works well for me.

### DIFF
--- a/js/browserUI.js
+++ b/js/browserUI.js
@@ -22,9 +22,12 @@ function addTask () {
 /*
 options
   options.enterEditMode - whether to enter editing mode when the tab is created. Defaults to true.
-  options.openInBackground - whether to open the tab without switching to it. Defaults to false.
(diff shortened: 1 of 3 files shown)


PalmerAL:
It seems fairly confusing to me that the "openInBackground" option now represents two different things - whether the caller of `addTab` wants the tab to open in the background, and what the value of the setting is. With this change, if you're looking at a place where addTab() is called, it's no longer clear whether the tab will open in the background or not, but in many cases, the tab needs to always open in the foreground or background, regardless of the preference, so that could become a problem (particularly if you're testing your code with one value of the preference, and forget to test the other).

As far as I know, the only two places where "open new tabs in the foreground" should apply are the context menu and when right-clicking on links. So I think it'd be better to just check the preference in those cases, and call addTab with the correct value of `openInBackground` as needed.

PR #1168, opened 2020-07-26
package: support ARM64 platform

jyangchisyan:
The package compatable for arm64 target platform

Signed-off-by: Samuel Jiang <chyishian.jiang@gmail.com>


PalmerAL:
Rather than duplicating `buildDebian.js`, I think it would be better to have it take a `platform` argument that it passes to createPackage (since the code is otherwise identical). So the three commands would look something like this:
```
    "buildDebian": "npm run build && node ./scripts/buildDebian.js --platform=amd64",
    "buildRaspi": "npm run build && node ./scripts/buildDebian.js --platform=raspi",
    "buildARM64": "npm run build && node ./scripts/buildDebian.js --platform=arm64"
```

Does that sound OK to you?

jyangchisyan:
That sounds great! 
I am not familiar the JavaScript language features. Could you give me sometime time for rewrite it?

Thanks for your review!

scripts/createPackage.js
```diff
@@ -62,6 +62,12 @@ var platformOptions = {
     platform: 'linux',
     arch: 'armv7l',
     fpm: ['--architecture', 'armhf']
+  },
+  arm64: {
```
PalmerAL, line 66: Maybe this could be renamed to `linuxArm64`, and the other target to `linuxX64`?

jyangchisyan: > Maybe this could be renamed to `linuxArm64`, and the other target to `linuxX64`?

Do we need consider other platform with arm64 architecture?  Maybe the furture mac / windows  platform compiler will auto handle this?

PalmerAL: The Mac ARM build should share the same binary with the x86 one, but if we wanted to add a Windows ARM build, I think it would be separate. And either way, right now we're assuming "arm64" means "Linux", which isn't right.

jyangchisyan: > The Mac ARM build should share the same binary with the x86 one, but if we wanted to add a Windows ARM build, I think it would be separate. And either way, right now we're assuming "arm64" means "Linux", which isn't right.

Yes, we should consider windows ARM build.  I will update it to `linuxArm64` for sperate other system support arm in the furture.

I prefer other linux target use `linuxAmd64`.  AMD64 more popular present.

jyangchisyan: > Maybe this could be renamed to `linuxArm64`, and the other target to `linuxX64`?
If we need also handle linux i386 ,the it renamed to linuxX64 / linuxAmd64 will make sense.
Otherwise the original linux target keep the same name.

PalmerAL: > I prefer other linux target use linuxAmd64. AMD64 more popular present.

Sounds fine to me.


> If we need also handle linux i386 ,the it renamed to linuxX64 / linuxAmd64 will make sense.

Chromium dropped i386 support a while ago, so that won't be needed.

scripts/createPackage.js
```diff
@@ -62,6 +62,12 @@ var platformOptions = {
     platform: 'linux',
     arch: 'armv7l',
     fpm: ['--architecture', 'armhf']
+  },
+  LinuxArm64: {
```
PalmerAL, line 66: This should probably be lowercase.

package.json
```diff
@@ -40,10 +40,11 @@
     "build": "npm run buildMain && npm run buildBrowser && npm run buildPreload",
     "buildWindows": "npm run build && node ./scripts/buildWindows.js",
     "buildMac": "npm run build && node ./scripts/buildMac.js",
-    "buildDebian": "npm run build && node ./scripts/buildDebian.js",
-    "buildRaspi": "npm run build && node ./scripts/buildDebian.js --raspi",
+    "buildDebian": "npm run build && node ./scripts/buildDebian.js --platform=amd64",
+    "buildRaspi": "npm run build && node ./scripts/buildDebian.js --platform=armhf",
+    "buildARM64": "npm run build && node ./scripts/buildDebian.js --platform=arm64",
```
PalmerAL, line 45: Maybe include "linux" in the command name somewhere?

scripts/createPackage.js
jyangchisyan: done

package.json
jyangchisyan: update buildARM64 to buildLinuxArm64

scripts/createPackage.js
jyangchisyan: done

PalmerAL:
Looks good, thank you!

PR #1169, opened 2020-07-26
Bulgarian translation

Syndamia:
Added a translation to Bulgarian language. 

For some stuff, I followed the proven translations from `Firefox`. For others I searched a lot to insure that my choices of words are correct. I have also checked a couple of times for misspellings (*didn't find any in the last check*).


PalmerAL:
Thanks! I don't see anything wrong, but I don't really have any way of judging either; I'll trust that your spelling is right.

Syndamia:
> Thanks! I don't see anything wrong, but I don't really have any way of judging either; I'll trust that your spelling is right.

I checked it by hand and with software (correct spelling) so it should be fine.

PR #1175, opened 2020-07-29
Use openTabsInForeground setting for context menu

### LINKED ISSUE #60: Open in Background Tap Option for Right Clicked links
I often right click on links and select open in new background tap. This is the most important thing missing from Min before I can use it as my default browser. It is very helpful well searching and opening mutiple things.

### LINKED ISSUE #119: Preference: Don't jump to new tab
in 1.1.2, the viewport is hostage to every new tab opened. this shouldn't be the case or ideally, there should be a toggle somewhere that allows opening tabs in the background.

### LINKED ISSUE #1153: Context menu should use the "open new tabs in foreground" preference

Sestowner:
fixes #60, fixes #119, fixes #1153

### DIFF
--- a/js/webviewMenu.js
+++ b/js/webviewMenu.js
@@ -4,6 +4,7 @@ const webviews = require('webviews.js')
 const browserUI = require('browserUI.js')
 const searchEngine = require('util/searchEngine.js')
 const userscripts = require('userscripts.js')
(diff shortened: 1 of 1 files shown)


PalmerAL:
Looks good, thank you!

PR #1176, opened 2020-07-29
Add Paste and Go to address bar context menu

### LINKED ISSUE #564: in the address bar: a "Paste and go" would be welcome
with a right click in the address bar, I would like to be able to paste a URL and then go to that URL.

integrate the function "Paste and go"

Sestowner:
fixes #564


PalmerAL:
Looks good, thanks!

PR #1185, opened 2020-08-07
Add drag and drop support to tab bar

### LINKED ISSUE #94: Enhancement: Drag text to new tab
In Firefox and Chrome, you can highlight text on a website and drag it to your tab bar to create a new tab with that searched in the default search engine. Or if it's a valid URL, it goes to that website.

Could we have support for this in Min?

Sestowner:
fixes #94

### DIFF
--- a/js/navbar/tabBar.js
+++ b/js/navbar/tabBar.js
@@ -3,6 +3,7 @@ const focusMode = require('focusMode.js')
 const urlParser = require('util/urlParser.js')
 const readerView = require('readerView.js')
 const dragula = require('dragula')
+const settings = require('util/settings/settings.js')
 
 const tabEditor = require('navbar/tabEditor.js')
 const progressBar = require('navbar/progressBar.js')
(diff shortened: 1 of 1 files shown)


js/navbar/tabBar.js
```diff
@@ -225,4 +225,18 @@ if (window.platformType === 'mac') {
...
+tabBar.container.addEventListener('drop', e => {
+  e.preventDefault()
+  var data = e.dataTransfer
+  var tab = tabs.add({
+    url: data.files[0] ? 'file://' + data.files[0].path : data.getData('text'),
+    private: tabs.get(tabs.getSelected()).private
+  })
+
+  tabBar.addTab(tab)
+  webviews.add(tab)
```
PalmerAL, line 239: It would probably be better to do this:

```
browserUI.addTab(tabs.add(...), {enterEditMode: false, openInBackground: !settings.get('openTabsInForeground')})
```

So that the other things that addTab does (replacing an existing empty tab if there is one, scrolling the tab into view) work as expected.

Doing that normally would create a circular dependency between tabBar and browserUI, which browserify doesn't like, but I think you can fix it by doing this:

```
require('browserUI.js').addTab(...
```
instead.

Another example of this: https://github.com/minbrowser/min/blob/master/js/navbar/contentBlockingToggle.js#L86

js/webviews.js
```diff
@@ -224,7 +224,8 @@ const webviews = {
           allowPopups: false,
           partition: partition,
           enableWebSQL: false,
-          autoplayPolicy: 'user-gesture-required'
+          autoplayPolicy: 'user-gesture-required',
+          navigateOnDragDrop: true
```
PalmerAL, line 228: Is this actually useful? The only times I remember encountering this is trying to drag a file onto a page to upload it, dragging it to the wrong spot, and accidentally navigating instead. If you think it's useful, I'm fine with adding it, but if no one is specifically requesting it I'd probably leave it out.

PalmerAL:
Thanks!

PR #1202, opened 2020-08-17
Add proxy settings

### LINKED ISSUE #82: min behind a proxy

Sestowner:
fixes #82

### DIFF
--- a/js/util/proxy.js
+++ b/js/util/proxy.js
@@ -1,0 +1,39 @@
+let proxyConfig = {}
+
+const generateProxyRules = pxs => {
+  let rules = []
+
+  Object.keys(pxs).forEach(key => {
+    let ptcl = key.replace(/Host|Port/, '')
+    let port = key.replace('Host', 'Port')
+
+    if (pxs[key] && key.endsWith('Host')) {
+      rules.push(`${ptcl == 'socks' ? pxs.socksVersion || 'socks5' : ptcl}://${pxs[key]}${pxs[port] ? ':' + pxs[port] : ''}`)
+    }
+  })
+
+  return rules.join(';')
+}
+
+settings.listen('proxy', (proxy = {}) => {
+  switch (proxy.type) {
+    case 1:
+      if (proxy.proxies) {
+        proxyConfig = {
+          pacScript: '',
+          proxyRules: generateProxyRules(proxy.proxies),
+          proxyBypassRules: proxy.proxyBypassRules
+        }
+      }
+      break
+    case 2:
+      proxyConfig.pacScript = proxy.pacScript
+      break
+    default:
+      proxyConfig = {}
+  }
+
+  webContents.getAllWebContents().forEach(wc => wc.session && wc.session.setProxy(proxyConfig))
+})
+
+app.on('session-created', session => session.setProxy(proxyConfig))
--- a/localization/languages/en-US.json
+++ b/localization/languages/en-US.json
@@ -157,6 +157,12 @@
         "customSearchEngineDescription": "Replace the search term with %s",
         "settingsKeyboardShortcutsHeading": "Keyboard Shortcuts",
         "settingsKeyboardShortcutsHelp": "Use commas to separate multiple shortcuts.",
+        "settingsProxyHeading": "Proxy",
+        "settingsNoProxy": "No Proxy",
+        "settingsManualProxy": "Manual Configuration",
+        "settingsAutomaticProxy": "Automatic Configuration",
+        "settingsProxyBypassRules": "No proxy for:",
+        "settingsProxyConfigurationURL": "Configuration URL",
         /* app menu */
         "appMenuFile": "File",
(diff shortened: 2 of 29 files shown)


js/util/proxy.js
```diff
@@ -0,0 +1,39 @@
...
+      proxyConfig.pacScript = proxy.pacScript
+      break
+    default:
+      proxyConfig = {}
+  }
+
+  webContents.getAllWebContents().forEach(wc => wc.session && wc.session.setProxy(proxyConfig))
+})
+
+app.on('session-created', session => session.setProxy(proxyConfig))
```
PalmerAL, line 39: Does this event also fire for the app's default session? I remember it only firing for sessions created after the default session (i.e. private tabs), but I could be wrong about that.

localization/languages/en-US.json
```diff
@@ -157,6 +157,12 @@
         "customSearchEngineDescription": "Replace the search term with %s",
         "settingsKeyboardShortcutsHeading": "Keyboard Shortcuts",
         "settingsKeyboardShortcutsHelp": "Use commas to separate multiple shortcuts.",
+        "settingsProxyHeading": "Proxy",
+        "settingsNoProxy": "No Proxy",
```
PalmerAL, line 161: If no proxy is configured in Min, but there's a proxy configured at the OS level, the OS proxy would still be used, right? If so, this should probably be something like "Use system settings".

pages/settings/index.html
```diff
@@ -120,6 +120,52 @@ <h3 data-string="settingsPasswordAutoFillHeadline"></h3>
...
+				<label for="socks-proxy-host-input">SOCKS</label>
+				<input id="socks-proxy-host-input" type="text" name="socksHost" placeholder="proxy.example.com" />
+				<input type="number" id="socks-proxy-port-input" name="socksPort" min="0" placeholder="8080" />
+			</div>
+			<div class="setting-option">
+				<input id="socks4-input" type="radio" value="socks4" name="socksVersion" />
+				<label for="socks4-input">SOCKS v4</label>
+				<input id="socks5-input" type="radio" value="socks5" name="socksVersion" />
+				<label for="socks4-input">SOCKS v5</label>
+			</div>
```
PalmerAL, line 155: If I'm understanding correctly, this could all just be combined into a single text input, so you could type in "http://example.com:8080" directly, which would simplify this code a lot (like I mentioned here: https://github.com/minbrowser/min/pull/820#issuecomment-529302002)

Related to this, it's not clear to me what happens if you fill out more than one of the field sets, so there should probably be some UI or explanation for that. But if there's only a single text input, it would only be possible to set one proxy, so that issue would go away.

pages/settings/index.html
```diff
@@ -120,6 +120,52 @@ <h3 data-string="settingsPasswordAutoFillHeadline"></h3>
...
+
+			<div class="setting-option">
+				<label for="proxy-bypass-rules" data-string="settingsProxyBypassRules"></label>
+				<input name="proxyBypassRules" id="proxy-bypass-rules" placeholder="localhost, 127.0.0.1/8, ::1" />
```
PalmerAL, line 159: I can see this placeholder potentially being confusing, because it implies that localhost is bypassed by default (is it?).

If it's not, the placeholder should probably something more generic, maybe like "hostname or IP address"?

pages/settings/settings.js
```diff
@@ -434,3 +434,57 @@ passwordManagersDropdown.addEventListener('change', function (e) {
...
+})
+
+proxyTypeInput.addEventListener('change', e => {
+  const proxyType = e.target.options.selectedIndex
+  setProxy('type', proxyType)
+  toggleProxyOptions(proxyType)
+})
+proxyInput.forEach(item => item.addEventListener('change', e => setProxy('proxies', { [e.target.name]: e.target.value })))
+proxyBypassInput.addEventListener('change', e => setProxy('proxyBypassRules', e.target.value))
+pacInput.addEventListener('change', () => setProxy('pacScript', pacInput.value))
```
PalmerAL, line 490: Can you split these into multiple lines?

pages/settings/index.html
Sestowner: > ...because it implies that localhost is bypassed by default (is it?).

Yes, localhost is bypassed by default.

localization/languages/en-US.json
Sestowner: With these changes, if No Proxy is selected, Min doesn't use a proxy.

js/util/proxy.js
Sestowner: > Does this event also fire for the app's default session?

Yes, it does.

Sestowner:
> Have you been able to test all these options and confirm that they work?

I've been testing these on Linux Mint 20 and Windows 7, and they work for me.

PR #1217, opened 2020-08-27
Update cs.json

BlueManCZ:
Czech localization updated.


localization/languages/cs.json
```diff
@@ -187,16 +192,16 @@
...
-        "appMenuHistory": null, //missing translation
-        "appMenuDeveloper": "Vývojář",
+        "appMenuBookmarks": "Záložky",
+        "appMenuHistory": "Historie",
+        "appMenuDeveloper": "Nástroje pro vývojáře",
         "appMenuReloadBrowser": "Znovu načíst prohlížeč",
-        "appMenuInspectBrowser": "Prozkoumat prohlížeč",
-        "appMenuInspectPage": "Prozkoumat stránku",
+        "appMenuInspectBrowser": "Prozkoumat DOM a styly prohlížeče",
+        "appMenuInspectPage": "Prozkoumat DOM a styly stránky",
```
PalmerAL, line 200: Could you make sure the meaning of these matches the English version (unless there's a particular reason to alter them)?

PalmerAL:
Everything looks good aside from my one comment!

localization/languages/cs.json
BlueManCZ: I think "Inspect DOM and styles of browser" is more understandable for developers than only "Inspect browser". But I can return it to the previous translation, if you don't agree.

PalmerAL: I think "inspect" is probably sufficient; for comparison, Chrome uses "inspect elements", Firefox uses "inspector", and Safari uses "show web inspector", so I'd prefer to return it to the previous translation.

BlueManCZ: Alright, I returned it to the previous one.

PalmerAL:
Thank you!

PR #1227, opened 2020-09-02
Fix switch-task-button size

BlueManCZ:
On the Linux the SVG image in the button for switching tasks looks affected by interpolation. Other buttons are affected as well, but it is not so obvious.

Default:
Snímek z 2020-09-02 16-03-56

Fixed:
Snímek z 2020-09-02 16-03-31

### DIFF
--- a/css/taskOverlay.css
+++ b/css/taskOverlay.css
@@ -18,10 +18,10 @@
 }
 
 #switch-task-button svg {
-	width: 0.8em;
-	height: 0.8em;
+	width: 15px;
+	height: 12px;
 	display: block;
-    margin-top: 2px;
+	margin-top: 1px;
 }
 #task-overlay {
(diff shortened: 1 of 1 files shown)


Milo123459:
+1 on this.
When is this going to be merged?

lubiedo:
@PalmerAL Do you think we can either merge this (after conflicts are fixed) or close it if not applicable?

PalmerAL:
#1617 also adjusts the switch task button; probably better to merge that one. I'm not sure if it fixes this exact issue though.

The root issue is that the buttons are like 13px tall and misaligned; fixing this is going to require changing more than the switch task button, since the heights should still match.

And @BlueManCZ sorry for not doing anything with this, I think it was on my list at some point and then I got distracted.

Issue #1231, opened 2020-09-05
shift+letter not registering as single letter shortcut


joeysnclr:
* **Min Version: 1.16** 
* **Operating system: macOS 10.15.6** 

## Expected Behavior
When focused on an input single letter shortcuts should not be activated

## Actual Behavior
single letter shorcuts that are saved as shift+letter are activated when focused on input

## To Reproduce
save a shortcut as shift+letter
navigate to a webpage and focus on an input
press the key that would activate the shortcut

PR #1232, opened 2020-09-05
Fix: shift+letter not registering as single letter shortcut

joeysnclr:
Fixes issue #1231 

My approach simply discards "shift+" from the string that is being tested. I am not so experienced with regex so there is probably a better way to exclude shift+ instead of .replace.

### DIFF
--- a/js/keybindings.js
+++ b/js/keybindings.js
@@ -5,35 +5,48 @@ There are three possible ways that keybindings can be handled.
   - If a BrowserView is focused, shortcuts are handled by the before-input-event listener.
   */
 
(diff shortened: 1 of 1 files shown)


Syndamia (approved):
Seems like a fine change. Couldn't find problems after a bit of testing.

PalmerAL:
If you want to use a regex, this should work:

```
/^(shift)?\+?\w$/
```

joeysnclr:
Thank you that regex was exactly what I was looking for.

This bug is now fixed.

PalmerAL:
LGTM, thanks.

Issue #1221, opened 2020-08-31
Can't Mute Tab


iFarhaan:
While playing media in the website the tab can't be **muted**. Neither it shows any option in the **GUI** nor it supports **hotkeys**.

Almost Every Web Browser has this feature. We would like to have this in **Min** too. Such as speaker appearance in tab corner **while playing media** which can be toggled using **GUI** and hotkeys shortcut like **Ctrl+M**.

BenStigsen:
The ability to mute tabs from the GUI is now available in [version 1.17.1](https://github.com/minbrowser/min/releases/tag/v1.17.1)

PR #1233, opened 2020-09-06
Added shortcut to mute a tab

joeysnclr:
In response to Can't Mute Tab Issue #1221
This enables the user to mute a tab with a shortcut. By default it is `shift+mod+m`.

Whenever a tab has muted setting enabled it will show the muted icon next to the tab, even if the tab has no audio playing. The playing speaker icon will only show if a tab is playing audible audio. Otherwise, no icon is shown.

I did run into an issue with the icon being taller than the tab title. Not sure if we want to shrink the icon or keep at the same size as the other icons in the nav bar.

Screen Shot 2020-09-05 at 6 29 35 PM

### DIFF
--- a/css/tabBar.css
+++ b/css/tabBar.css
@@ -215,6 +215,13 @@ body.dark-mode #overlay {
 	font-size: 1.6em !important;
 }
 
+/* audio button */
+.tab-item .tab-audio-button {
+	vertical-align: text-bottom;
+	padding: 0 0.5em 0 0;
+	font-size: 1.1em !important;
+}
+
 /* reader button */
 
 .tab-item .reader-button {
--- a/css/tabBar.css
+++ b/css/tabBar.css
@@ -221,9 +221,6 @@ body.dark-mode #overlay {
 	padding: 0 0.5em 0 0;
 	font-size: 1.1em !important;
 }
-.tab-audio-button.hidden {
-    display: none;
-}
 
 /* reader button */
 
--- a/css/tabBar.css
+++ b/css/tabBar.css
@@ -221,6 +221,9 @@ body.dark-mode #overlay {
 	padding: 0 0.5em 0 0;
 	font-size: 1.1em !important;
 }
+.tab-audio-button.hidden {
+    display: none;
+}
 
 /* reader button */
 
--- a/js/navbar/tabBar.js
+++ b/js/navbar/tabBar.js
@@ -2,6 +2,7 @@ const webviews = require('webviews.js')
 const focusMode = require('focusMode.js')
 const urlParser = require('util/urlParser.js')
 const readerView = require('readerView.js')
+const tabAudio = require('tabAudio.js')
 const dragula = require('dragula')
 const settings = require('util/settings/settings.js')
 
@@ -48,6 +49,7 @@ const tabBar = {
     tabEl.setAttribute('role', 'tab')
 
     tabEl.appendChild(readerView.getButton(data.id))
+    tabEl.appendChild(tabAudio.getButton(data.id))
     tabEl.appendChild(progressBar.create())
 
     // icons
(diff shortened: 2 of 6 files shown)


Syndamia:
I like it, but I think the muted icon should be a button. On other browsers it's a button and it will be pretty nice if you have a background tab playing, whose sound you can switch off without going to it.

Also I feel like it should look like the [`permissionRequests`](https://github.com/minbrowser/min/blob/master/js/navbar/permissionRequests.js) button(s) for consistency sake:

min-mic

I'm not too familiar with the Min code base yet, but I believe that we can just copy and modify some of the code in [`permissionRequests`](https://github.com/minbrowser/min/blob/master/js/navbar/permissionRequests.js) (i'll try to implement it, but for now it is just a suggestion).

joeysnclr:
That’s a good point. I’ll look into making it a button and how the permissionRequests icons are implemented. Thanks for the feedback!

PalmerAL:
It'll be a bit before I have time to fully read through this, but a couple other notes:

* I agree making the icon a button would be nice. You shouldn't need any of the permissions JS code; you can probably just use the same CSS classes.
* The muted and playing buttons could probably just be a single element where you flip the icon classes.
* Electron is removing support for calling methods directly on webContents in the future (and doing that can also be slow), so you want to use `webviews.callAsync` instead.
* You might need to reset `hasAudio` and `muted` when the page URL changes, but I haven't tested it yet.

joeysnclr:
@PalmerAL just seeing this now after I did a refactor for how I implemented it. In my refactor I am basically just handling tab audio the same way reader view is being handled (instead of `permissionRequests` is handled) . Regarding your notes and @Syndamia's suggestions, the icon is now a button and the icon is just being flipped when toggled. I will look into  `webviews.callAsync` and resetting the values on URL change. Thanks for the feedback!

Syndamia:
I just tested it out and it's awesome, good job! However, I would like to bring your attention to two problems.

***
#### Problem 1

Firstly, there is some extra space before the title of the tab:

min-tabs-space
*On top it's my testing environment for this branch, and on bottom it's the latest release.*

#### Why it happens

It seems to be the padding of the icon, which doesn't disappear after the icon does (the padding is in orange):

min-audio-padding
*The padding in normal use*

min-audio-padding-no-audio
min-audio-padding-html
*Padding after there is no sound (Inspect Element doesn't outline the padding, so I also added a screenshot of the element)*

#### Proposed solution

I'm not sure how viable it is, but rather than just removing the icon, can't we just hide the whole element when it shouldn't be seen, and change the icons only when going from unmuted to muted and vise versa? 

Perhaps, we could have a CSS class, named hidden:
```
.hidden {
  display: none !important;
}
```
and add it to the element when we shouldn't show it, and remove it when we should show it.

***

#### Problem 2

The second and final problem is that you can mute tabs that don't play any sound. This can either be done by using the mute shortcut or by changing the url of a muted tab.

min-new-tab-muted

#### Proposed solution

The solution from Problem 1 should take care of the icon showing. That does bring the problem of you accidentaly muting a tab that shouldn't be muted and not knowing about it.

Again, I'm not sure how viable it is, but making the muting system work only when there is audio should suffice. Maybe we could block muting the tab while there is no audio playing (mute only if the `media-started-playing` event was triggered?).

***

Also, as a reminder before the actual merge to master, the `console.log` commands in `tabAudio.js` should be removed.

Otherwise, you have done a really good job with the whole system!

joeysnclr:
@Syndamia Thank you for catching those problems! I think your proposed solutions were great and shouldn't compromise on the functionality so I went ahead and implemented them.

#### Problem 1 (see screenshots)
Removed extra padding by adding a hidden class that is toggled on/off.

##### Before
Screen Shot 2020-09-07 at 9 15 42 AM
##### After
Screen Shot 2020-09-07 at 9 15 56 AM
##### With Icon
Screen Shot 2020-09-07 at 9 16 04 AM

#### Problem 2
User can only toggle mute if tab **is playing audio** or if the tab **is already muted**

Other than that, one last check through from @PalmerAL would be appreciated to make sure there aren't any other matters about the implementation. Once again, thank you for your input!

Syndamia:
I'm happy I could help. I did some testing on my end and everything seems perfect. Thank you for your contribution to the project!

PalmerAL:
I'm really busy this week, and I haven't had any time to read through your comments or look at the code again, but I'll probably get to it this weekend!

PalmerAL:
Everything looks good to me, thanks for working on this!

In general, we need a better way to store data related to a tab that shouldn't be saved and restored - this would apply to `muted ` and `hasAudio`, and also to existing properties like `readerable`. But I think it's fine to leave it as-is for now.

PR #1253, opened 2020-09-27
Update zh-TW Translation

z-huang:


localization/languages/zh-TW.json
```diff
@@ -206,7 +206,7 @@
         "appMenuKeyboardShortcuts": "快速鍵",
         "appMenuReportBug": "提交 Bug",
         "appMenuTakeTour": "導覽",
-        "appMenuViewGithub": "在 GitHub 上查看",
+        "appMenuViewGithub": "去 GitHub 查看 Min",
```
PalmerAL, line 209: The original seems closer to the English text to me; is there a reason for changing it?

z-huang: I think that users will be confused with what to view on GitHub, so I changed it to "View Min on GitHub" in traditional Chinese. Is that necessary or not?

PalmerAL: That does sound better to me (actually, I think it could be even more specific, maybe something like "Contribute to Min on GitHub"). But if we change that, we need to update the other language files to match as well.

If you wanted to do that, you could change the English version, and then update all the other language files to have `null //missing translation`.

z-huang: Well, that may be a major modification. Personally, we can just leave it unchanged until a better word comes up.
Also, I've thought of a better translation of "View on GitHub" with the same meaning in traditional Chinese, so I'll make a new pull request with that.

z-huang: It seems that this pull request auto updates itself.

PalmerAL:
Looks good, thank you!

PR #1294, opened 2020-10-25
Add German Translations

Nesaijn:
There were some keys missing which were present in the [en-US.json](https://github.com/minbrowser/min/blob/master/localization/languages/en-US.json#L239) starting at line 239 to the end of the file, which I added.
I also changed some translations which were already present, because I thought that I have found a better translation and also because of consistency.
This is my first time making a pull request. Did I miss something?

### DIFF
--- a/localization/languages/de.json
+++ b/localization/languages/de.json
@@ -239,6 +239,7 @@
         /* Autofill settings */
         "settingsPasswordAutoFillHeadline": "Passwort Automatisch Ausfüllen",
         "settingsSelectPasswordManager": "Wähle eines der unterstützten Passwortmanager:",
+        "keychainViewPasswords": null, //missing translation
         /* Password manager setup */
         "passwordManagerSetupHeading": "Beende die Konfiguration von %p um die automatische Ausfüll Funktion zu nutzen",
         "passwordManagerSetupStep1": {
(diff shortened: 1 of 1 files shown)


PalmerAL:
Everything looks good to me!

I just merged another PR (#1283) that added some additional strings for the password manager; I added spaces for them to the file, and if you could translate those as well that would be great.

Also, totally unrelated to this PR, but I've been studying German a bit recently, so if you don't mind, I have a couple of small grammar questions:
* For settingsCustomizeFiltersLink, is there a reason for switching from a verb form in English ("Customize filters") to a noun form ("Benutzerdefinierte Filter", which I guess is "user-defined filters")?
* In "Benutzerdefinierte Skripte erlauben es das Verhalten von Webseiten zu modifizieren", what is "es" referring to?

Nesaijn:
Done!
Regarding `passwordCaptureSavePassword` I instead of just asking to save the password, I asked to save the login data/information since it seemed appropriate and Firefox also seems to be doing it like this.

Sure! I will try to explain.
- This one about `settingsCustomizeFiltersLink` is just a mistake of mine. Thanks for reviewing. :D
- "es" is referring to "Benutzerdefinierte Skripte". In English putting a "es" there would make no sense, but in German it does sound better. 
Like the sentence I just used. I could have said at the end "(...), but in German it sounds better." but saying it like I did "(...), but in German it **does** sound better.". It also gives it a bit more weight. But the main reason is the sounding I guess. Both versions, with and without "es" would be viable, but using the former is a better choice.

PalmerAL:
I think everything looks good, and thanks for the explanation!

> "es" is referring to "Benutzerdefinierte Skripte". Actually after giving it some thought, it seems that " es" is kind of a part of "erlauben" but I am not quit sure. Because you could also say "Es erlaubt es (...)". In English putting a "es" there would make no sense, but in German it does sound better.

I'm reading this almost like it has a missing "dass", like "Userscripts permit it, that [they] modify...", but that doesn't make much sense written out in English either. I guess you're probably right that I should just treat it as an emphasis word and not try too hard to find a literal meaning.

PalmerAL:
Actually, I missed one string from the password manager PR:
```
         "deletePassword": "Delete password for %s?",
```

Could you translate that as well?

Nesaijn:
You are welcome. :)

Regarding that missing string, I don't know how to add that right know. :/
Do I need to start a new pull request? But then how since I can't find the de.json file where it has the recent changes. 
Maybe you could add them:

`"deletePassword": "Zugangsdaten für %s löschen?"`

And could you also make a change to `keychainViewPasswords` for consistency reasons?
`"keychainViewPasswords": "Gespeicherte Zugangsdaten anzeigen"`

PalmerAL:
The normal way to do this is to follow [these instructions](https://www.freecodecamp.org/news/how-to-sync-your-fork-with-the-original-git-repository/) and then make a new PR, but it's somewhat annoying to do, so I added them both in 9ef7a639c7aebf544af6392af16811921c0b3d01.

Thank you!

Nesaijn:
You are welcome! Thank you for your work and effort!

PR #1298, opened 2020-10-27
Update ru.json

KagamineP:


PalmerAL:
I think you set a new record for the fastest translation update :)

Thanks!

PR #1313, opened 2020-11-05
use prettier, update all dependencies

Milo123459:
uses [prettier](https://npm.im/prettier)
updates all dependencies to there latest published version, uses [deepercheck](https://npm.im/deepercheck)

thanks :)


Milo123459:
bump
@PalmerAL 
the code style for prettier is very maintained, I suggest every now and then running deepercheck.
https://npm.im/deepercheck
(I have created this, if it bugs, please contact me in the min discord! I'm Salvage_Dev.)

ext/abp-filter-parser-modified/abp-filter-parser.js
```diff
@@ -1,106 +1,123 @@
...
+  subdocument: 64,
+  ping: 128,
+  websocket: 256,
+  webrtc: 512,
+  document: 1024,
+  elemhide: 2048,
+  generichide: 4096,
+  genericblock: 8192,
+  popup: 16384,
+  other: 32768,
```
Syndamia, line 38: I'm going through the changes. I don't know a lot about JavaScript, but is it really OK to remove the quotes on (most of) the keys?

ext/abp-filter-parser-modified/abp-filter-parser.js
```diff
@@ -110,382 +127,419 @@ function parseOptions (input) {
       */
 
       // the option is an element type to skip
-      if (option[0] === '~' && elementTypes.indexOf(option.substring(1)) !== -1) {
+      if (
+        option[0] === "~" &&
+        elementTypes.indexOf(option.substring(1)) !== -1
+      ) {
```
Syndamia, line 133: This looks uglier in my opinion

PalmerAL:
@Milo123459  Most of the code is already formatted with [standard](https://www.npmjs.com/package/standard); I'm not sure I see the value in switching to a different formatter. If might be better to focus on fixing areas where the formatting hasn't been applied, and adding a `format` command to make it easier for contributors to format changes. (I have a standard-format plugin in my editor, but I don't think it's documented anywhere to do that).


Auto-updating dependencies to a new major version generally isn't a good idea, because they can have breaking changes. You'd need to go through each component that uses these and re-test them with the new package version.

PR #1236, opened 2020-09-10
Added escaped quotes to concurrently script

icedjuro:
For some reason, when I run `npm run start`, I get an error that '.' is not defined. It seemed that npm had trouble running `./node_modules/.bin/concurrently` for whatever reason. I ran this command through PowerShell on Windows 10. Enclosing the command in escaped quotes fixed the issue. It is very likely that I am doing something wrong as I would think a Windows developer would have encountered this immediately.

PR #1319, opened 2020-11-06
npm run start path not recognized

BenStigsen:
'.' is not recognized as an internal or external command, operable program or batch file.
This seems to be fixed by adding escaped quotation marks.
This is the same fix as proposed in the PR https://github.com/minbrowser/min/pull/1236


PalmerAL:
Thanks.

PR #1323, opened 2020-11-09
update readme

Milo123459:
I added a badge that displays how many downloads, also, 400k downloads! Amazing!


Syndamia (changes requested):
The only thing I suggest is that there is no text for the image, so from `Downloads it becomes `. This way, if for some reason the image breaks, there won't be just a random "Downloads" text.

Milo123459 (commented):
looks good

PalmerAL:
👍

PR #1324, opened 2020-11-09
update filters

Milo123459:
Update filters via running `npm run updateFilters` 😎


PalmerAL:
I usually just run this before a release, so I don't think there's a need to make PRs for it in between. Good point though!

PR #1325, opened 2020-11-09
update package json scripts

Milo123459:
Hello! I know this was previously "fixed" - but it wasn't. 
This changes using `node_modules/.bin/concurrently` *just an example* to using `concurrently` *still an example*

This allows all scripts to be ran on windows now but requires that you have to have the NodeJS path variable configured correctly.

Thanks! 🎉

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,10 +34,10 @@
   },
   "license": "Apache-2.0",
   "scripts": {
-    "test": "node_modules/.bin/standard --verbose js/**/*.js main/*.js",
+    "test": "standard --verbose js/**/*.js main/*.js",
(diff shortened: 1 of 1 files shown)


PalmerAL:
Wait, so the `start` command still doesn't work even with the quotes?

Milo123459:
@PalmerAL Doesn't, no.

PalmerAL:
OK, I tried the original and this change again on windows, and this does seem to fix it. Thanks!

(I'm almost certain that the command worked before on Windows after I added quotes, which is strange).

Milo123459:
Glad it worked :)

PR #1338, opened 2020-11-15
fix password manager bug

### LINKED ISSUE #1336: Cannot select password auto-fill as none in 1.17
* **Min Version:** 1.17
* **Operating system:** Linux Mint 20 Cinnamon (Based on Ubuntu 20)

## Expected Behavior

Should be able to set the Password Auto-Fill to `none`

image

## Actual Behavior

Password Auto-Fill chooses `Built-in password manager` when I try to select none. It works fine with other password managers.

Peek 2020-11-15 10-05

## To Reproduce

Open the settings, go to `Password Auto-Fill` and try to select none from the drop down menu.

Milo123459:
Fixes #1336

### DIFF
--- a/pages/settings/settings.js
+++ b/pages/settings/settings.js
@@ -448,7 +448,7 @@ for (var manager in passwordManagers) {
 }
 
 settings.listen('passwordManager', function (value) {
-  passwordManagersDropdown.value = currentPasswordManager.name
+  passwordManagersDropdown.value = value ? value.name : 'none'
 })
 
 passwordManagersDropdown.addEventListener('change', function (e) {


Syndamia (approved):

PalmerAL:
This isn't completely the right fix; `currentPasswordManager` reflects whatever manager is being used internally (by autofill, by the password viewer, etc.), so the dropdown's value should actually reflect that.

The actual issue is that lines 454-456 do this:
```
passwordManagersDropdown.addEventListener('change', function (e) {
  if (this.value === 'none') {
    settings.set('passwordManager', null)
```

So if you select "none" in the dropdown, the setting value will be set to "null". That causes the rest of the code to assume that you've never modified the setting, (since settings you've never changed start out as `undefined`), and so they use the default value of the setting, which is to use the built-in password manager.

I've made a PR with what I think is the correct fix here: #1339

Milo123459:
Both solutions work, up to you which one you merge.

Milo123459:
Actually, I say merge both - Fix visually & internally so anyone using an old format (with the value of null) will display none still.
@PalmerAL What do you think?

PalmerAL:
The problem with "just display none in the dropdown" is that the rest of the code will still think that there's a password manager enabled, so you'll still see the key icon in input fields, you'll still see the option to save a password, etc., which isn't right.

The only way the setting could be `null` is if you enabled Bitwarden or 1Password and then disabled them in a previous release. I think in that case it'd be fine for the built-in password manager to be enabled after you upgrade, and then you could go into settings and disable it again if you don't want it.

And if your setting is any other value, then by fixing the underlying data, the second one will make the dropdown work correctly as well.

So I'd suggest just merging the second one.

Syndamia:
You know, I feel like Palmer's fix is more proper. I say lets go with that, rather than this.

Milo123459:
Sounds good.

Milo123459:
Before I close this, I just want to address that if someone has null as a value in the password manager setting, it'll break. (It won't display correctly)

PalmerAL:
Wait, what do you see on the settings screen if the setting is set to null? If I try that, I just see that the built-in manager becomes enabled by default, which is what I would expect.

Milo123459:
I see the same thing as you, palmer.

Issue #1331, opened 2020-11-12
key icons appearing randomly


shalva97:
* **Min Version:** 1.16.0 but probably it is 1.16.1 because i have master branch
* **Operating system:** Arch

## Expected Behavior
I should only see key icons for login pages

#### on Github:
![image](https://user-images.githubusercontent.com/22417494/98943819-89194480-2509-11eb-97bd-95bf7c0d802a.png)

#### on my locally hosted qbittorrent:
![image](https://user-images.githubusercontent.com/22417494/98943892-a2ba8c00-2509-11eb-9402-4de2ebd85b9f.png)

PalmerAL:
For the second two, can you right click > inspect element, and post the HTML for the input elements?

shalva97:
from qbittorrent

```
<table style="margin: auto;">
                <tbody><tr>
                    <td>
                        <label for="autoTMM">Torrent Management Mode:</label>
                    </td>
                    <td>
                        <select id="autoTMM" name="autoTMM" onchange="qBittorrent.Download.changeTMM(this)">
                            <option selected="" value="false">Manual</option>
                            <option value="true">Automatic</option>
                        </select>
                    </td>
                </tr>
                <tr>
                    <td>
                        <label for="savepath">Save files to location:</label>
                    </td>
                    <td>
                        <input type="text" id="savepath" name="savepath" style="width: 16em;">
                    </td>
                </tr>
                <tr>
                    <td>
                        <label for="rename">Rename torrent</label>
                    </td>
                    <td>
                        <input type="text" id="rename" name="rename" style="width: 16em;">
                    </td>
                </tr>
                <tr>
                    <td>
                        <label for="category">Category:</label>
                    </td>
                    <td>
                        <div class="select-watched-folder-editable">
                            <select id="categorySelect" onchange="qBittorrent.Download.changeCategorySelect(this)">
                                <option selected="" value="\other"></option>
                            </select>
                            <input name="category" type="text" value="">
                        </div>
                    </td>
                </tr>
                <tr>
                    <td>
                        <label for="startTorrent">Start torrent</label>
                    </td>
                    <td>
                        <input type="hidden" id="startTorrentHidden" name="paused">
                        <input type="checkbox" id="startTorrent">
                    </td>
                </tr>
                <tr>
                    <td>
                        <label for="skip_checking">Skip hash check</label>
                    </td>
                    <td>
                        <input type="checkbox" id="skip_checking" name="skip_checking" value="true">
                    </td>
[...]

PR #1345, opened 2020-11-17
fix key icons appearing randomly

BenStigsen:
https://github.com/minbrowser/min/issues/1331

### DIFF
--- a/js/preload/passwordFill.js
+++ b/js/preload/passwordFill.js
@@ -122,8 +122,7 @@ function getUsernameFields () {
 
 // Shortcut to get password fields from a page.
 function getPasswordFields () {
-  // return getInputs(['pass', 'password'], ['password'])
-  return document.querySelectorAll('input[type=password]')
+  return Array.from(document.querySelectorAll('input[type=password]'))
 }
 
 // Removes credentials list overlay.
--- a/js/preload/passwordFill.js
+++ b/js/preload/passwordFill.js
@@ -84,7 +84,7 @@ function createUnlockButton (input) {
 function checkAttribute (element, attribute, matches) {
   const value = element.getAttribute(attribute)
   if (value == null) { return false }
-  return matches.some(match => value.toLowerCase().includes(match))
+  return matches.includes(value.toLowerCase())
 }
 
 // Gets all input fields on a page that contain at least one of the provided
@@ -118,7 +118,7 @@ function getUsernameFields () {
 
 // Shortcut to get password fields from a page.
 function getPasswordFields () {
-  return getInputs(['pass'], ['password'])
+  return getInputs(['pass', 'password'], ['password'])
 }
 
 // Removes credentials list overlay.
--- a/js/preload/passwordFill.js
+++ b/js/preload/passwordFill.js
@@ -79,12 +79,12 @@ function createUnlockButton (input) {
   return unlockDiv
 }
 
-// Tries to find if an element has a specific attribute value that contains at
+// Tries to find if an element has a specific attribute value that is like at
 // least one of the values from 'matches' array.
 function checkAttribute (element, attribute, matches) {
   const value = element.getAttribute(attribute)
   if (value == null) { return false }
-  return matches.includes(value.toLowerCase())
+  return value.split(/[\W_]/g).some(word => matches.includes(word.toLowerCase()))
 }
 
 // Gets all input fields on a page that contain at least one of the provided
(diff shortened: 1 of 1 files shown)


PalmerAL:
Have you tested this on many sites with login forms that should be detected? If I'm reading this code correctly (and it's been a while since I looked at this, and it's kind of confusing me), it would only match username inputs where the attribute is exactly one of "user", "name", "mail", etc. So IIUC something like this wouldn't be detected:

```
<input id="username-input" type="text"/>
```

I think the regex-based solution you mentioned might be better - you probably want to break the attribute value on non-letter characters and see if any of the individual words match a list entry.

BenStigsen:
I didn't even think about IDs like that. Yeah you're right, this wouldn't get a lot of password and username fields. I'll change it.

Milo123459:
@BenStigsen any news?

BenStigsen:
@Milo123459 I'm trying to get the task reassigned to someone other than me 😅. The problem and solution has been described here: https://github.com/minbrowser/min/pull/1345

Milo123459:
Ah, that's fine 👍

BenStigsen:
The issue has been fixed (it seems) I'll do some testing and merge later today if the behaviour seems to be correct.

BenStigsen:
Ah, while it fixes the keyicon showing up random places, it does remove it from sites like Facebook or Reddit.

BenStigsen:
For now this works. But perhaps it should use regex in the array of username names/IDs.
So that instead of:
`[..., 'user', 'name', 'username', ...]`
it becomes:
`[..., /(user(name)?)/g, ...]`

Milo123459:
Why not create a regex that checks the ID of elements and if they include password or user. Might be easier.

BenStigsen:
Because this was done before, which made the keylock icon appear in random places. 
In the list of username fields, one of the things that are looked for is `"auth"`, but in the issues section on GitHub, there's an `author` field, which makes the keylock icon show.

These random popups have been fixed (it seems), and there's not really an issue anymore, but I was just thinking about simplifying the code a bit even though it doesn't change any functionality. That's a separate PR though if we choose to do it like that.

js/preload/passwordFill.js
```diff
@@ -101,7 +101,8 @@ function getInputs (names, types) {
 
     // We expect the field to have either 'name', 'formcontrolname' or 'id' attribute
     // that we can use to identify it as a login form input field.
-    if (checkAttribute(field, 'name', names) ||
+    if (field.type == "password" ||
+        checkAttribute(field, 'name', names) ||
```
PalmerAL, line 105: This would also return the password field for getUsernameFields() I think; it'd probably be better to remove this and just rewrite `getPasswordFields` to be:

```
return document.querySelectorAll('input[type=password]')
```

PalmerAL:
I think the approach looks good; I'll try it out for a few days to see if I notice any compatibility issues and then it should be fine to merge.

js/preload/passwordFill.js
BenStigsen: > This would also return the password field for getUsernameFields()
Yeah I thought about that. 

Changing it to what you suggested seems better though.

Milo123459:
Shouldn't the target branch be master?

BenStigsen:
You ask why it's set to the 1.17.x branch. To that I will answer... I don't even know. It has been changed.

PalmerAL:
I went through the passwords I have saved, and found a couple sites where this breaks:

* My [University's SSO system](https://connect.wustl.edu/login/wulogin.aspx):
```
<input name="ucWUSTLKeyLogin$txtUsername" type="text" id="ucWUSTLKeyLogin_txtUsername" tabindex="10" autocapitalize="none" autofocus="">
```

* [GroupMe](https://web.groupme.com/signin):
```
<input type="text" class="form-control ng-pristine ng-valid ng-empty ng-touched" id="signinUserNameInput" placeholder="Email or Phone Number" ng-model="username">
```

The two options I can think of for fixing this are:
* Use a regex that looks for lowercase letter + uppercase letter and treats it as a word boundary (so "signinUserNameInput" becomes "signin user name input")
* Go back to the old approach of simple string matching for things like "name", but require whole-word matches for just "auth", which would fix the github issue.

BenStigsen:
Hmm, well the issue isn't only on GitHub. The regex word boundary check would probably be better.

PalmerAL:
Sounds fine to me.

PR #1366, opened 2020-12-01
Lint

Milo123459:
This just lints JS with standard, css with prettier

Adds command lint


PalmerAL:
Files in `/ext` need to be excluded from formatting because they're imported from external repositories. Other than that, I think everything looks good.

Milo123459:
> Files in `/ext` need to be excluded from formatting because they're imported from external repositories. Other than that, I think everything looks good.

+1, fixed

PalmerAL:
I'm guessing the JS files in /ext would need to be excluded as well; prettierignore only applies to CSS, right?

The existing `test` command is this:
```
    "test": "standard --verbose js/**/*.js main/*.js",
```

So I think you could use the same path arguments for the lint command.

Also, can you revert the formatting changes in the `ext` directory?

Milo123459:
The prettierignore ignores CSS linting, as thats what prettier is doing, linting css. I'll update the script, and reverting it might take a bit but I'll get to work.

Milo123459:
@PalmerAL all fixed, you can go over it if you want

PalmerAL:
Looks good to me!

PR #1373, opened 2020-12-09
lint & update standard

Milo123459:
Lint with an updated standard

I have a problem with linting things, like seriously, I need help!

Haha, I'll manage the lint PRs

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,8 +9,6 @@
   "dependencies": {
     "dexie": "^1.2.0",
     "dragula": "^3.6.2",
-    "electron-installer-debian": "^3.1.0",
-    "electron-installer-redhat": "^3.2.0",
     "electron-squirrel-startup": "^1.0.0",
     "keytar": "^7.0.0",
     "mousetrap": "^1.5.3",
--- a/package.json
+++ b/package.json
@@ -9,6 +9,8 @@
   "dependencies": {
     "dexie": "^1.2.0",
     "dragula": "^3.6.2",
+    "electron-installer-debian": "^3.1.0",
+    "electron-installer-redhat": "^3.2.0",
     "electron-squirrel-startup": "^1.0.0",
     "keytar": "^7.0.0",
     "mousetrap": "^1.5.3",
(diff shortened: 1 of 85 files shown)


Milo123459:
@PalmerAL LG?

package.json
```diff
@@ -9,6 +9,8 @@
   "dependencies": {
     "dexie": "^1.2.0",
     "dragula": "^3.6.2",
+    "electron-installer-debian": "^3.1.0",
+    "electron-installer-redhat": "^3.2.0",
```
PalmerAL, line 13: These are already in optionalDependencies, and they need to stay there, because they aren't (weren't?) installable on some OS's.

Milo123459: I don't remember changing that but 1 second

Milo123459: Fixed.

Milo123459:
> Yeah it looks good aside from my comment, I've just been really busy. But this is probably enough linting for now :)

Haha, fixed it, anything else?

PalmerAL:
Looks good!

PalmerAL:
This prevented the browser from running (#1378), so I've reverted it in 6cc3d81d3e3f7d1f7f3e095bcbf9d09eb91bc82a.

The cause of the issue is that some of the files in Min have been converted to modules (that you can load with require()), but others haven't been converted yet. All these files just get concatenated together and run on startup; You can see the list here: https://github.com/minbrowser/min/blob/master/scripts/buildBrowser.js#L13

The issue happens when two legacy files both load a file that's been converted to a module. For example, sessionRestore and taskOverlay.js both have this:
```
var webviews = require('webviews.js')
```

At the top of the file. Previously, one would just overwrite the other, but since they both reference the same thing it would still work. However, newer versions of standard convert these to `const`s, because it thinks these variables are only used a single file, and doesn't know they're combined with other files. Once the files are combined together, there are two definitions of the same `const` in a single context. Since you can't redefine consts, this causes an error.

The actual fix here is to convert the legacy files to modules (and that's a good idea in general - the current system makes no sense). However, in the meantime, we need to keep using the old version of standard.

Sorry for not noticing this earlier.

PR #1417, opened 2021-01-04
lint with some semi colons

Milo123459:
Referring to palmers comment in my last lint PR, he said "The lack of semicolons in the JS files is confusing me!" so I took matters into my own hand and now we are using semistandard 🎉

### DIFF
--- a/js/autofillSetup.js
+++ b/js/autofillSetup.js
@@ -1,30 +1,30 @@
-const setupDialog = require('passwordManager/managerSetup.js')
-const settings = require('util/settings/settings.js')
-const PasswordManagers = require('passwordManager/passwordManager.js')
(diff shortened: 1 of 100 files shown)


shalva97:
why would lack of semicolons be confusing? is not it just a boring boilerplate code that is only useful for interpreter?

Milo123459:
Ask palmer, also, semicolons are amazing :)

shalva97:
why are they amazing?

Milo123459:
Who knows

PalmerAL:
I wasn't being too serious with that comment, sorry for not making that more clear.

Anyway, I think the style is good enough for now. Maybe there is a more ideal style (and maybe semicolons would be better), but it's not worth reformatting everything continously in search of something better.

PR #1418, opened 2021-01-05
Add usage statistics collection

PalmerAL:
This PR adds a way to collect usage statistics, which allows us to answer questions like "How many users does Min have?", and "which operating systems are most frequently used?". Eventually, this can also be expanded to include things like "How often is feature X used?", which should make it easier to prioritize what to work on.

I'm going to start writing a document soon that describes what's collected in more detail (which will be linked to from the settings page), but in summary:

* We won't collect anything that can be used to identify individual users.
* We won't collect any browsing history, passwords, or other data stored in Min.
* There's an option to opt-out on the settings page.

Todo:

 - [x] Add app version to collected data

### DIFF
--- a/docs/statistics.md
+++ b/docs/statistics.md
@@ -1,0 +1,17 @@
+## Usage statistics collection in Min
+
+By default, Min sends some statistics about how it's used. Currently, these include:
+
+* Your operating system and computer language
+* When you installed Min
+* The version of Min that you're using
+* How often you use certain features in Min
+* An anonymous ID representing your installation of Min
+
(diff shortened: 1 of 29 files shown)


Milo123459:
@PalmerAL Cool idea, will this be enabled by default?

PalmerAL:
I think so; given the data that's being collected I don't think there's a need to prompt about it.

Milo123459:
Makes sense. Can you check Discord?

PalmerAL:
I saw your message, I'll get to it in a few minutes.

PR #1438, opened 2021-01-16
update deps

Milo123459:
Hey 👋,

I've updated all the deps,

I've checked all the docs for breaking changes and everything still works.

I linted the code because standard updated.

:)

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,8 +9,6 @@
   "dependencies": {
     "dexie": "^3.0.3",
     "dragula": "^3.6.2",
-    "electron-installer-debian": "^3.1.0",
-    "electron-installer-redhat": "^3.2.0",
     "electron-squirrel-startup": "^1.0.0",
     "keytar": "^7.0.0",
     "mousetrap": "^1.5.3",
--- a/package.json
+++ b/package.json
@@ -9,27 +9,29 @@
   "dependencies": {
     "dexie": "^3.0.3",
     "dragula": "^3.6.2",
+    "electron-installer-debian": "^3.1.0",
+    "electron-installer-redhat": "^3.2.0",
     "electron-squirrel-startup": "^1.0.0",
     "keytar": "^7.0.0",
     "mousetrap": "^1.5.3",
-    "pdfjs-dist": "2.4.456",
+    "pdfjs-dist": "^2.5.207",
     "regedit": "^3.0.3",
     "string_score": "^0.1.22"
   },
   "devDependencies": {
-    "archiver": "^4.0.1",
-    "browserify": "^16.5.1",
-    "chokidar": "^3.4.0",
+    "archiver": "^5.2.0",
+    "browserify": "^17.0.0",
+    "chokidar": "^3.5.1",
     "concurrently": "^5.2.0",
     "decomment": "^0.9.0",
-    "electron": "^11.1.0",
+    "electron": "^11.2.0",
     "electron-installer-windows": "^3.0.0",
     "electron-packager": "^15.1.0",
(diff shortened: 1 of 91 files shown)


PalmerAL:
This has the same problem as https://github.com/minbrowser/min/pull/1373#issuecomment-744136807, I think? The conclusion from that was that we can't update standard until we finish converting the rest of the old files into modules.

Besides that I think the other updates are fine.

package.json
```diff
@@ -9,27 +9,29 @@
   "dependencies": {
     "dexie": "^3.0.3",
     "dragula": "^3.7.3",
+    "electron-installer-debian": "^3.1.0",
+    "electron-installer-redhat": "^3.2.0",
```
PalmerAL, line 13: These should only be listed in `optionalDependencies` I think, I'm not sure why npm is moving them (although it did the same thing to me last time I installed).

Milo123459: Got it.

Milo123459:
> This has the same problem as [#1373 (comment)](https://github.com/minbrowser/min/pull/1373#issuecomment-744136807), I think? The conclusion from that was that we can't update standard until we finish converting the rest of the old files into modules.
> 
> Besides that I think the other updates are fine.

I mean I'd say set your node version higher :) 14.0.0> should be fine.

Milo123459:
:o This is a lot of conflicts! I'll have to get this fixed! 😰

Milo123459:
@PalmerAL Hey, running yarn start now opens the browser, no errors but I can't press any buttons. Looking into it. BTW, Standard broke like 5 files and I've had to fix them manually.

Milo123459:
Also, standard has formatted everything really wierdly. I'm seriously thinking we should switch to prettier for JS linting.

Milo123459:
Can't open devtools by running CTRL+ALT+SHIFT+I

PalmerAL:
> Standard broke like 5 files and I've had to fix them manually.

This seems bad? If anyone reruns lint, wouldn't they break again? And telling people they have to manually undo some of Standards changes every time they lint isn't a good solution.

I really think that we should just leave the formatting as-is and merge the other dependency updates. The code formatting looks fine to me now after the other PRs, and it's not worth it to constantly reformat everything and break things in exchange for possibly making it look a bit nicer.

Milo123459:
Edit, it was you! (Not trying to throw shade), but I do think we should use prettier for js linting. Reasoning is that it just didn't lint the file.

PalmerAL:
Switching linters again doesn't seem worth it; let's either figure out why Standard is breaking and fix it, or just leave it as-is (which I think is totally fine as a solution, the code is already linted reasonably well).

Milo123459:
Okay, lets leave as is. I'd like to go back to the fact where when I'm starting it, I can't press buttons and can't open dev tools with CTRL+ALT+SHIFT+I

PalmerAL:
Is that with the linting changes, or without? My guess is that the linting probably broke something that prevents those functions from working.

Milo123459:
^^ Nope, it is your module stuff.

PalmerAL:
The app still starts for me, so I'm guessing it's probably not that. Try using the master branch instead of the branch from your PR. Also, you'll need to run `npm install` again to downgrade standard after changing package.json, if you haven't done that already.

Milo123459:
Hmm. Let me try that.

Milo123459:
Already done and same issue.

Milo123459:
Works on master. What a pain.

Milo123459:
this is just annoying me so im closing this

PR #1442, opened 2021-01-17
Fix opening local files

### LINKED ISSUE #1295: Files don't open
* **Min Version:** 1.16.0 (although I think it did happen on 1.15.0)
* **Operating system:** Linux Mint Cinnamon 20 (based on Debian 20)

## Expected Behavior

When opening a (`.html` or `.pdf`) file with Min, it should show up.

## Actual Behavior

The Min window comes up, but nothing happens. 

In the following GIF, I tried to open a `.html` file via double-clicking, opening with right-click menu and terminal (the file isn't blank, as you'll see a bit later in the issue):
Peek 2020-10-25 18-10

## Workaround

I've found that if I can get the file link, I can just paste it in Min and it works fine.

In the following GIF I open the same file as before in Brave browser (from right-click and open with, which doesn't show up on the GIF), copy the link and paste it into Min, which doesn't give me any problems:
Peek 2020-10-25 18-22

Sestowner:
fixes #1295

### DIFF
--- a/main/main.js
+++ b/main/main.js
@@ -111,10 +111,10 @@ function handleCommandLineArguments (argv) {
           sendIPCToWindow(mainWindow, 'addTab', {
             url: arg
           })
-        } else if (/[A-Z]:[/\\].*\.html?$/.test(arg)) {
-          // local files on Windows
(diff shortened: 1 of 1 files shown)


PalmerAL:
LGTM, sorry for slow review.

PR #1448, opened 2021-01-19
Convert taskOverlay to a module

PalmerAL:


js/navbar/menuButton.js
```diff
@@ -1,3 +1,5 @@
+var keybindings = require('keybindings.js')
```
PalmerAL, line 1: This was a pre-existing issue where `keybindings` was being declared from taskOverlay, which was wrong - with taskOverlay as a module, that doesn't happen anymore.

js/taskOverlay/taskOverlay.js
```diff
@@ -89,7 +120,27 @@ window.taskOverlay = {
...
+
+          // if there are no tabs left, remove the task
+          if (task.tabs.count() === 0) {
+            // remove the task element from the overlay
+            getTaskContainer(task.id).remove()
+            // close the task
+            browserUI.closeTask(task.id)
+          }
+        }
+      })
```
PalmerAL, line 143: Previously, there was a loop where taskOverlay depended on taskOverlayBuilder, and taskOverlayBuilder depended on taskOverlay, which it isn't possible to do any more. I fixed this by passing functions into the builder instead.

Milo123459 (approved):

Sestowner (approved):

PR #1450, opened 2021-01-20
Added textarea for ads and trackers whitelist in settings

### LINKED ISSUE #1427: single line for filter exceptions is  not enough
**Can you describe the problem you're facing?**
it is awkward to have just a single line for so many sites I have to add
image

**What solution would you like?**
multi line input field. and also disable grammar check...

icedjuro:
Note: Spellcheck **should** be disabled by default for textarea, but I think it's safer to include in case it ever changes.
Fixes #1427

### DIFF
--- a/css/base.css
+++ b/css/base.css
@@ -91,7 +91,8 @@ a {
   color: dodgerblue;
 }
 
-.dark-mode input {
+.dark-mode input,
+.dark-mode textarea {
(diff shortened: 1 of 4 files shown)


PalmerAL:
I modified the input so that it starts out the same height as the old input, and expands as you add more domains. Other than that, LGTM.

Issue #1322, opened 2020-11-07
Add localization support for keyboard shortcut names


PalmerAL:

Milo123459:
Whilst this has nothing to do with me as I don't translate thing, I'd be happy to just go around the files an add null to everything, for the keybindings. Let me know If I should do this 👍

PR #1460, opened 2021-01-25
Translation for the sending usage statistics

PEPERSO:
I've noticed that the "Custom bangs", and same for the shortcut section at the end of the preferences is not in the translation file, I'll be happy to translate it if you add it there :)

### DIFF
--- a/localization/languages/fr.json
+++ b/localization/languages/fr.json
@@ -159,7 +159,7 @@
     },
     "settingsUserAgentToggle": "Utiliser un user-agent particulier",
     "settingsUpdateNotificationsToggle": "Vérifier automatiquement pour des mises à jour",
-    "settingsUsageStatisticsToggle": "Envoyer des statistiques d'utilisation",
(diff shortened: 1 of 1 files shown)


PalmerAL:
Thanks for translating!

For the usage statistics, you need to copy the entire HTML string from the en-US file:
```
"settingsUsageStatisticsToggle": {
      "unsafeHTML": "Send usage statistics (<a href=\"https://github.com/minbrowser/min/blob/master/docs/statistics.md\">More info</a>)"
    },
```

And replace "send usage statistics" and "more info" with translations.

You're right about custom bangs, I'll fix that.

Translating the keyboard shortcut names isn't supported at the moment; see issue #1322.

PEPERSO:
Alright, sorry for the missing URL, it's done now let me know when you add the custom bangs, I'll translate it too :)

PalmerAL:
You need the `unsafeHTML` part too; it's probably easiest to just copy the whole code block from the last comment, exactly as it's written.

I'll update this branch so that it has places to add the custom bang strings.

PEPERSO:
Oops, sorry, I'll remember for next translations !

PalmerAL:
I updated the branch, they should be there already!

PR #1461, opened 2021-01-25
Update pt-PT.json

smarquespt:
Those strings were already there.


PalmerAL:
Sorry about that, thanks for fixing it.

PR #1462, opened 2021-01-26
Update uk.json & ru.json

Sestowner:
Removed duplicate translation strings


PalmerAL:
Oh, sorry; I used find-and-replace to add these everywhere and didn't notice you already translated them here.

PR #1471, opened 2021-01-27
Removes feature requests from issues in an attempt to move them to discussions.

Milo123459:
Related: [this thread](https://github.com/minbrowser/min/discussions)


PalmerAL:
I think this is a reasonably good idea. I'm going to replace the feature request template with a note to use discussions instead, in case people don't read the readme.

PR #1473, opened 2021-01-27
Add keyboard shortcuts

Sestowner:


PalmerAL:
Cmd+h is used for "hide window" on macOS, so we can't use it here. Maybe shift+cmd+h instead?

Also, it might be nice to support customizing these. There's an example of how to do that here: https://github.com/minbrowser/min/commit/6c61b94d3a3170031f7c863a7fc62778816d74b5

PalmerAL:
LGTM; thank you!

PalmerAL:
Ctrl+b conflicts with the shortcut for "bold text" in Google Docs, and we don't have a way to let web pages override shortcuts, so I'm going to change it to shift+ctrl+b for the next release.

PR #1475, opened 2021-01-28
Custom bangs translations

PEPERSO:


PalmerAL:
Looks good, thanks!

I agree documentation would be good.

> how to search something on a specific website

What do you mean by this part?

PEPERSO:
I mean when I type `!osm Paris`, it leads me to `https://www.openstreetmap.org/search?query=Paris#map=12/48.8589/2.3469`, which is quite the same page as the result of searching "Paris" on the OSM page. 
But in the custom bang setting, I don't know which URL associate to to my phrase, so that `!custombang test` leads to the custom bang page, with `test` in th search bar, if any.

Is that clearer ?

PalmerAL:
Yeah, we should document that. To answer your question, you would put `osm` in the "phrase" field, and "https://www.openstreetmap.org/search?query=%s" in the redirect field.

PEPERSO:
Thanks, I'll remember if I need a new bang

PR #1484, opened 2021-02-02
improve restart

Milo123459:
This basically persists the restart banner until you actually restart.

### DIFF
--- a/pages/settings/settings.js
+++ b/pages/settings/settings.js
@@ -11,8 +11,12 @@ var userAgentCheckbox = document.getElementById('checkbox-user-agent')
 var userAgentInput = document.getElementById('input-user-agent')
 
 function showRestartRequiredBanner () {
-  banner.hidden = false
+  banner.hidden = false;
+  settings.set('restartNow', true)
 }
+settings.get('restartNow', (value) => {
+    if(value && value == true) showRestartRequiredBanner();
+})
 
 /* content blocking settings */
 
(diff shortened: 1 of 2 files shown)


pages/settings/settings.js
```diff
@@ -11,8 +11,12 @@ var userAgentCheckbox = document.getElementById('checkbox-user-agent')
 var userAgentInput = document.getElementById('input-user-agent')
 
 function showRestartRequiredBanner () {
-  banner.hidden = false
+  banner.hidden = false;
+  settings.set('restartNow', true)
 }
+settings.get('restartNow', (value) => {
+    if(value && value == true) showRestartRequiredBanner();
```
PalmerAL, line 18: Three style issues here:

* Space after "if" and === rather than == (standard should catch this)
* Use braces around the if statement (If you haven't read it, here's an [interesting and good example](https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/) of why you should do this)

pages/settings/settings.js
```diff
@@ -11,8 +11,12 @@ var userAgentCheckbox = document.getElementById('checkbox-user-agent')
 var userAgentInput = document.getElementById('input-user-agent')
 
 function showRestartRequiredBanner () {
-  banner.hidden = false
+  banner.hidden = false;
```
PalmerAL, line 14: There shouldn't be a semicolon here right?

Milo123459: I mean I just did that because I like semi-colons

PR #1513, opened 2021-02-19
Added a setting to switch to single tab mode.

Ordissimo:
We have been using your application for several years, it serves as a basis for desktop web-apps.
I propose you a single tab mode configurable in preferences.

### DIFF
--- a/js/browserUI.js
+++ b/js/browserUI.js
@@ -169,15 +169,21 @@ webviews.bindEvent('new-window', function (tabId, url, frameName, disposition) {
     focusMode.warn()
     return
   }
-  var newTab = tabs.add({
-    url: url,
-    private: tabs.get(tabId).private // inherit private status from the current tab
-  })
+  let useSingleTab = settings.get('useSingleTab')
+  if (useSingleTab === false || tabs.count() == 0) {
+    var newTab = tabs.add({
+      url: url,
+      private: tabs.get(tabId).private // inherit private status from the current tab
+    })
 
-  addTab(newTab, {
-    enterEditMode: false,
-    openInBackground: disposition === 'background-tab' && !settings.get('openTabsInForeground')
-  })
+    addTab(newTab, {
+      enterEditMode: false,
+      openInBackground: disposition === 'background-tab' && !settings.get('openTabsInForeground')
+    })
+  } else {
+    webviews.update(tabs.getSelected(), url)
+    tabEditor.hide()
+  }
 })
 
 webviews.bindIPC('close-window', function (tabId, args) {
--- a/js/menuRenderer.js
+++ b/js/menuRenderer.js
@@ -1,5 +1,6 @@
 /* Handles messages that get sent from the menu bar in the main process */
 
+var settings = require('util/settings/settings.js')
 var webviews = require('webviews.js')
 var webviewGestures = require('webviewGestures.js')
 var browserUI = require('browserUI.js')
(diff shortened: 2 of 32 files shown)


shalva97:
can you explain usecase of this and show what does it look like?

Ordissimo:
I have a small program that takes an url as an argument and generates a desktop file, an icon.
In the past I used chrome which allowed me to do this work directly.
Instead of Chrome, I use Min, I have only one window, I redirect everything in the active tab, adding and removing tabs are removed.
This mechanism, on Ordissimo allows to launch about twenty applications. It is possible to add them manually.
Capture d’écran du 2021-02-19 12-05-45

On the OS Ordissimo with its theme :
capture-2021-02-19-11h53m55s

Ordissimo:
For the translations I used Google Translate :)

Ordissimo:
Preferences window :
Capture d’écran du 2021-02-19 13-15-14

PalmerAL:
This is a cool use case, but it seems like it's somewhat outside the scope of what we're trying to do. A lot of features are designed with the assumption that multiple tabs are available, so maintaining this is going to be a lot of work. At a point in the past, we actually had a single-tab focus mode (although not exactly the same as this), but it was buggy for those reasons, and we ended up getting rid of it because it appeared that almost no one used it.

It seems like two better solutions here for you might be:
 - Keep maintaining a fork of Min, but use the multi-tab focus mode code as the basis and modify that so this is easier to support.
 - Write another electron app for this. It sounds like the features you want are fairly simple (no UI except a basic navbar, no navigating to other sites, etc.), and you're not using most of Min's features anyway. You could still import adblocking and other features from Min, so this may not be too much work.

(If you do end up deciding to write a custom app for this, I'm open to working on that - email is in my profile).

Ordissimo:
I use your application since version 1.7.1.
I tried different solutions, your application is a solid base, for the adaptation to the needs of Ordissimo I finally have few changes to make.
25 files are translations.
1 file is index.html, I add the checbox.
6 files with minor changes.
Your application is installed on more than 70000 Computers, I can perfectly maintain this change, but maintaining a fork (the code is already in github), while your project suits us perfectly makes no sense.

PalmerAL:
There are a number of issues with this as-implemented:
* The task overlay is still enabled (and the button renders incorrectly without the add-tab button next to it)
* It's still possible to create new tabs in a bunch of places: from the context menu, by pressing ctrl+t, etc
* If you do open multiple tabs, it becomes impossible to close them
* If you enable this by accident, there's no obvious way to escape out of it (unlike focus mode, which resets on restart, and shows dialogs explaining how to disable it).

All of these things are fixable, but doing so would add a lot of complexity, which goes back to my original point: to actually make this work in a high-quality way and avoid these issues, almost every UI feature in the codebase is going to need a check for "are we in single tab mode?" and handle that specially. Which is a lot of work, and not something I want to maintain.

Furthermore, the main purpose of Min is to add a tab-management UI and some browser features on top of Chromium. So if you don't want tab management, Min really isn't the right solution; you're going to be better off using your own build of Electron and adding the features that you need.

For those reasons, I'm going to close this, but I do understand that you'd like to keep using the existing solution. Looking at this PR, it looks like it should be fairly easy to maintain in a fork, and this code doesn't change very often. So I would expect it to remain possible to maintain this separately.

Ordissimo:
hi @PalmerAL, 
I will maintain a branch of your project in my Git, taking into account your feedback.
Thank's

PR #1518, opened 2021-02-24
[WIP] Add page highlighting feature

PalmerAL:
Discussion post: https://github.com/minbrowser/min/discussions/1496

This PR adds an option to highlight snippets of text on a page. Highlights are created from the context menu, and once they're created, they are saved even after the page is reloaded.

Screen Shot 2021-02-24 at 4 30 37 PM

The idea behind this is that if you're bookmarking a longer article, it can be hard to remember why you saved it. With this, you can highlight specific parts of the page so that you can find them more easily again / remember why you saved it in the first place.

(You don't have to bookmark a page in order to highlight it - highlights get kept for as long as the page remains in your history / bookmarks).

I wrote this fairly quickly as a proof of concept, so there's a lot of things left to fix before this can be merged. However, if anyone wants to try this out, I'd be really interested in hearing feedback! - whether this is useful, what you'd use this for, and if you have any other ideas for it. Site compatibility issues would also be good to catch at this point (eg. if creating a highlight breaks the formatting of a particular page).


blackgwe:
Nice feature. Formerly I used [hypothes.is](https://web.hypothes.is/help/annotation-basics/) to tag / highlight some pages. The main idea differs a lot, but imho it's worth to take a closer look at this concept which is in short:  "share highlights, tags and notes of web-sites (and pdf)" between internet users (groups) as  as one aspect of a collaboration.

Maybe you could extends this issue #1518:  
- add tags (for better finding the highlights) & extend the bookmark search to browse/find them
- maybe add private notes (IndexedDB)
- maybe some integration with hypothes.is

PalmerAL:
Thanks for the feedback! I haven't heard about hypothesis before, but I'll look into that.

We have support for adding tags to bookmarks already; it sounds like you want to be able to add a separate set of tags for each individual highlight? Could you give some examples of how you would use that?

blackgwe:
It's difficult to explain, maybe by examples
* overview on latest highlights / tags / sites, for example [security](https://hypothes.is/search?q=tag%3Asecurity) tag (see screenshot)
* [Here](https://via.hypothes.is/https://www.inboundlogistics.com/cms/article/shipment-monitoring-technology-picking-up-the-signals/#annotations:1dl7inseEeu2fheC53Za8A) you can see tags on individual highlights / annotations

I do not know, if or how to integrate hypothes.is in min browser, but I think it's worth to think about it. Maybe most can be done using scripting support of the bowser???

https://web.hypothes.is/about/
> Our efforts are based on the annotation standards for digital documents developed by the W3C Web Annotation Working Group. We are partnering broadly with developers, publishers, academic institutions, researchers, and individuals to develop a platform for the next generation of read-write web applications. 

In fact this highlighting feature has very little in common with the [W3C Web Annotation](https://w3c.github.io/web-annotation/), but with a little imagination/fantasy, you can certainly recognize some junctions/interfaces/crossing points.

hypothes_is_for_security_tag
four_annotations_in_one_page

Issue #1520, opened 2021-02-26
Concurrent modification of "settings" object corrupts "settings.json"


yulucodebase:
- **Min Version:** 1.18.1
- **Operating system:** Raspberry Pi OS and should happen on all Linux

## Expected Behavior
Settings are correctly saved and the app can be restarted.

## Actual Behavior
**`settings.json`** (in user's data directory)** is **corrupted when the change requires restart**. The app crashes during restart due to the corrupted settings. The corrupted json file ends with an extra `}`. For example:

```JSON
{"restartNow":true,"clientID":"5525817712610477","installTime":1614348000000,"useSeparateTitlebar":true}}
```

Currently, I can only reproduce this when changing **"use separate title bar"** but in theory it should happen for any setting that requires restart (so "use a custom agent" should also has the issue but I haven't been able to reproduce it yet).

The possible cause of the issue is likely related to the following calls in **`pages/settings/settings.js`**:
```Javascript
separateTitlebarCheckbox.addEventListener('change', function (e) {
  settings.set('useSeparateTitlebar', this.checked)
  showRestartRequiredBanner()
})
```
Notices above, `settings.set()` will update the "settings" object and call the async file write (saving "settings.json"). 

It's possible that by the time`showRestartRequiredBanner()` is executed, the file operation hasn't completed yet. The trouble with this is that `showRestartRequiredBanner()` also modifies the `settings` object (it updates "restartNow" property) and write to the "settings.json" file. 

**Basically, it's a race condition between two file write operations (to the same file) with slightly different contents.**

_BTW, I found this issue while trying to enable **"use separate title bar"** for Mac OS, which is a work in progress (will raise separate pull request later)._

## To Reproduce
_As described above, this is a race condition issue. In order to reproduce it you might need to repeat the following steps multiple time since it's related to the timing of two concurrent operations._

1. Launch your app on Linux or Raspberry Pi.
2. Go to "settings" and just change "use separate title bar" setting.
3. Once the setting is changed and the "You need to restart to apply these changes" warning is displayed, check "settings.json ( this file is under user's data directory, for example, on my Raspberry Pi it's under `/home/{your user home}/.config/Min-development`).
4. Once confirmed the file content is corrupted, try to restart the app. It should crash.

yulucodebase:
I can think of two ways to fix this:

[...]

PR #1521, opened 2021-02-28
Enable "use separate title bar" setting for all platforms (not just Linux)

yulucodebase:
### Purpose of this commit
As a Mac user, the only way to "maximize" a window is to double click on the title bar of the application. **However, it's not possible to do it with Min browser since title bar is hidden on Mac.** 

Although @PalmerAL [kindly point out](https://discord.com/channels/764269005195968512/764544014259060797/814756935316013066) that I can hold "cmd" + clicking on the "green traffic light", it's not as intuitive as double clicking title bar, as most users should be familiar with. 

Also, if we are giving Linux users that have special desktops the option to show title bar, why wouldn't we make it consistent for all platforms and give users flexibility of having a separate title bar? 

I therefore propose to enable "Use separate title bar" option for all platforms.

This commit also adjust the size/position of the "navbar" on both Windows and Mac OS when title bar is shown. Please find below the result of this change when the option is enabled:

**Mac OS**
Screenshot 2021-02-28 at 21 40 55
Screenshot 2021-02-28 at 21 40 46
 
**Windows 10**
minbrowser_win_settings_titlebar
minbrowser_win_titlebar

### DIFF
--- a/css/windowControls.css
+++ b/css/windowControls.css
@@ -9,13 +9,13 @@ body {
   display: none;
 }
 
-body.windows:not(.maximized):not(.fullscreen),
+body.windows:not(.maximized):not(.fullscreen):not(.separate-titlebar),
 body.linux:not(.maximized):not(.fullscreen):not(.separate-titlebar) {
   --control-space-top: 12px;
 }
 
 /* On Windows, draggable regions aren't clickable, so the drag region is a separate area above the tabstrip */
-body.windows:not(.maximized):not(.fullscreen) .window-drag-area,
+body.windows:not(.maximized):not(.fullscreen):not(.separate-titlebar) .window-drag-area,
 body.linux:not(.maximized):not(.fullscreen):not(.separate-titlebar)
   .window-drag-area {
   display: block;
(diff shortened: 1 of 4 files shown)


yulucodebase:
Please only merge this, if it is accepted, after #1520 is fixed. The bug in persisting settings could be triggered by changing "Use separate title bar" option.

css/windowControls.css
```diff
@@ -41,21 +41,21 @@ body.mac .window-drag-area {
```
PalmerAL, line 41: `window-drag-area` should probably be hidden on mac when the separate titlebar is enabled.

css/windowControls.css
```diff
@@ -9,13 +9,13 @@ body {
...
+body.windows:not(.maximized):not(.fullscreen):not(.separate-titlebar),
 body.linux:not(.maximized):not(.fullscreen):not(.separate-titlebar) {
   --control-space-top: 12px;
 }
 
 /* On Windows, draggable regions aren't clickable, so the drag region is a separate area above the tabstrip */
-body.windows:not(.maximized):not(.fullscreen) .window-drag-area,
+body.windows:not(.maximized):not(.fullscreen):not(.separate-titlebar) .window-drag-area,
 body.linux:not(.maximized):not(.fullscreen):not(.separate-titlebar)
   .window-drag-area {
```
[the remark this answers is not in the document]
ylu02: done

css/windowControls.css
ylu02: looks like it has no impact but I've just changed it as that probably makes more sense.

PalmerAL:
LGTM, thank you!

PR #1539, opened 2021-03-15
Update es.json

ZePotente:
Added some translations and code for the ones that were in the english file, but not in this one.


PalmerAL:
I think this looks OK. There's still some strings that aren't translated (search for "missing translation") - do you want to add those as well?

ZePotente:
There are a few that I didn't translate, which are:
Line 155: Where can I see the separate title bar? I don't really know what it is.
Line 156: What does opening new tabs in the foreground do and how do I make it happen? I see it's an option in the preferences but can't tell what it does.
Lines 262 and 264: I don't really understand what unlocking the password store means, is it in the sense of being then able to store the passwords?
Lines 274 and 280: How can I make these dialogs (login and secret key) come up? Just so I know the context they're being used in.

PalmerAL:
>  Line 155: Where can I see the separate title bar? I don't really know what it is.

It's a setting option that's only visible on Linux; here's a screenshot of what it does: https://github.com/minbrowser/min/pull/977#discussion_r414918201

>  Line 156: What does opening new tabs in the foreground do and how do I make it happen? I see it's an option in the preferences but can't tell what it does.

It changes what happens when you right-click on a link and choose "open in new tab". If it's enabled, the browser switches to the new tab; if it's disabled, the current tab stays active.

>  Lines 262 and 264: I don't really understand what unlocking the password store means, is it in the sense of being then able to store the passwords?

This is for when you are connecting Min to an external password manager (in settings > password auto fill). You enter your password manager credentials, then Min sends them to the external password manager, and it decrypts/unlocks your saved passwords. Once it unlocks, you can autofill passwords that are read from the password manager.

>  Lines 274 and 280: How can I make these dialogs (login and secret key) come up? Just so I know the context they're being used in.

274 (loginPromptTitle) is for a login dialog that's displayed when a site requests a password: https://github.com/minbrowser/min/pull/977#discussion_r414918890

280 (secretKey) is part of the password manager setup process - for some password managers, you need to enter a "secret key" as part of the credentials to unlock the password store.

localization/languages/es.json
```diff
@@ -54,7 +54,7 @@
     "searchHistory": "Historial de búsqueda",  //noun?
     "importBookmarks": "Importar marcadores",
     "exportBookmarks": "Exportar marcadores",
-    "runUserscript": "Correr userscript", //userscript doesn't have a Spanish translation, afaik
+    "runUserscript": "Correr userscript", //userscript doesn't have a Spanish translation, afaik maybe "script de usuario"
```
PalmerAL, line 57: Whatever translation you think is best here is fine with me.

ZePotente: i think I'd leave it as userscript because if someone wants to know what it is it's easier to find it that way.

PalmerAL: OK, sounds good.

PalmerAL:
I reset the English names to `null`, since the code will just use the English string if there's no translation available.

I think everything else looks good to me - is there anything you wanted to change?

ZePotente:
No, I think that's it at least for now.

PalmerAL:
OK, thank you!

PR #1560, opened 2021-04-11
Autofill improvements

PalmerAL:
See discussion for context: https://github.com/minbrowser/min/commit/80d69f186c04c49dda8b84dae3f7cecc58ce2a82#commitcomment-49226470

Main change here is to try to identify a single "best" username field on a page to autofill, rather than autofilling every field that could be a username. This improves the situation on some pages, but also increases the risk of broken autofill (if the field we identify as "best" isn't the right one).

### DIFF
--- a/js/passwordManager/passwordCapture.js
+++ b/js/passwordManager/passwordCapture.js
@@ -50,13 +50,8 @@ const passwordCapture = {
       domain = domain.slice(4)
     }
 
-    var username = args[0][1][0] || ''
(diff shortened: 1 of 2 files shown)


PalmerAL:
Going to merge this for now, but I'd really appreciate knowing about any sites that this breaks - if anyone sees something wrong, please make an issue.

blackgwe:
@PalmerAL Thanks a lot! I really like it! 👍

PR #1562, opened 2021-04-12
Fix getting menu keybinding when multiple bindings are set

### LINKED ISSUE #1558: App crashes on launch and preferences are erased after update to 1.19.0
- **Min Version:** 1.19.0

## Expected Behavior
Update and successful run :)

## Actual Behavior
Started to receive error alerts after installing the update. Also all my settings were gone including "ads blocked" counter, but not history and logins. At first 2 launches it created 5 processes of Min, 42% RAM were used, so those are not just sub-processes, I assume. Then the app crashed.

Screenshot (165)

Then it launched without an error. I customized the browser again, and set shortcuts with Cyrillic letters as well (!), and it doesn't crash, but gives me alerts on launch again, although they are not as persistent (only once per launch).
(And just crashed again after opening settings to take a screenshot...)

## To Reproduce
I don't know why it crashes for me on launch. I guess the real question is what are those lines in your code for. Hope you will be able to fix it with ease. Have a good day :)

PalmerAL:
Fixes #1558 

CC @Milo123459 Hopefully this makes it clearer what the problem is. It's also easier to see if you look at the settings file [here](https://github.com/minbrowser/min/issues/1558#issuecomment-817263440), specifically this:

```
"addTab":["mod+t","mod+е"],"addPrivateTab":["shift+mod+p","shift+mod+з"]
```

### DIFF
--- a/main/menu.js
+++ b/main/menu.js
@@ -1,8 +1,23 @@
 function buildAppMenu (options = {}) {
+  function getFormattedKeyMapEntry (keybinding) {
+    const value = settings.get('keyMap')?.[keybinding]
+
+    if (value) {
+      if (Array.isArray(value)) {
+        // value is array if multiple entries are set
+        return value[0].replace('mod', 'CmdOrCtrl')
+      } else {
+        return value.replace('mod', 'CmdOrCtrl')
+      }
+    }
+
+    return null
+  }
+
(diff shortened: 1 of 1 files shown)


Milo123459 (approved):
LGTM

PR #1566, opened 2021-04-15
Update main.js (Object destructuring)

miko-github:
# Destructruing Electron Modules

Hi, :wave:
I did only **destructuring** object from electron module :heart: 

i commented prev code, then, if you want to remove, remove comment and use :bulb: 

[**mozila docs**](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#object_destructuring)

## thank you for this awesome browser :tada:

### DIFF
--- a/main/main.js
+++ b/main/main.js
@@ -1,15 +1,32 @@
 const electron = require('electron')
 const fs = require('fs')
 const path = require('path')
-const app = electron.app // Module to control application life.
-const protocol = electron.protocol // Module to control protocol handling
-const BrowserWindow = electron.BrowserWindow // Module to create native browser window.
-const webContents = electron.webContents
-const session = electron.session
-const ipc = electron.ipcMain
-const Menu = electron.Menu
-const MenuItem = electron.MenuItem
-const crashReporter = electron.crashReporter
+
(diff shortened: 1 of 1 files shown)


main/main.js
```diff
@@ -1,15 +1,32 @@
 const electron = require('electron')
```
PalmerAL, line 1: This should also be part of the object destructuring

PalmerAL:
I think this is fine, can you remove the old code?

main/main.js
miko-github: # Hi PalmerAL :wave: 
exactly  yes, but i think, **may** that need to use **electron** module :wink: 

## **thanks for your review :heart:**

miko-github:
> I think this is fine, can you remove the old code?

so, we need to accept of owner of this repo, if he accepted, he can remove or he can save :hugs:

main/main.js
PalmerAL: Oh, nevermind, I misread what was happening here.

PalmerAL:
Thanks.

PR #1587, opened 2021-04-25
Update main.js

miko-github:
Update importing (require from electron, fs , path) :star:  :heart:  :fist_raised:

### DIFF
--- a/main/main.js
+++ b/main/main.js
@@ -1,6 +1,8 @@
-const electron = require('electron')
-const fs = require('fs')
-const path = require('path')
+const [electron, fs, path] = [
+  require('electron'),
+  require('fs'),
+  require('path'),
+];
 
 const {
   app, // Module to control application life.


PalmerAL:
I don't think I agree with this change - it's not any shorter, and it's harder to read. With the old way, you can read the code, and see that `path` corresponds to `require('path')` for example. Now, you have to count how many items there are in the list to see which name matches each require() call.

miko-github:
i first import main modules, then second module
for express i fist importing `express` and then `cors` `jwt`, and next `fs`,`path` and other things
this is my sort for any thing like this but actually i know and Accept :+1: :hugs:

PalmerAL:
I still think it's easier to just have the module name and the `require` call on the same line, so I'm going to leave this as-is.

miko-github:
# Hi :wave:,
if we add @babel then, that create any bug or problem?
i want to use @babel/cli and etc. to this project and change to `require` to `import`

that is impossible? :thinking:
:fist_oncoming:

PalmerAL:
I'm not opposed to replacing Browserify with a different build system, but I think I'd want to analyze that more carefully before making a decision; it would be good to look at what alternatives there are to Babel first.

PR #1590, opened 2021-04-25
Update keybindings.js

miko-github:


PalmerAL:
https://github.com/minbrowser/min/pull/1587#issuecomment-830716587

PR #1593, opened 2021-04-25
Create is.js

miko-github:
# `IS`

> I saw many data type checking for usability  (not undefined)
> in this repos used but they are be only function

you can only import like this
```javascript
const { notEmpty } = require('path/is.js');
// or
const is = require('path/is/js');
```
for empty you can rename to `empty` then for using in target file you may need `if(!is.empty(...));`\
for using to check if module are available you use
```javascript
// ... other codes

if(is.notEmpty(module)) module.exports = anything_you_want;
else anythings_else;

// instead

if(typeof module !== undefined) {
    module.exports = any_things;
} else other_things;
````
if you want any thing else like these,
only add here that function :hugs:
and if i find anything else in repo files, i will add here that function :+1: 
and if you want to replace `undefined` checking with this function, update or let me know and i update them :hugs:  :fist_raised:

best wish for your, your friends @miko-github  :heart:

### DIFF
--- a/js/util/is.js
+++ b/js/util/is.js
@@ -1,0 +1,13 @@
+/**
+  * If i need to check data types or any thing like that
+  * first, add here that funciton, then import in target file,
+  * next, use function
+  * WHY : I saw many data type checking for usable (not undefined) in this browser used but they are be only function
+  */
+
+const type = (value) => typeof value;
+const is = {
+  notEmpty: (value) => type(value) !== undefined, // or rename to `empty` => in usage => `if(!is.empty(...));` for module `if(is.notEmpty(module))`
+  // ... and etc. do you think you want
+  // but i read the codes, and if find any thing else of undefined, i add them here, and you just need to importing or tell me :huge:
+};

### CODE REVIEW

**Comment by miko-github** [comment]
if you think this will be work done, use this for module availability checking
```javascript
const is = { 
...
    module: (m, exp, other) => is.notEmpty(m) ? (module.exports = exp) : other, // OR => (m.exports = exp)
...
```
usage is
```js
// parameter module is current file module i think (may be :thinking:) 
is.module(module, main, runMain)

// equal

if(is.notEmpty(module)) module.exports = main;
else runMain(); // initialize, but i should update `is.module` for last arg as callable :smiley: 
```
:heart:

PalmerAL:
I don't really see the point of this. If you're comparing these two things:

```
if (typeof module !== 'undefined') {
  module.exports = { db }
}
```
```
if (is.notEmpty(module)) {
  module.exports = { db }
}
```

They seem equal in complexity. And the second one requires more work to understand, because then you have to remember what `notEmpty` does. With the first example, you can directly see what it's doing.

PR #1601, opened 2021-05-09
Add reload ignoring cache shortcut

Sestowner:


PalmerAL:
WDYT about making the normal "reload" shortcut ignore the cache? I think it was like that at some point, and I don't remember it causing any issues.

Sestowner:
> WDYT about making the normal "reload" shortcut ignore the cache?

I'd like to keep both options, because for example, on a slow internet connection to load the missing image, you'd need to reload all over again.

PalmerAL:
OK, sounds good.

js/defaultKeybindings.js
```diff
@@ -256,22 +256,17 @@ const defaultKeybindings = {
...
       webviewGestures.zoomWebviewIn(tabs.getSelected())
     })
 
     keybindings.defineShortcut('showHistory', function () {
       tabEditor.show(tabs.getSelected(), '!history ')
     })
-
-    // reload the webview when the F5 key is pressed
-    document.body.addEventListener('keydown', function (e) {
-      if (e.keyCode === 116) {
```
PalmerAL, line 269: This was previously just the standard `f5` key; I think what we want to do is add `f5` to the `reload` shortcuts list in defaultKeyMap to match the previous behavior, and then keep `mod+f5` as `reloadIgnoringCache`.

Sestowner: Done.

PalmerAL:
LGTM!

Issue #1281, opened 2020-10-16
whole browser crashes if i press CTRL + P


shalva97:
* **Min Version:** 1.16.0
* **Operating system:** Arch

## Expected Behavior
when i press CTRL + P, doesnot matter what it does, browser shouldnot crash

## Actual Behavior
pressing CTRL + P crashes whole browser with:
```
shalva@archlinux ~ [SIGTRAP]> electron playground/min/ --disable-smooth-scrolling
fish: “electron playground/min/ --disa…” terminated by signal SIGTRAP (Trace or breakpoint trap)
```

## To Reproduce
CTRL + P

shalva97:
wrong electron version here too, cannot reproduce

shalva97:
okay, this is not a electron version issue.

I had bound CTRL + P to "run shortcut", but it is also triggering print dialog and so it then crashes when i press "cancel"...

PalmerAL:
Are you sure it's not the Electron version? I can't reproduce with Electron 9, but I can reproduce with Electron 11. I've filed an issue about that here: https://github.com/electron/electron/issues/26016

shalva97:
yeah.... im using this [package](https://www.archlinux.org/packages/community/any/min/) and they have it written electron9 here https://github.com/archlinux/svntogit-community/blob/220175238b72a476d59e699ac24d47f89d926256/trunk/min.js#L1

Milo123459:
Update to electron 11 beta, electron has fixed this issue.

shalva97:
which beta do you mean? this still happens on electron 1.11.1 and is beta version supported by Min?

Sestowner:
I can reproduce this on Arch Linux & Debian 10.

To reproduce:
1) Ctrl+L to activate the address bar
2) Ctrl+P
3) Click "Cancel" button on the print dialog.

Sestowner:
I think this might be related to Electron's issues:
https://github.com/electron/electron/issues/27954
https://github.com/electron/electron/issues/28848

PR #1604, opened 2021-05-12
Workaround for #1281

Sestowner:


PalmerAL:
Can you add a comment explaining this?

Sestowner:
This prevents opening the print dialog when the address bar is focused or password manager setup dialog opened etc. Because in those cases pressing the "Cancel" button on the dialog causes the browser to crash.

PalmerAL:
Oh I meant adding a comment to the code. Will fix.

PR #1606, opened 2021-05-14
Add search suggestions URLs

Sestowner:


PalmerAL:
This looks pretty good!

I do have a bit of a concern about the Google URL; it looks like they require `client=chrome` for this to work, and they've previously made attempts to block non-approved clients from using the API: https://developers.google.com/search/blog/2015/07/update-on-autocomplete-api. I wasn't able to find any written explanation of what usage is allowed, but I don't think I want to deal with potentially being blocked from using it, so I'd suggest removing that one for now.

All the other search engines have autocomplete URLs listed in their opensearch.xml and don't require any authentication, so they should all be fine to use.

PalmerAL:
Looks good!

rotemad:
Is there any option to manually enable/add back the google suggestionsURL?

PalmerAL:
You could edit searchEngine.js to include the file and re-compile (https://github.com/minbrowser/min#building-binaries); otherwise no.

rotemad:
@PalmerAL got it. I thought that might be an easier way.
Thanks :)

Issue #140, opened 2016-05-10
Improve window.open support


Smona:
When trying to download a free song from toneden.io, the page which redirects from a soundcloud verification gets stuck, making it impossible to download. 

[Example of where this happens](https://www.toneden.io/infuze/post/flosstradamus-roll-up-baauer-remix-infuze-re-roll)

PalmerAL:
I'm guessing this is probably because we don't support window.open very well.

miere:
The same happens on many sites when I try to use Facebook or Twitter login integrations...

PalmerAL:
from #446:

> Site netlify.com (and many others) do authentication after logging in in the new window and refreshing the main window.
> 
> Problem:
> window.opener is nuil.
> 
> See
> 
> electron/electron#9533
> electron/electron#9961
> jiahaog/nativefier#164

igk1972:
May by better migrate to https://github.com/brave/muon as framework ?

Muon is a fork of the Electron framework.
Some of Muons goals include:
- add security focused features for the renderer:
   - scriptable **window.opener support**.
   - **full sandbox**.
   - remove node completely (from the renderer process).
- faster and more streamlined end-to-end build process.
- use the Chromium source directly (eliminating electron's copy of chrome_src) with minor patches
make integrating chrome components less painful.

PalmerAL:
@igk1972 see #440:

> Muon does seem to solve a lot of the issues with this (and support for popup windows would be nice as well). The biggest problem with Muon (and what's stopped me from considering a switch in the past) is that there seems to be almost no documentation on it. Brave's added a bunch of API's specific to their browser, and changed a bunch of things about how Electron's existing features work, and the only way (unless I'm missing something?) to figure out how to use it is to look at Brave's source code and guess what the correct API is. Meanwhile, glancing through their issue tracker, they don't seem to answer questions about it (which I completely understand, they're not obligated to, but it's a problem for us). If we migrate to Muon, I think we're going to have a lot of problems with this; I'm not really sure what the best option is here.

PalmerAL:
But I do agree that Muon has a lot of advantages, maybe it's worth trying to figure out how to use it based on the Brave source code.

dbfreem:
Is this still something being considered? I ran into a situation today where popups not being supported was a blocker for me using min.

PalmerAL:
[...]

PR #1608, opened 2021-05-14
[WIP] Add experimental popup window support

PalmerAL:
This adds support for opening popup windows from webpages, using [this](https://github.com/electron/electron/pull/26802) new feature of Electron. Implementing this would fix most of the compatibility issues mentioned in #140

Known issues:
* ctrl+clicking on links to open them in a new tab no longer works. To fix that, we need https://github.com/electron/electron/pull/28518, which requires upgrading to Electron 14 (or maybe 13?)
* Occasionally the browser crashes when closing the popup window.
  * It would be really helpful to find a consistent way to reproduce this - if anyone finds one while testing, please let me know.

### DIFF
--- a/js/browserUI.js
+++ b/js/browserUI.js
@@ -186,7 +186,9 @@ webviews.bindEvent('new-window', function (tabId, url, frameName, disposition) {
 */
 
 webviews.bindEvent('did-create-popup', function (tabId, popupId) {
-  var popupTab = tabs.add()
+  var popupTab = tabs.add({
+    private: tabs.get(tabId).private
+  })
   tabBar.addTab(popupTab)
   webviews.add(popupTab, popupId)
   switchToTab(popupTab)
--- a/js/browserUI.js
+++ b/js/browserUI.js
@@ -167,8 +167,8 @@ function switchToTab (id, options) {
   })
 }
 
+/*
 webviews.bindEvent('new-window', function (tabId, url, frameName, disposition) {
-  /* disabled in focus mode */
   if (focusMode.enabled()) {
     focusMode.warn()
     return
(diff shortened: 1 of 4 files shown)


PalmerAL:
CC @blackgwe - this fixes the issue with the hypothes.is login page.

blackgwe:
@PalmerAL Thank you very much. It works for me. Unfortunately, I don't know any other page with pop-up windows where I can test the browser crash. If I can help, pls let me know!

I tested minbrowser with current electron 14.0.0-nightly.20210413 successfully, but the "ctrl+clicking on links" bug does not disappear. 

Another idea: The "ctrl+clicking on links" bug shows that it's possible to support multiple minbrowser windows  (of course it's confusing to have tasks, tabs and windows…) But maybe you can have a config value in the min://settings) to enable "windows support" which means, there is a menu item "File Menu" → "Open new window".
-> https://github.com/minbrowser/min/discussions/1582
-> https://github.com/minbrowser/min/discussions/1573

PalmerAL:
Great!

Once we upgrade, I think we need to add a `windowOpenHandler` (https://www.electronjs.org/docs/api/web-contents#contentssetwindowopenhandlerhandler) and then it should probably work.

Multiple windows would be nice, but there's a lot of features of Min that won't work without substantial work to modify them (I listed some of them here: https://github.com/minbrowser/min/issues/169#issuecomment-635713365). Fixing all of those problems hasn't been a priority, and so I don't think it makes sense to have a hidden setting for something that doesn't really work.

PalmerAL:
ctrl+clicking is fixed now, and I'm not seeing any other crashes (or any other issues really), so I'm hoping this is done.

I'll probably test this for a little bit, and assuming no issues, merge and release a beta version with this change.

Issue #952, opened 2020-04-02
Bug Menu


idAlencar:
Hi @PalmerAL, I noticed an error in the menu, when it is hidden (Ctrl + M) when exiting full screen mode, the horizontal menu and the menu points appear together

![menu](https://user-images.githubusercontent.com/58848807/78241689-311c5900-74b7-11ea-9f8a-f7660c463537.png)

PalmerAL:
This might be fixed once the window controls are inline with the tab bar: https://github.com/minbrowser/min/issues/909#issuecomment-592997788, since in this mode, the horizontal menu would be disabled completely.

PR #1612, opened 2021-05-18
Hide menu when page leaving fullscreen

Sestowner:


PalmerAL:
I think this is fine, although I'm surprised it'd be necessary, since we already do this in the `leave-full-screen` on line 262. Is #952 still reproducible in the latest version?

Sestowner:
> Is #952 still reproducible in the latest version?

Yes.

PalmerAL:
OK, thanks.

PR #1617, opened 2021-05-21
Some css improvements

Sestowner:
* add hover states for buttons, .download-item, .task-container
* fix margin of .navbar-right-actions
* fix padding of #add-tab-button


PalmerAL:
The download manager changes look pretty good. The buttons on the right side of the download could probably use some more padding though:

Screen Shot 2021-05-21 at 5 19 47 PM

I also wonder about making the close button outline circular to match the tab close buttons... not sure what I think about that one though.

The add tab and task buttons seem like they have too much spacing between them now:
Screen Shot 2021-05-21 at 5 23 41 PM

I think they look better as-is (fixing the underlying CSS issue is probably good though).

Having a square hover background that extends to the window edge works well on Windows, but it doesn't feel like it fits in well on macOS, because nothing else does that. For example, Firefox leaves a gap between the new tab button and the window:

Screen Shot 2021-05-21 at 5 26 02 PM

For comparison:

Screen Shot 2021-05-21 at 5 24 58 PM

It's possible something like that could fix the issue here, or it might be better to just leave it without a hover background. And I haven't looked at it on Linux.

PalmerAL:
Will review soon.

PalmerAL (commented):
* The download manager changes look good!
* The spacing on the tab bar looks better now.
* The find in page changes also seem fine.

A couple things I think should be changed:

* I put the forward button near the back button so that after hovering to make it show up, you wouldn't have to move the mouse very far to click on it. Since it only appears when you're hovering on the back button, it shouldn't be possible to click on it instead of the back button accidentally. So I think that should be reverted to the previous spacing.
* Adding a background to the add tab / add task buttons on hover still doesn't fit in very well on macOS, so I think we should remove it.
* Having a background behind a task when it's hovered feels like it adds necessary noise; I don't think it helps very much at selecting a task vs. just having the tab highlighted. So I think that should be removed as well.

Also, I changed the tab editor buttons to circular outlines in 6af6282d7e1f0a9721019cf90b1233fe757f507d so they would fit better with the tab close buttons, and also because they look better on macOS that way.

Sestowner:
> * Adding a background to the add tab / add task buttons on hover still doesn't fit in very well on macOS, so I think we should remove it.

WDYT about don't apply the hover state only on macOS?

PalmerAL:
Yeah I think that would be fine. Linux is going to have the same issue where it doesn't fit in with the rest of the desktop UI, but it could be enabled on Windows only.

PalmerAL:
I wanted to make sure this got into the next release (which I'm going to make in a couple days), so I modified the hover state to only show on Windows. With that, I think everything else here is good.

soffieswan041:
> * add hover states for buttons, .download-item, .task-container
> * fix margin of .navbar-right-actions
> * fix padding of #add-tab-button

PR #1622, opened 2021-05-24
Added button to clear all tasks.

GalFajon:
Added a button that clears all tasks alongisde the "New Task" button. I also adjusted the css and index.html accordingly. In addition to all that, I added a custom bang for destroying all tasks as well.

I'll be frank, I wasn't sure how to fit it in from a UI design standpoint, but it was a feature I always wanted to see.

minbutton

### DIFF
--- a/css/taskOverlay.css
+++ b/css/taskOverlay.css
@@ -166,7 +166,8 @@ body:not(.touch)
   opacity: 0.8;
 }
 #add-task {
-  width: 80%;
+  display:inline-block;
+  width: 40%;
   margin-left: 10%;
   padding: 1.25em;
   text-align: center;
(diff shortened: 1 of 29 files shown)


PalmerAL:
Thanks!

I don't think this is the right place for the button - it blends in with the "new task" button, and is located in an area where the "new task" button used to be. As a result, I imagine a lot of people accidentally clicking on this when they mean to create a task.

How about putting the button in the upper-left corner, inside `#task-overlay-navbar`? For the styling, you could do something simple like the "clear history" button:
Screen Shot 2021-06-06 at 6 29 18 PM

Also, I think it'd be a good idea to add a confirmation dialog for this - when I'm working on a project, for example, I usually have a lot of tasks open for several weeks with all the resources for the project, and accidentally deleting them all would be a big issue. You can just use [`confirm()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm) for this.

GalFajon:
Very fair response. I'll get on it.

flightmansam:
@GalFajon, any updates here?

PR #1635, opened 2021-06-04
Default to HTTPS

lubiedo:
Simple commit to avoid defaulting to an insecure protocol when non existent in URL.

### DIFF
--- a/js/util/urlParser.js
+++ b/js/util/urlParser.js
@@ -63,7 +63,7 @@ var urlParser = {
 
     // if the url doesn't have a space and has a ., or is a host from hosts file, assume it is a url without a protocol
     if (urlParser.isURLMissingProtocol(url)) {
-      return 'http://' + url
+      return 'https://' + url
     }
     // else, do a search
     return searchEngine.getCurrent().searchURL.replace('%s', encodeURIComponent(url))
--- a/js/webviews.js
+++ b/js/webviews.js
@@ -435,7 +435,11 @@ webviews.bindEvent('page-title-updated', function (tabId, title, explicitSet) {
 
 webviews.bindEvent('did-fail-load', function (tabId, errorCode, errorDesc, validatedURL, isMainFrame) {
   if (errorCode && errorCode !== -3 && isMainFrame && validatedURL) {
-    webviews.update(tabId, webviews.internalPages.error + '?ec=' + encodeURIComponent(errorCode) + '&url=' + encodeURIComponent(validatedURL))
+    if (validatedURL.indexOf("https://") === 0) {
+      webviews.update(tabId, validatedURL.replace(/^https/, 'http'))
+    } else {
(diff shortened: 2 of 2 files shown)


PalmerAL:
This would be nice to do; however, I think the portion of sites that don't support HTTPS is still too high - around 10% of page loads in Firefox: https://telemetry.mozilla.org/new-pipeline/dist.html#!cumulative=0&end_date=2021-06-04&include_spill=0&keys=__none__!__none__!__none__&max_channel_version=nightly%252F91&measure=HTTP_PAGELOAD_IS_SSL&min_channel_version=null&processType=*&product=Firefox&sanitize=1&sort_by_value=0&sort_keys=submissions&start_date=2021-05-31&table=0&trim=1&use_submission_date=0

If we try to do this and the site doesn't support https, the page load will fail.

The way Chrome is solving this is that they try to load the https version, wait a few seconds, and then fall back to http if it fails: https://9to5google.com/2021/01/11/google-chrome-address-bar-may-soon-default-to-https/. That's probably what we should do as well, but it'll be more complicated than this one line change (you're still welcome to do it if you want).

lubiedo:
Thanks for the feedback. Yes, I think you are right and that it's a good idea. I will work on this PR and see what I can do!

lubiedo:
Got some time to add a redirection to HTTP in case of HTTPS fails on 471600a. Let me know if useful or needs more work!

PalmerAL:
I think that solves the original issue I mentioned, but I'm wondering if it introduces a security issue. If you were on an HTTPS site and someone malicious was interfering with traffic from your computer, I'm thinking it might be possible for them to force a connection refused error. Then we would silently redirect to the HTTP version of the website, which would allow them to view all traffic from the website.

One possible other way of doing this (I'm not sure if it's the right way, let me know what you think) would be to have urlParser keep a list of URLs that were automatically upgraded to HTTPS. Then inside the connection refused event, we could check if the URL is on that list, and only redirect back to HTTP if it is. URLs would get added to the list either inside `urlParser.parse` or directly from the searchbar code, and would be removed from the list after a few seconds via a timeout.

lubiedo:
Thanks for pointing that out. Found myself contemplating the same security issue yesterday and I believe the only realistic solution would be to whitelist domains ([as done by HTTPS-Everywhere](https://github.com/EFForg/https-everywhere/tree/0a5edc8a4cb2c8f14ba2c327124e219e2a660b7e/src/chrome/content/rules)) and default to HTTPS-only on whatever domain matches. Maybe implement this with a top-1000 domains list?

This will only affect URLs where the protocol was not specified. Who knows, maybe in the future enforced to domains where HTTPS should be used instead of HTTP :)

lubiedo:
I think it would be a good idea to re-do this with a new approach in order to avoid either generating more noise or new attack vectors. I'll be closing this PR. Thanks!

PR #1640, opened 2021-06-09
Stop loading page when the Esc key is pressed

Sestowner:


js/defaultKeybindings.js
```diff
@@ -132,6 +132,8 @@ const defaultKeybindings = {
     })
 
     keybindings.defineShortcut({ keys: 'esc' }, function (e) {
+      !e && webviews.callAsync(tabs.getSelected(), 'stop')
```
PalmerAL, line 135: Why do we have to check the value of `e` here?

Sestowner: To prevent stop loading the page for e.x. when hiding the tab editor.

PalmerAL: `e` exists when the shortcut is triggered while the keyboard focus is in the browser UI, and doesn't exist when the keyboard focus is inside the webpage. But I think it's sometimes possible to have the UI focused even when the tab editor isn't opened. In that case, pressing esc would do nothing, which seems unexpected.

I think you could do this instead by adding a `tabEditor.isVisible` property, and then doing `if (!tabEditor.isVisible) { webviews.callAsync...` That would also avoid depending on the internal implementation details of the `keybindings` module, which seems like a bad idea.

Sestowner: > I think you could do this instead by adding a `tabEditor.isVisible` property...

Maybe this would be better, this also should prevent from stop loading page when dismissing the find in page

```
if (webviews.placeholderRequests.length === 0 && document.activeElement.tagName !== 'INPUT') {
        webviews.callAsync(tabs.getSelected(), 'stop')
}
```

or maybe just:

```
if (document.activeElement.tagName !== 'INPUT') {
        webviews.callAsync(tabs.getSelected(), 'stop')
}
```

PalmerAL: Yeah that seems like it would work. Probably good to keep the placeholder check.

PalmerAL:
Looks good.

PR #1644, opened 2021-06-10
Fix #1385

### LINKED ISSUE #1385: Downloads bar overlaps New Task button
- **Min Version:** 1.17.1
- **Operating system:** Arch KDE

## Expected Behavior
downlaods should not overlap anything

## Actual Behavior
New Task button is half visible
image

## To Reproduce
download something and open Task View

Sestowner:


PalmerAL:
It's probably better to find a new z-index value for the download bar that is less than the task overlay's. This might also require adjusting a few other indexes (in particular, I think the download bar needs to be in front of modal overlays).

Sestowner:
I'll look into that.

PalmerAL:
OK, makes sense.

PR #1645, opened 2021-06-11
Fix #1638

### LINKED ISSUE #1638: wrong file download on https://www.ekn.io/calendar/
- **Min Version:** 1.19.2

## Expected Behavior
it should let me download pdf file

## Actual Behavior
it downloads wrong file which is not pdf

## To Reproduce

go to this https://www.ekn.io/calendar/ and enter some date, press Download Calendar then it will open new page that doesnot load and download.html will appear in downloads bar

lubiedo:
Check if `Content-Disposition` header is provided and: if exists and defined as `attachment` then download the PDF, if not then use the PDF viewer.

### DIFF
--- a/main/download.js
+++ b/main/download.js
@@ -6,9 +6,15 @@ ipc.on('cancelDownload', function (e, path) {
   }
 })
 
+function isAttachment(header) {
+  return /^\s*attache*?ment/i.test(header)
+}
+
 function downloadHandler (event, item, webContents) {
   var itemURL = item.getURL()
-  if (item.getMimeType() === 'application/pdf' && itemURL.indexOf('blob:') !== 0 && itemURL.indexOf('#pdfjs.action=download') === -1) { // clicking the download button in the viewer opens a blob url, so we don't want to open those in the viewer (since that would make it impossible to download a PDF)
+  var attachment = isAttachment(item.getContentDisposition())
+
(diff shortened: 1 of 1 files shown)


main/download.js
```diff
@@ -54,7 +60,9 @@ function listenForPDFDownload (ses) {
   ses.webRequest.onHeadersReceived(function (details, callback) {
     if (details.resourceType === 'mainFrame' && details.responseHeaders) {
       var typeHeader = details.responseHeaders[Object.keys(details.responseHeaders).filter(k => k.toLowerCase() === 'content-type')]
-      if (typeHeader instanceof Array && typeHeader.filter(t => t.includes('application/pdf')).length > 0 && details.url.indexOf('#pdfjs.action=download') === -1) {
+      var attachment = isAttachment(details.responseHeaders[Object.keys(details.responseHeaders).filter(k => k.toLowerCase() === 'content-disposition')])
```
PalmerAL, line 63: Learned something new here :) I didn't know you could use an array to look up an object property like this, but I guess it works:
```
var x = {a: 1}
var y = x[["a"]] // = 1
```
It even works with nested arrays:
```
var y = x[[[[[["a"]]]]]] // = 1

PalmerAL:
Looks great, thanks!

PR #1647, opened 2021-06-11
Use URL if site has no <title>

lubiedo:
URL will be the tab title if no `title` tag exists. Currently tabs just show "New Tab" if that's the case.

### DIFF
--- a/js/navbar/tabBar.js
+++ b/js/navbar/tabBar.js
@@ -135,7 +135,11 @@ const tabBar = {
   updateTab: function (tabId, tabEl = tabBar.getTab(tabId)) {
     var tabData = tabs.get(tabId)
 
-    var tabTitle = (tabData.title || l('newTabLabel')).substring(0, 500)
+    // update tab title
+    var tabTitle = l('newTabLabel')
+    if (tabData.url != "") {
+      tabTitle = (tabData.title || tabData.url).substring(0, 500)
+    }
     var titleEl = tabEl.querySelector('.title')
     titleEl.textContent = tabTitle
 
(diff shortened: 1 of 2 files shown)


PalmerAL:
Good idea. One issue I noticed is that after you type a URL in the searchbar, it appears in the tab bar for a little bit before the page title shows up:

https://user-images.githubusercontent.com/10314059/121787220-7bed3280-cb8a-11eb-9c5d-e42e78010f6d.mov

It's a bit distracting, so it'd be good to hide that. I think what we could do is add an `isLoading` property to the tab object: https://github.com/minbrowser/min/blob/master/js/tabState/tab.js, connected to the `did-start-loading` and `did-stop-loading` events, and then keep using the "new tab" label instead of the URL if the page is loading.

lubiedo:
@PalmerAL Yes, added a `loaded` property to the tabs and now it will check that before changing any tab title.

PalmerAL:
Sorry for the delay. I made two changes:

* 57fa1bd6b52119ce92427142a172bbd729a3c5c4 changes updateTab so that if there is a `title` property, it will display even if the page is still loading. (Because the title is in the beginning of the document, it's available much sooner than loading is complete). On pages with no title, the URL will still not be displayed until loading is done.
* d824b146c05b041e0ef8f808def2a25ad461d283 marks `loading` as a temporary property so that it doesn't get saved to disk and restored when the browser is restarted.

If those changes look good to you, then I think this is done.

lubiedo:
Both changes look great, thanks!

PalmerAL:
LGTM then!

PR #1648, opened 2021-06-12
Display download bar in front of modal overlays

Sestowner:


PalmerAL:
Which overlay are you referring to? The searchbar and dialog overlays have a z-index of either 1 or 5, so the download bar should be on top of them already: https://github.com/minbrowser/min/blob/7749a35ea71a5f373c05e1d587b620a7a5f7c1fc/css/modal.css

In the other PR, I was thinking that if you changed the download bar z-index, the overlay's z-index might need to change as well, but since the download bar is unchanged, everything should still just work.

Sestowner:
Oops , I meant "Display download bar in front of modal **dialogs**".

Sestowner:
Before:
image

After:
image

PalmerAL:
Oh, that's a good point. It looks like we should just limit the height of the dialog? Something like `.modal { max-height: 85vh }` might solve this better.

PR #1652, opened 2021-06-13
Change bookmark editor behavior

PalmerAL:
Previously, clicking on the bookmark star or pressing ctrl+d while on a page you already bookmarked would delete the bookmark. There were two problems with this:

* If you didn't remember whether you'd bookmarked the page, but wanted to bookmark it, pressing ctrl+d would sometimes do the opposite of what you wanted (by deleting the bookmark).
* Once you'd bookmarked a page, there wasn't an easy way to edit the bookmark's tags without going to the bookmark viewer.

This PR changes the behavior so that clicking the star or using the shortcut will always save the page and open the bookmark editor, regardless of whether it was previously bookmarked. The editor now has a separate button to delete the bookmark:
Screen Shot 2021-06-13 at 3 03 29 PM

### DIFF
--- a/css/bookmarkManager.css
+++ b/css/bookmarkManager.css
@@ -17,9 +17,18 @@
   box-shadow: 0 0 0 1px rgba(255, 255, 255, 0.2);
 }
 
+.bookmark-editor .tag-edit-area {
+  display: inline-block;
+}
+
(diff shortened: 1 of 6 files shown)


lubiedo (approved):

PalmerAL:
Thanks for reviewing.

Issue #1646, opened 2021-06-11
Page history - back and forward buttons


ep1955:
I notice there is still no page forward button to go with the page backward button.  Would be a really great feature to have since in a browser your hand is typically on the mouse anyway.  Can't be any more difficult than a back button, any chance we'll see one soon?  I can't be the only user that would like to see this.  Thanks!

lubiedo:
I second this and would like to propose a solution: what if the back button is transformed to the forward button if the meta key is pressed?

Like this:

![image](https://user-images.githubusercontent.com/63729414/121922262-0652a500-ccf7-11eb-9633-7e1d98f9fea0.png)

lubiedo:
Merged on #1665.

PR #1653, opened 2021-06-14
Forward button on Meta key pressed

lubiedo:
Issue: #1646 

This PR turns the back button into a forward button when the meta key is pressed down.

### DIFF
--- a/js/navbar/goBackButton.js
+++ b/js/navbar/goBackButton.js
@@ -15,6 +15,14 @@ var goBackButton = {
     goBackButton.element.classList.remove('carbon:chevron-' + (!this.forward ? 'right' : 'left'))
     goBackButton.element.classList.add('carbon:chevron-' + (!this.forward ? 'left' : 'right'))
 
+    // update button localization
+    let dataLabel = (!this.forward ? 'goBack' :  'goForward')
+    goBackButton.element.setAttribute('data-label', dataLabel)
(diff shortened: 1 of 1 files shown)


PalmerAL:
I like the idea! I think it's probably easy enough to click on. One concern I have is that this also activates when you use any other keyboard shortcut - for example, if you press cmd+1 to switch tabs, the back button will briefly flash. Maybe alt would work better, although that might cause the same issue when typing in some languages.

One other idea I had was to show the back button on hover:

https://user-images.githubusercontent.com/10314059/122296290-dacee680-cebf-11eb-83d5-222b75d7fa23.mov

And branch (with some fairly messy code) if you want to try it out: https://github.com/minbrowser/min/tree/forward-2

Not sure what's best yet; let me know what you think.

lubiedo:
I agree, it's kind of annoying seeing that changing every time the Meta key is pressed and also liked your approach. What I tried to do was to keep the navbar as simple as possible thus only changing the arrow. But after some analysis I believe your approach is better as it wouldn't require a new user to know what key to press to enable the forward button and would be more intuitive.

Using the Alt key can be another option as it is not used on any key mapping in Min.

Anyway, I've been using this branch and I can say that a forward button is really useful :)

lubiedo:
Closing this PR as #1665 was created.

PR #1659, opened 2021-06-18
Right arrow to partially autocomplete URL

lubiedo:
Pretty simple PR but kinda useful. While you are writing a URL you can go to the suggestion and autocomplete the rest using the right arrow.

Example:

https://user-images.githubusercontent.com/63729414/122600972-d1638c80-d02d-11eb-8656-41d53417b883.mov

### DIFF
--- a/js/searchbar/searchbarPlugins.js
+++ b/js/searchbar/searchbarPlugins.js
@@ -93,6 +93,16 @@ const searchbarPlugins = {
       item.addEventListener('click', function (e) {
         URLOpener(data.url, e)
       })
+
+      item.addEventListener('keyup', function(e) {
+        /*  right arrow or space should autocomplete with selected item if it's
+            a search suggestion */
+        if (e.keyCode === 39 || e.keyCode === 32) {
+          let input = document.getElementById('tab-editor-input')
+          input.value = data.url
+          input.focus()
+        }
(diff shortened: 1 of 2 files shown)


PalmerAL:
Good idea!

Two comments:
* `searchbarUtils.createItem` is sort of a generic list item, and is also used for the list in the task overlay, so I'm hesitant to put searchbar-specific code in there. However, I can't think of a better place for it; the only other possibility would be [searchbarPlugins.addResult](https://github.com/minbrowser/min/blob/master/js/searchbar/searchbarPlugins.js#L75), but not all of the searchbar items are created using that path. So I'm fine with merging this as-is unless you can think of a better idea.
* There was a very similar PR a while ago that never got finished (#1009), and one of the ideas there was to do the partial autocomplete when the space key was pressed. It would probably be nice to support both space and right arrow keys here.

lubiedo:
Thanks for the feedback!
Regarding your comments:

 * Agree. Although it didn't make that much difference it made more sense to move it to `searchbarPlugins.addResult` as the check for the `data.url` was already there along with the click event for the result item.
 * Yeah, added the Space as a key for this functionality.

Let me know what you think.

PalmerAL:
Looks good!

PR #1663, opened 2021-06-23
Add HTTPS upgrade from list of known sites

lubiedo:
Another approach to PR #1635. In this case all HTTPS upgrades are done based on a top 1 million sites list freely available on https://tranco-list.eu/. By default, the list is truncated at 1000 sites to not overdo it.

The URL will be upgraded to HTTPS in two situations when the domain appears in the whitelist:
 * When the protocol is missing, `https://` will be prepended.
 * When the protocol is `http://`, the protocol will be replaced.

This will require more testing to avoid any issue.


js/util/urlParser.js
```diff
@@ -118,6 +130,11 @@ var urlParser = {
     } else {
       return encodeURI('file://' + path)
     }
+  },
+  isHTTPSUpgreadable: function (url) {
+    let domain = /^([^\/$]+)/.exec(urlParser.removeProtocol(url))
```
PalmerAL, line 135: I'm not sure why there's a $ inside the regex here: if you had something like `example$.com/abc`, it matches "example". I don't think that could ever realistically happen, but probably still worth fixing.

PalmerAL:
This looks pretty good overall!

This was in place before your PR (and in fact goes back to the [very first commit, 6 years ago](https://github.com/minbrowser/min/blob/1acb58d19ad25960971e7116f3858a0dffc4f765/js/urlParser.js)), but I was looking at the `urlParser.removeProtocol` function again, and it's kind of silly. It should only be removing protocols at the beginning of the URL, and it shouldn't be removing "www.". That should be fixed before this is merged.

It might also be worth checking how fast `isHTTPSUpgradeable` is with a lot of iterations. It's possible that putting the domain list inside an object like `{example.com: true}` could be faster, since you would only be doing 1 object lookup instead of searching through the whole array. But it might also be fast enough already to not matter.

js/util/urlParser.js
lubiedo: Yes, you are right, I don't know what that `$` is doing there. Will fix it

lubiedo:
Thanks! Regarding your comments:

1) Fixed on e2fcee1 but will require testing as it will affect `urlParser.basicURL`. For example, secondary texts of tabs in the task list will show the complete FQDN showing `www.google.com` instead of `google.com` which can't be a great deal other than a detail.
2) Thought about it but during some tests I had always the same results around **0.001 ms** on all domain searches. If this is a real impact we can create a buffer that will cache latest upgrades. What do you think?

PalmerAL:
1. It's probably best to move the www. removal to happen in basicURL? But I agree it's a pretty minor detail.
2. Thanks for checking! That's fast enough that it should be fine (since it's not run inside a loop or anything); a cache would be overkill.

lubiedo:
Pushed your suggestion on point 1. in 718f41b

PalmerAL:
Looks good!

PR #1665, opened 2021-06-25
Show forward button on hover

PalmerAL:


PalmerAL:
@lubiedo Let me know what you think. (Github wouldn't let me mark you as a reviewer; I just added you to the repository so that should work in the future).

lubiedo:
Looks really good and code looks clean. There's only one (small) detail which is the horizontal scrollbar under the tabs that it gets visible for a second. Maybe disabling the overflow while the mouse is over the buttons will avoid this behavior?

https://user-images.githubusercontent.com/63729414/123441480-24d75c80-d591-11eb-961d-867f37e9c8ac.mov

If this is not an issue then I think that it can be merged!

PalmerAL:
Thanks for pointing that out! I've done what you suggested in b27b5315bc5869e3b640f8aa3da34207a3df0994 - let me know if you see any more issues after that.

lubiedo:
Yep, that did the trick. Did some tests and looks really good.

lubiedo (approved):

PalmerAL:
Great!

Issue #1571, opened 2021-04-16
can't import bookmarks from safari 


vivek-ram:
Mac OS m1 Mac can't import bookmarks from safari

PalmerAL:
Can you provide more information? What happens when you try to import them?

vivek-ram:
well I downloaded the bookmarks from safari and I imported them but can see any bookmarks in min browser I don't know if it actually imported or not

lubiedo:
@vivek-ram This should be fixed with PR #1666. Let us know if you still are experiencing issues.

PR #1666, opened 2021-06-29
Fix issue #1571

lubiedo:


PalmerAL:
I reformatted with standard-format in cccc750bbe3140d173157cd3112654d4043ded58 (that's not documented very clearly, sorry about that). Other than that, looks good!

PR #1667, opened 2021-06-30
Validate domain names via urlParser

lubiedo:
Based on discussion #1658. Tests can be found in the source code but this should be thoroughly tested.


PalmerAL:
I like the idea of trying something simpler, but I’m not sure this is enough to solve the issue. It will correctly identify “android.arch.lifecycle:extensions:2.1.0” as being a search, but it will still treat “android.arch.lifecycle” as a domain. I don’t think there’s any way to fix that without using the public suffix list.

Also, if we do go with the regex approach, there’s a few cases this fails on:
* domains with capital letters
* Anything with a port, like “localhost:3000”

lubiedo:
Yes. That's why I thought it was a _primitive_ approach. Let me work on a way to include the public suffix while still keeping it simple.

Regarding the regex:
* I'm transforming it to lower case via `urlParser.getDomain` but I can add a **i** modifier to the regular expression to ensure that it will match with case insensitive.
* Yes, my bad. I can add that to the regular expression.

js/util/urlParser.js
```diff
@@ -25,9 +28,8 @@ var urlParser = {
     if (url.indexOf(' ') === -1 && url.indexOf('.') > 0) {
       return true
     }
-    // a host from the hosts file is also a URL
-    var hostPart = url.replace(/(:|\/).+/, '')
-    return hosts.indexOf(hostPart) > -1
```
PalmerAL, line 30: I don't see this check in `validateDomain`; I think it needs to be added back there so that custom hosts added to the hosts file work. localhost is usually part of the host file, so you should be able to get rid of the localhost check once you add this back.

js/util/urlParser.js
```diff
@@ -25,9 +28,8 @@ var urlParser = {
     if (url.indexOf(' ') === -1 && url.indexOf('.') > 0) {
```
PalmerAL, line 28: The meaning of `isURLMissingProtocol` is sort of confusing at this point. Originally, the intent was to determine "could this be a URL, if you added a protocol to it?". So "google.com" would return true, but "google" would return false.

Since `validateDomain` is also used to determine whether a string is a valid URL, I think this function should call validateDomain before returning. And then inside parse, you could remove validateDomain from this line:
```
     if (urlParser.isURLMissingProtocol(url) && urlParser.validateDomain(urlParser.getDomain(url))) {
```

js/util/urlParser.js
```diff
@@ -130,8 +132,64 @@ var urlParser = {
...
+        google.com..            = false
+        [:::1]:8080             = true
+        [2001:0db8::0001:0000]  = true
+        [invalid]               = false
+        localhost               = true
+    */
+    if (!(/^(?!-)(?:.*@)*?([a-z0-9-\._]+[a-z0-9]|\[[\:a-f0-9]+\])(?:\:[0-9]*|\.{1}){0,1}$/i.test(domain) || domain.length > 255)) {
+      return false
+    }
+    let clean_domain = RegExp.$1
```
PalmerAL, line 164: This is cool, didn't know this feature existed.

js/util/urlParser.js
```diff
@@ -130,8 +132,64 @@ var urlParser = {
...
+    }
+
+    return false
+  },
   isHTTPSUpgreadable: function (url) {
-    let domain = /^([^\/]+)/.exec(urlParser.removeProtocol(url))[0]
+    let domain = urlParser.getDomain(url)
+
+    if (urlParser.validateDomain(domain) === false)
+      return false
```
PalmerAL, line 192: If the url isn't a valid domain, it won't be in the top sites list, so this is redundant.

PalmerAL (commented):
I think this is looking pretty good overalll; I've left a couple comments.

> I'm transforming it to lower case via urlParser.getDomain but I can add a i modifier to the regular expression to ensure that it will match with case insensitive.

Oops, sorry; I was directly testing the regex against strings instead of calling the function, so I didn't notice that.

> EDIT: also, just learnt that lifestyle is an actual TLD :D

Yeah, there's a lot of fun ones now!


One more thing: the combination of checking the public suffix list and doing HTTPS upgrades does seem to be causing some performance issues. I took a profile of this after typing in the searchbar, and a single call to parse() is taking 2.5 ms, which is not great:
Screen Shot 2021-07-03 at 10 17 47 PM

Some ideas for how we could fix it:
* When generating the suffix list, we could strip out lines that start with ! or are wildcards, and add a `.` before the suffix if necessary. Then, we could remove the chain of `if` statements that does this in the `validateDomain` call, which should help a lot.
* Similarly, we could encode the items in the suffix / https lists as punycode ahead of time, which would allow us to get rid of the punycode decoding in urlParser.
* If that isn't enough, we could turn the suffix list into a dictionary, and then look up each domain going backwards (i.e. if you're given "a.b.example.com", you look up "com", "example.com", "b.example.com", "a.b.example.com"... and see if any are a suffix).

PalmerAL:
Also - can you format this with standard-format as well?

js/util/urlParser.js
lubiedo: Isn't `isURLMissingProtocol` supposed to check if the URL is just missing the protocol (ie. https://)?
I added a check using the `hosts.js` list to the validateDomain function to ensure that domains specified in the hosts files are treated as valid. Gonna push it as soon as I get the other reviews done.

js/util/urlParser.js
lubiedo: Wouldn't that confuse more the purpose of `isURLMissingProtocol`? Also, `validateDomain` is used to determine if the domain contained in a URL is valid, that's why I think that keeping those two in separated procedures is important.

To be honest, I think that `isURLMissingProtocol` should be resumed to this:
```js
  isURLMissingProtocol: function (url) {
    return !urlParser.protocolRegex.test(url)
  }
```
As everything that wasn't validated before it should fall into the search spectrum.

lubiedo:
I had time to work on your reviews and got some good output. I was able to shorten the amount of time that it takes to parse and validate the domains by generating checksums (using the fletcher32 algorithm) of the domains already validated and store them in an array, or cache, for future use. This is a fresh navigation to minbrowser.org:

image

After the domain is cached the validation takes usually less than 15 ms. The validation against the public suffix is done if it is not an IP address and if it doesn't exist on the locally defined hosts (ie. /etc/hosts).

I also took the liberty to clean the code a little bit and get rid of some regular expressions. Regarding punycode, now it's  saving it already converted and just checked when needed if the domain contains any unicode character. Also, formatted the code according to the standard, as requested.

Please, let me know what you thing and if you had the chance to test it.

PalmerAL:
It's looking better! I've done some testing with various URLs, and haven't run into any issues so far.

The issue with caching is private tabs - if the user types a URL while in a private tab that gets parsed, we don't want to keep that in memory after the tab is closed. It is true that it's just a checksum, not the URL, but it would be pretty easy to figure out the real domain by guessing and comparing the checksum.

You could probably fix that by coming up with some method of evicting things from the cache when a private tab is closed, but you may not need the cache anyway. I took the latest commit, removed the cache, and then benchmarked this with a script that calls parse() on random domains:

```
console.time()
for (var i = 0; i < 1000; i++) {
    urlParser.parse(Math.random() + ".com");
}
console.timeEnd();
```

And got the following results:
```
before - 4ms
The original PR - 289 ms
The latest commit, with caching removed - 23ms
```

So I think the other changes (not doing punicode parsing / removing if statements) are enough that we probably don't need caching any more.


<hr>

Regarding `isURLMissingProtocol`: this is used in a [few places in the searchbar](https://github.com/minbrowser/min/search?q=isURLmissingprotocol) as a fast way to determine if a string is "URL-like". So for example, if you start typing a URL like "example.com", either with or without a protocol, the instant answer plugin can stop providing suggestions, but if you're typing something that looks like a search it will show suggestions:
https://github.com/minbrowser/min/blob/7749a35ea71a5f373c05e1d587b620a7a5f7c1fc/js/searchbar/instantAnswerPlugin.js#L172

I agree the naming is confusing there, so I'm open to whatever approach you want to take for solving that. Perhaps we could get rid of `isURLMissingProtocol` entirely, and make a new method `looksLikeURL?`

lubiedo:
Thanks for the reviews! 

Regarding the cache: I think you have a good point there and decided that the checksum and validation of the domain maybe wasn't a good idea in that context. I still kept a cache for the public suffixes that are valid for a reason which is this: in some circumstances it cuts to half the validation if the public suffix is already known. 

image
image

It is not a big deal but could be useful so the checking of a valid suffix is done against a smaller array compared to the public suffix list (which I shrank to avoid looking for unnecessary suffixes).
Should we keep this cache or remove it?

---

Wrote the function `isPossibleURL` as a replacement for the previous `isURLMissingProtocol`. Let me know what you think about the naming and function. I've tested it with searches and instant answers and it's been working fine so far.

--- 

Made some small changes to the code like getting rid of regexes in favor of functions, letting`getDomain` get rid of the port in the URL (ie. `:8080` in `localhost:8080`), among others.

PalmerAL:
> It is not a big deal but could be useful so the checking of a valid suffix is done against a smaller array compared to the public suffix list (which I shrank to avoid looking for unnecessary suffixes).
Should we keep this cache or remove it?

That seems like a good idea. I was thinking you could also make a short list of the most common suffixes (com, net, org, etc) as an alternative, but that wouldn’t be quite as effective as this cache - depending on where you live, for example, most urls may have a specific country code ending.

There is some privacy risk that if you visit a site with an unusual suffix, it might be possible to look at suffixes in memory and figure out what the site was. But that seems like a really remote possibility, so I’m fine with accepting that.

All the other changes look good - I’m good with merging this if you are.

lubiedo:
Yes, I think that it could be identifiable to an specific session but that would require a very specific suffix, I suppose. What we can do, also, is to add a check to ensure that this is not applied to private sessions:
```js
    if (tabs.get(tabs.getSelected()).private) {
      return true
    }
```
What do you think?

If you are OK with it I also think that we can merge it. Or maybe there's another input from a third collab?

PalmerAL:
I don't think you can assume the URL being parsed is from the current tab. (In practice, it probably is true, but it's not a good assumption to make). So you'd need to pass the tab ID into parse(), which seems like a less-than-ideal design.

Either way is fine with me - I think the cache could be removed without too much impact either.

lubiedo:
Removed the cache and made other minor changes. Let me know if everything looks good and if we can merge.

PalmerAL:
The change in the last commit to remove entries with * from the suffix list entirely, instead of just removing the * from the front, doesn't seem right. Looking at the first removed entry in the list, `*.bd`, a domain like "example.com.bd" should be allowed, but now it won't be counted as a valid domain.

Everything else looks good.

lubiedo:
That was a test that slid into the PR 😅 Reverted.

PalmerAL:
Everything looks good - thanks for working on this!

PR #1674, opened 2021-07-07
Popup filtering and block of third-party cookies

lubiedo:
Simple popup filtering by using the current filters via easylist/easyprivacy/customFilters and adblock's 'popup' option.
Discussion: #1660 

Filter that will block popups (via `window.open`, for example) that open google.com:
```
||google.com^$popup
```

### DIFF
--- a/main/filtering.js
+++ b/main/filtering.js
@@ -49,9 +49,6 @@ function initFilterList () {
     if (err) {
       return
     }
-
-    // data = data.replace(/.*##.+\n/g, '') // remove element hiding rules
-
     parser.parse(data, parsedFilterData)
   })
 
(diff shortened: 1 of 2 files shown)


lubiedo:
An addition to this PR that I believe it is important (the idea of privacy is what makes Min attractive): blocking third-party cookies.
For testing I used [this resource](https://alanhogan.github.io/web-experiments/3rd/third-party-cookies.html) which I believe is pretty accurate as I tested it also on FF and Safari and blocked third-party cookies successfully. Just make sure you already cleaned cache and no cookies from the testing sites are present (or just open a private tab).

More information on third-party cookies and tracking [1] [2].

The question is: blocking all third-party cookies? Enabled by default if blocking is enabled?

[1] https://en.wikipedia.org/wiki/HTTP_cookie#Third-party_cookie
[2] https://blog.mozilla.org/security/2021/06/01/total-cookie-protection-in-private-browsing/

PalmerAL:
Pop up filtering looks good from skimming it; I'm going to read over it more closely and test it when I get a chance.

It's possible to set cookies from JS too, which I don't think this would block. It looks like chrome has a "disable third party cookies" flag; ideally we could just enable that, because it would be easier and more reliable. But it's not clear to me if Electron supports that or not.

If we do the filtering ourself, I would vote for enabled by default following the existing exception system. But if we can use Chrome's global flag, it may not be possible to have site exceptions.

lubiedo:
You're right. The JS option will be another world. AFAICT Chrome has SameSite set as Lax by default and haven't found a chromium flag other than`--enable-experimental-cookie-features` that actually worked. I've just tested JS third-party cookies using [this](https://github.com/mindmup/3rdpartycookiecheck). I was thinking that another way would be to partition cookie sessions by site and that way filter/isolate cookies on both cases. What do you think?

PalmerAL:
> partition cookie sessions by site and that way filter/isolate cookies on both cases

Do you mean something like "first party isolation" in Firefox?

lubiedo:
Exactly, that's what I was trying to describe.

PR #1675, opened 2021-07-08
:hammer_and_wrench: + :lipstick: |  fix + pretty + clean

miko-github:
# :hammer_and_wrench: + :lipstick: :=  fix + pretty + clean
- fix : string concats with literally string
- pretty : hard readable if condition with constant value
- fix :  hard readable constant with functionality checking

:v: :heart:

### DIFF
--- a/main/main.js
+++ b/main/main.js
@@ -12,7 +12,31 @@ const {
   Menu, MenuItem,
   crashReporter,
   dialog
-} = electron
+} = electron;
+
+// TODO : this constant and other should be in other files
+// [?] OR -> PLATFORM_WIN32
+const PLATFORM_WIN = 'win32';
+/**
+ * and, we can use unformal constants as like
+ * ```js
+ * // assign
+ * PLATFORM = { WIN: 'win32' };
+ * // usage
+ * PLATFORM.WIN
+ * ```
+ */
+// OTHER WAY IS:
+/**
+ * use function check, be like :
+ * `isPlatform('win32')` or `isWin32()`
+ * or `isPlatformWin()`
+ */
+const PLATFORM_DARWIN = 'darwin';
+
+function isPlatform(pf){
+  return process.platform === pf;
+}
 
 crashReporter.start({
   submitURL: 'https://minbrowser.org/',
@@ -21,11 +45,12 @@ crashReporter.start({
 })
 
 if (process.argv.some(arg => arg === '-v' || arg === '--version')) {
-  console.log('Min: ' + app.getVersion())
-  console.log('Chromium: ' + process.versions.chrome)
+  console.log(`Min: ${app.getVersion()}`)
+  console.log(`Chromium: ${process.versions.chrome}`)
   process.exit()
 }
 
+// arg === '--development-mode' -> arg === CMD_DEV_MODE
 let isInstallerRunning = false
 const isDevelopmentMode = process.argv.some(arg => arg === '--development-mode')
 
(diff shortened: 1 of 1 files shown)


main/main.js
```diff
@@ -21,21 +45,23 @@ crashReporter.start({
...
+// arg === '--development-mode' -> arg === CMD_DEV_MODE
 let isInstallerRunning = false
 const isDevelopmentMode = process.argv.some(arg => arg === '--development-mode')
 
 function clamp (n, min, max) {
   return Math.max(Math.min(n, max), min)
 }
 
-if (process.platform === 'win32') {
+if (process.platform === PLATFORM_WIN) {
```
lubiedo, line 61: Isn't `process.platform === PLATFORM_WIN` exactly the same as doing `process.platform === 'win32'` and similars?

main/main.js
```diff
@@ -277,7 +303,7 @@ function createWindowWithBounds (bounds) {
   so registering a handler causes events to happen twice.
   See: https://github.com/electron/electron/issues/18322
   */
-  if (process.platform === 'win32') {
+  if (isPlatform('win32')) {
```
lubiedo, line 306: Why here are you using `isPlatform('win32')` but in other changes `process.platform === PLATFORM_WIN`?

main/main.js
```diff
@@ -21,21 +45,23 @@ crashReporter.start({
...
 
 if (process.argv.some(arg => arg === '-v' || arg === '--version')) {
-  console.log('Min: ' + app.getVersion())
-  console.log('Chromium: ' + process.versions.chrome)
+  console.log(`Min: ${app.getVersion()}`)
+  console.log(`Chromium: ${process.versions.chrome}`)
   process.exit()
 }
 
+// arg === '--development-mode' -> arg === CMD_DEV_MODE
```
lubiedo, line 53: What's the purpose of this?

lubiedo (commented):
Hi @miko-github. I'm kinda confused about these changes and left some reviews.

miko-github:
> Hi @miko-github. I'm kinda confused about these changes and left some reviews.

# Hi @lubiedo 👋

Actually which line? tag them to me and let me to explain 😉🤗

main/main.js
miko-github: Yes, but, constant variables be clean and human readable as like Enums  😉✌

main/main.js
miko-github: I just commuted options 😉

I just say, we can use this very clear....

Use function or constant or enum 😁

main/main.js
miko-github: Actually, constant variables again but in optional, and you can change name to anything else you want 🤗

But I'm suggesting you, using enum, I know enum isn't available in JS but we can make similar of them...

```js
const SHELL_FLAG = {
  DEVELOPER_MODE: '--developer-mode'
}
```
In usage this be like enum,
```js
SHELL_FLAG.DEVELOPER_MODE
```

PalmerAL:
I’m also fairly confused by this. Putting things in constants doesn’t automatically make them more readable; constants can help at reducing duplication and making it easier to change values, but here nothing’s really being duplicated to start with. And you can’t even reuse the same value everywhere, because `navigator.platform` and `process.platform` use different platform names, and we use both.

miko-github:
> I’m also fairly confused by this. Putting things in constants doesn’t automatically make them more readable; constants can help at reducing duplication and making it easier to change values, but here nothing’s really being duplicated to start with. And you can’t even reuse the same value everywhere, because `navigator.platform` and `process.platform` use different platform names, and we use both.

# Okay :+1: , now i understood :smile:, thanks :hugs: :heart:

PR #1676, opened 2021-07-08
:lipstick: | removed unusable `else` keyword :v:

miko-github:
# :lipstick: | removed unusable `else` keyword :v:

## with return keyword, else is useless

when we use `return` in if statement, that meant, when condition is true, return value and close function, , and if not, then, go on outside of if statement, and run next line, and then return other value and break function

### DIFF
--- a/main/menu.js
+++ b/main/menu.js
@@ -6,9 +6,8 @@ function buildAppMenu (options = {}) {
       if (Array.isArray(value)) {
         // value is array if multiple entries are set
         return value[0].replace('mod', 'CmdOrCtrl')
-      } else {
-        return value.replace('mod', 'CmdOrCtrl')
       }
+      return value.replace('mod', 'CmdOrCtrl')
     }
 
     return null


lubiedo (approved):

PalmerAL:
This isn’t wrong, but in general I don’t think it makes sense to spend time on these minor formatting changes; they take time to review, and there just isn’t much benefit to doing them.

I’m going to close this (and similar formatting PRs), but I have two suggestions for things you could do:
* You could work on setting up a better automatic formatting tool, like prettier, that would do this across the whole codebase automatically.
* You could take a look at some of the bugs / features in the issue tracker and work on those.

miko-github:
> This isn’t wrong, but in general I don’t think it makes sense to spend time on these minor formatting changes; they take time to review, and there just isn’t much benefit to doing them.
> 
> I’m going to close this (and similar formatting PRs), but I have two suggestions for things you could do:
> 
> * You could work on setting up a better automatic formatting tool, like prettier, that would do this across the whole codebase automatically.
> * You could take a look at some of the bugs / features in the issue tracker and work on those.

# Hi :wave: 
yes, i'm agree with you but i just didn't know about the min browser code base and core :hugs: 
i want but i can't, cause .... :smile: 

### But, I try :v:

PR #1684, opened 2021-07-16
Block unwanted popups (take 2)

lubiedo:
Closed previous PR to avoid confusion. This is to block popups by using the EasyList filters.


PalmerAL:
I think the overall approach looks pretty good, and I would like to see this implemented.

One thing I've noticed is that there's a difference in how other blockers and Min check element types in this case. Given these two filters in Easylist:
```
/advertising?
/earn.php?z=$popup,subdocument
```

Min will block popups matching both; while uBlock will only block popups matching the second one. I think that's what we want to do too (only consider rules with a `$popup` element type when blocking popups).

The filter parser parses the filter's element types here: https://github.com/minbrowser/min/blob/master/ext/abp-filter-parser-modified/abp-filter-parser.js#L117-L129, and then compares the filter's element types to the request here: https://github.com/minbrowser/min/blob/master/ext/abp-filter-parser-modified/abp-filter-parser.js#L448-L450. If the filter doesn't specify any element types, it assumes all element types should be blocked. I think that should be "all element types except popups", but it'd be nice to find some more documentation / investigate what other adblockers do here first.

lubiedo:
Had time to take a look today and realized that there is was virtually no strict matching after you mentioned it, for that reason I pushed 398be23. Do you have some time to take a look? Basically it makes sure that the matching will be forced (via `strictType` filter option) to also check the type and not only the rule matching. So, based on your example: a `/advertising?` popup will open and `/earn.php?z=$popup,subdocument` will not as the types being filtered (`popup` in this case) match.

PalmerAL:
Having a strict / non-strict mode doesn't seem entirely right to me, but I'm also having a hard time wrapping my head around what should happen here, so maybe I'm misunderstanding. There isn't much documentation on the filter parsing (and no formal specification), but the current behavior is that:

* `/advertising?` - matches all element types
* `/advertising?$script,style` - matches only scripts and stylesheets
* `/advertising?$~script` - matches all element types, except scripts
* `/advertising?$~script,style` - effectively the same thing as above

The way this gets implemented is that the element types of a filter are stored in a bitfield ([code](https://github.com/minbrowser/min/blob/master/ext/abp-filter-parser-modified/abp-filter-parser.js#L117-L129)), one bit per element type. If a `not` filter is parsed first, the filter starts with a bitfield that's all 1s; otherwise it starts with a bitfield that's all 0s. Then each element type listed in the filter is parsed, and the bit is set to 0 or 1.

What I think uBlock origin etc. are doing is:

* `/advertising?` - matches all element types, **except popups**
* `/advertising?$script,style` - matches only scripts and stylesheets
* `/advertising?$~script` - matches all element types, except scripts **and popups**
* `/advertising?$popup` - matches popups

The closest thing there is to a spec (https://help.eyeo.com/en/adblockplus/how-to-write-filters#options) seems to support this:
> popup - pages opened in a new tab or window Note: Filters will not block pop-ups by default, only if the $popup type option is specified.

So I think that the right solution is to change the "default bitfield with all 1s" to "bitfield with all 1s, except with a 0 for popups" ([code](https://github.com/minbrowser/min/blob/master/ext/abp-filter-parser-modified/abp-filter-parser.js#L120)), which would solve this issue without needing a separate "popup mode".

lubiedo:
I think that by having an strict filter by type is a good way to enforce it from the filtering perspective leaving aside the rules and how are they constructed. Meaning that, before 398be23, we wouldn't enforce that `matches` will *only* match the element type thus making the spec that you shared impossible to emulate. We can take it and copy what uBlock is doing, also.

PalmerAL:
The issue I see with a strict mode is that "strictness" shouldn't be dependent on the context in which the URL is evaluated - with this, we're using "strict mode" when calling matches() from the popup filtering function, and not using it elsewhere. But "strictness" is actually a property of the `$popup` filter option, and that strictness should apply from everywhere matches() is called.

The filter parser already takes in the element type options and builds a list of which element types are allowed for each filter, so it seems natural to me to make popups disallowed in that list unless specifically enabled. This matches exactly with the behavior of "Filters will not block pop-ups by default, only if the $popup type option is specified.".

I implemented that behavior in a branch here: https://github.com/minbrowser/min/commit/d66705bf3da011242865227ec95f8036c28e0c42. Unless you see an issue with it, I think that's the way to go.

lubiedo:
I'm kinda confused, https://github.com/minbrowser/min/commit/d66705bf3da011242865227ec95f8036c28e0c42 is supposed to block the rules with the `$popup` option? During some tests it wasn't able to stop any of the existent nor custom rules for popups.

PalmerAL:
Oh, sorry, that commit just implements the changes to the filter parser. The idea would be to combine that with the changes to filtering.js in this PR.

PalmerAL:
Yup, it's working now for me too. LGTM!

Issue #1681, opened 2021-07-15
Min fails to launch with Italian system language


macmau74:
I encountered an issue with the latest version of Min 1.20.0 on MacOs 10.14 Mojave on my iMac intel.
Min starts up, menus activate but without working. I can't type in the address bar and I can't navigate.

lubiedo:
Hi @macmau74, this happens on version 1.20.0-beta or 1.19.2? Also, if you open the Inspector console for the browser (Option + Cmd + Shift + i) do you see any kind of error?

macmau74:
Hi
I’m on iMac intel, I have installed Min-v1.20.0-darwin-x64.
This is the report of the Inspector console for the browser:

Uncaught TypeError: Cannot read property 'addEventListener' of null
    at Object.initialize (bundle.js:3236)
    at Object.27.browserUI.js (bundle.js:3375)
    at o (bundle.js:1)
    at bundle.js:1
    at Object.3.passwordManager/managerSetup.js (bundle.js:334)
    at o (bundle.js:1)
    at bundle.js:1
    at Object.1.autofillSetup.js (bundle.js:236)
    at o (bundle.js:1)
    at r (bundle.js:1)
bundle.js:1615 Uncaught TypeError: Cannot read property 'getSelected' of undefined
    at EventEmitter.<anonymous> (bundle.js:1615)
    at EventEmitter.emit (events.js:315)
    at Object.onMessage (electron/js2c/renderer_init.js:93)
bundle.js:2161 Uncaught TypeError: Cannot read property 'get' of undefined
    at bundle.js:2161
bundle.js:1623 Uncaught TypeError: Cannot read property 'getSelected' of undefined
    at EventEmitter.<anonymous> (bundle.js:1623)
    at EventEmitter.emit (events.js:315)
    at Object.onMessage (electron/js2c/renderer_init.js:93)
bundle.js:9791 Uncaught TypeError: Cannot read property 'get' of undefined
    at captureCurrentTab (bundle.js:9791)
    at bundle.js:10271
captureCurrentTab @ bundle.js:9791
(anonymous) @ bundle.js:10271
2bundle.js:2161 Uncaught TypeError: Cannot read property 'get' of undefined
    at bundle.js:2161
(anonymous) @ bundle.js:2161
bundle.js:9791 Uncaught TypeError: Cannot read property 'get' of undefined
    at captureCurrentTab (bundle.js:9791)
    at bundle.js:10271
[...]

PR #1687, opened 2021-07-18
Fix it.json

lubiedo:
Fixes issue #1681


PalmerAL:
Thanks! Looks good.

I feel sort of silly for not catching this; I'll make a new release pretty soon so this can go into it. I looked at the statistics data to see how big a problem this is, and Italian was about 3% of users prior to the 1.20 release; it's dropped to 2% since then (which is indicative of how slowly people upgrade to new releases more than anything else). So it's probably fine to advise anyone else with this issue to keep using 1.19.2 in the meantime.

PR #1692, opened 2021-07-20
Fix #1691: Add margin to bookmark-editor

### LINKED ISSUE #1691: Border on last bookmark disappears during bookmark edit
- **Min Version:** 0.20.0
- **Operating system:** macOS 11.4

## Expected Behavior
Line should be visible.

## Actual Behavior
image

## To Reproduce
Add a bookmark, go to the bookmark list (`!bookmarks`) and edit the last one.

lubiedo:


PalmerAL:
I added the margin to the bottom only in b5ff837b92d869ba10ee818e41449a808f3ccb74. That fixes the issue, and having no top margin makes the tab title stay in the same place when the bookmark editor opens, which makes the transition between the two states look slightly nicer.

(That commit also includes a small fix for an issue that was making a scrollbar show up on the bookmark manager unnecessarily).

PR #1694, opened 2021-07-25
Fix mac titlebar layout for RTL languages

### LINKED ISSUE #1688: Broken layout on macOS with RTL language
Found this while reviewing #1683. on macOS, when the system language is RTL, the OS moves the window controls to the right, which breaks Min's layout:

* I don't know of a way to directly check whether the OS is using an RTL layout, but we could make a list of RTL languages and use that.
* This shouldn't be an issue in windows and Linux, since we emulate our own window controsl there.
* This doesn't cover having a true RTL mode in the UI; we should probably do that too, but I expect that to be a lot more work.

PalmerAL:
Fixes #1688.

This still isn't perfect (and the layout isn't really RTL), but it's at least usable now.

### DIFF
--- a/localization/localizationHelpers.js
+++ b/localization/localizationHelpers.js
@@ -55,6 +55,10 @@ set the value attribute for all elements with a [data-value] attribute
  */
 
 if (typeof document !== 'undefined') {
+  if (languages[getCurrentLanguage()] && languages[getCurrentLanguage()].rtl) {
+    document.body.classList.add('rtl')
+  }
+
   document.querySelectorAll('[data-string]').forEach(function (el) {
     var str = l(el.getAttribute('data-string'))
     if (typeof str === 'string') {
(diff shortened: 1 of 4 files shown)


localization/localizationHelpers.js
```diff
@@ -55,6 +55,10 @@ set the value attribute for all elements with a [data-value] attribute
  */
 
 if (typeof document !== 'undefined') {
+  if (languages[getCurrentLanguage()] && languages[getCurrentLanguage()].rtl) {
```
lubiedo, line 58: Just a small detail but why not use `userLanguage` defined at line 32 instead of calling `getCurrentLanguage()` twice?
```suggestion
  if (languages[userLanguage] && languages[userLanguage].rtl) {
```

PalmerAL: `userLanguage` is null initially, and is only set when l() is called for the first time. The reason for that, IIRC, is that in the main process, getCurrentLanguage uses app.getLocale, which only works after Electron’s `ready` event. But this code here only runs in the renderer process, so getCurrentLanguage will use the `navigator.language` method, which works immediately.

lubiedo (approved):
Tested. Looks good!

Issue #1180, opened 2020-08-03
In-browser page translation


rajktech:
can we do translation of pages from google like we do in chrome.

PalmerAL:
There's an OSS library for doing client-side translations in JS now: https://github.com/browsermt/bergamot-translator

Could be very useful for this.

shalva97:
wow, if that gets added I can live inside Min 99% times.... last 1% would be screen sharing :D

Milad-Laly:
Ah thank you @PalmerAL , I'll look into this again. I couldn't find a fix during that time. I'll explain on discord later tonight after I finish being Oncall

PalmerAL:
Something I didn't realize is that Bergamot currently supports only a limited set of languages: https://github.com/mozilla-applied-ml/bergamot-models#currently-supported-languages. They'll hopefully expand to more eventually, but for now it isn't enough to be useful.


Another interesting option is LibreTranslate: https://github.com/LibreTranslate/LibreTranslate, which is essentially a self-hosted version of Google Translate with an API that supports a much wider set of languages. I imagine I could set up a cheap VPS running this, and then use that in Min.

The downside is that the page contents would have to be sent to the server to be run through the translation engine. I obviously wouldn't store any of the text uploaded or try to identify people through it, but it requires that people trust me to do that.

WDYT?

PalmerAL:
I ended up writing a userscript based on LibreTranslate: https://gist.github.com/PalmerAL/ed228d059002e3759bc89d59a22dfb42

If anyone's willing to test this out and see what the translation quality is like, that would be great. It would be fairly simple to host a translation instance somewhere and integrate this into the browser, but I want to confirm that the translations are good first.

The userscript installs like any other script, and there's a section at the top to set what language it translates pages into.

At the moment, it's slow to do the translation and doesn't work on some pages, but I think that should be fixable.

shalva97:
it looks promising. but there are lots of texts that are not translated and sometimes nothing happens when clicking "transalte page". sometimes this error shows up: `Uncaught SyntaxError: Identifier 'TRANSLATE_TO_LANG' has already been declared`

quality seems okay, tested on a russian site
Min:
![image](https://user-images.githubusercontent.com/22417494/125316094-d542b000-e348-11eb-8953-e6b99c242a74.png)
Edge:
![image](https://user-images.githubusercontent.com/22417494/125316056-cc51de80-e348-11eb-958e-01be5315c26c.png)
Chrome: 
[...]

PR #1701, opened 2021-08-01
Add in-browser translation

PalmerAL:
This PR adds an in-browser page translation feature using [LibreTranslate](https://github.com/LibreTranslate/LibreTranslate), an open-source translation engine, available in the page context menu.

This is discussed more in #1180, but the intent is to include this as an experimental feature (with a "beta" label in the UI), and see how many people use it. If it gets significant usage, there's options for improving the translation quality, running LibreTranslate on better hosting to improve the translation speed, etc.


PalmerAL:
I'm going to make a new release this week, so I'll probably merge this in a few days, but if someone wants to review it before then that would be great.

PR #1712, opened 2021-08-13
Calculation plugin

lubiedo:
More information and discussion on thread #1682

image
image

A minor change also goes into this PR: Places right now show no icon and that might be confusing. Commit 2e849be adds a default icon. It will be shown as follow:

image


PalmerAL:
I haven't had time to fully review this yet (I got sidetracked by the Google issues), but initializing mathJS is taking 500ms on startup:

Screen Shot 2021-08-16 at 1 58 18 PM

Which more than doubles the time needed to load the UI. I'm not sure what the best way to fix that is, but we probably need to figure out some way to reduce the performance impact.

Milo123459:
MathJS is chunky. Could probably find an alternative with WASM that'd load considerably faster

lubiedo:
Agree. mathjs was an overkill and replaced with [expr-eval](https://github.com/silentmatt/expr-eval), which is a much more simple package for mathematical expressions. This also discards the possibility of conversion (I tried using DDG API's conversion but looks out-of-service for JSON requests for some reason).

js/searchbar/calculatorPlugin.js
```diff
@@ -0,0 +1,50 @@
...
+  searchbarPlugins.reset('calculatorPlugin')
+  var result
+
+  try {
+    result = math.evaluate(text)
+  } catch (e) { return }
+
+  searchbarPlugins.addResult('calculatorPlugin', {
+    icon: 'carbon:calculator',
+    title: `${result.toString()}`
```
PalmerAL, line 21: Maybe `descriptionBlock: 'Click to Copy'` here? Most people won't hover over the search result, so that would make it clearer that the option exists. And then when you click on it, the text could change to "Copied" to provide feedback.

js/searchbar/calculatorPlugin.js
```diff
@@ -0,0 +1,50 @@
...
+      clipboard.writeText(e.target.innerText)
+    })
+  }
+}
+
+function initialize () {
+  searchbarPlugins.register('calculatorPlugin', {
+    index: 1,
+    trigger: function (text) {
+      if (text.length < 3 || /^[\d\s.]+$/.test(text)) {
```
PalmerAL, line 39: I don't understand what the regex here is doing - can you add a comment?

PalmerAL:
That's a lot better! I didn't see any kind of security guide for expr-eval; I'm assuming it's safe to use, but it would be good to check.

A conversion plugin would be a nice feature to add in the future, but that could be a separate plugin - there's a lot of other packages available to do that.

js/searchbar/calculatorPlugin.js
lubiedo: Avoids that inputs like `1234` or `  1234 5678` are processed.

js/searchbar/calculatorPlugin.js
lubiedo: That sounds like a really good idea, I will implement it.

lubiedo:
Commit cddd273 moves 'Click to copy' legend to the description block and gives feedback to the user after copying.
Commit 32af9c9 makes sure that some checks are done before parsing and evaluating expressions in order to avoid any security issue.

@PalmerAL: let me know what you think.

PalmerAL:
This looks pretty good to me. I made a couple minor fixes in f2153bca39d20ca43a0407f5884b8f1b8c1a720e - the instantAnswerPlugin change fixes a preexisting issue where if you searched for sqrt(2), a wikipedia article would be shown at the top instead of the calculation result.

There's a couple weird things I noticed about expr-eval: it doesn't do multiplication without a *, so `cos(2pi)` doesn't work, and it has issues with floating points - `tan(pi)` returns `-1.2246467991473532e-16` instead of 0. But I think those probably won't be issues too often, and I don't see any easy fix, so I'm fine with leaving it as-is.

Unless you see anything else, I think this is good to merge.

lubiedo:
Thanks for the fixes :) 
Yes, unlike maths, expr-eval doesn't recognize a multiplication without an asterisk. I think it has to do with the tokenization and that's why the expressions have to be there. Then, the floating point is indeed weird but I think it comes directly from JavaScript itself:
```js
> Math.tan(Math.PI)
< -1.2246467991473532e-16
```

I also believe that this can be merged if there's no other changes to do.

PalmerAL:
I think everything looks good!

Issue #868, opened 2019-12-04
Can't Sign Into Gmail


Bundito14:
I would love to use this browser as an everyday alternative, however, it won't let me sign into my Google account due to "security reasons." Is there a workaround to this?

<hr>

(from @PalmerAL):

For anyone else having this issue, please post the following:
1. Do you have 2FA enabled?
2. OS
3. Approximate location (if you're willing to share)

If you're running the latest version of Min, try enabling the "use a custom user agent" in the Min preferences, and pasting the following in the user agent field:
```
Mozilla/5.0 (Windows NT 10.0; WOW64; rv:74.0) Gecko/20100101 Firefox/74.0
```
Once you do this, you'll need to restart Min before signing in.

Please let us know whether this works for you.

PalmerAL:
Can you post a screenshot of the exact error you're getting? Multiple people have reported this in the past few weeks ([1](https://github.com/minbrowser/min/issues/863), [2](https://twitter.com/MaikoWapstra/status/1201476629593251841)), but I'm able to sign in without any issues, so I'm not sure why this is happening.

lee-yunseok:
In my case I can login my first account but after this I can't add any another accounts.

```
Couldn't sign you in

You are trying to sign in from a browser or app that doesn't allow us to keep your account secure.

Try using a different browser. Learn more
```

PalmerAL:
This looks to be the same as this issue unfortunately: https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/issues/3545. I've also tried this with two accounts, and I can log in to both simultaneously, so it appears to only be applied to some subset of people.

@lee-yunseok Assuming you're running min with electron 7 (which you should be if you used the default version from npm), can you try this?

* Go [here](https://duckduckgo.com/?q=user+agent) and find your current user agent.
* Replace the Chromium version string in the user agent with "78.0.3904.97" or "78.0.3904.108".
* In [UASwitcher.js](https://github.com/minbrowser/min/blob/master/main/UASwitcher.js), set `newUserAgent` to the string from step 2.

lee-yunseok:
@PalmerAL Where is the Chromium version string?

PalmerAL:
@lee-yunseok It's the part of the string that starts with "chrome", for example:

[...]

PR #1715, opened 2021-08-15
Implement partial UA client hints support for gmail

PalmerAL:
Starting a few days ago, Gmail now checks (for unknown reasons) for `navigator.userAgentData`. If it isn't found, it redirects to the basic HTML version.

Client hints currently doesn't work in Electron: https://github.com/electron/electron/issues/30201. So this PR adds a (somewhat ridiculous) workaround of creating a fake `userAgentData` object that mostly conforms to the one Chrome provides. With that, the standard version of Gmail seems to load again.

Separately, signing in to Gmail is broken, see https://github.com/minbrowser/min/issues/868. You can workaround that by setting your user-agent to Firefox in preferences, but then you'll have to switch back to the standard user-agent after signing in before trying this fix.

I wish Google would stop trying to block non-official Chromium browsers...

### DIFF
--- a/js/preload/siteUnbreak.js
+++ b/js/preload/siteUnbreak.js
@@ -96,6 +96,29 @@ if (window.location.hostname === 'calendar.google.com') {
   `)
 }
 
+/* Gmail - required for loading standard version (otherwise redirects to basic HTML) */
+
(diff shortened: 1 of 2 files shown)


lubiedo (approved):
Looks good!

PR #1716, opened 2021-08-15
Use Firefox UA for google accounts

PalmerAL:
This PR switches Google sign in to use a Firefox UA rather than an edge one. To avoid having to update the UA on every Firefox release, it uses the current date to guess what the latest Firefox version is.

#868


PalmerAL:
Yeah, I think so. Ideally we'd figure out how Google is detecting Min as being different from standard Chrome builds (there shouldn't be any differences!), but I'm totally stumped at this point.

Issue #1720, opened 2021-08-19
when downloaded file name is changed, it doesnot update in downloads panel


shalva97:
- **Min Version:** 1.21.0
- **Operating system:** Windows 10

## Expected Behavior
it should show the name I write in save dialog

## Actual Behavior
it shows old file name

## To Reproduce
download some file, change name on save dialog, and it will show old name in the panel

lubiedo:
Thanks @shalva97. I was able to reproduce the issue and created PR https://github.com/minbrowser/min/pull/1721 for this one.

lubiedo:
Issue was fixed with #1721.

PR #1721, opened 2021-08-19
Fix issue #1720

lubiedo:
Check if the item filename differs from the basename in path and update
the name in the download bar.

### DIFF
--- a/main/download.js
+++ b/main/download.js
@@ -31,12 +31,19 @@ function downloadHandler (event, item, webContents) {
     })
 
     item.on('updated', function (e, state) {
+      var filename = item.getFilename()
+      const savePathFilename = path.basename(item.getSavePath())
+
+      if (filename !== savePathFilename) {
+        filename = savePathFilename
+      }
+
       if (item.getSavePath()) {
         currrentDownloadItems[item.getSavePath()] = item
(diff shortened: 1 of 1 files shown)


PalmerAL:
I think the filename calculation should happen once at the beginning of the `else` block, and then be used for both IPC calls. The call on line 26 makes the item appear initially, and then `item.on('updated')` happens every time the item's download progress changes (once or twice a second).

Also, we don't need to call `item.getFilename()` at all any more, right? It could just be `name: savePathFilename`.

BTW, your last few commits don't have an email address associated with them - if you set it with `git config`, GitHub will associate them with your account.

lubiedo:
The thing is that, for what I've tested, `item.getSavePath()` is not really defined until the user selects a path or changes the name and falls into the `updated` event.

PalmerAL:
Oh, right - I was thinking that the file starts downloading after the save dialog is closed, but the download actually starts before that. This is the right fix then.

GitHub still isn't linking your commits - there's a missing `.com` at the end of the email:


But it doesn't matter too much for this, so I'm going to go ahead and merge.

lubiedo:
Oh, dammit! Will fix that

PR #1736, opened 2021-09-10
Update tr.json

eneskutlay:


localization/languages/tr.json
```diff
@@ -267,11 +267,11 @@
     "passwordCaptureSavePassword": "% için parola kaydedilsin mi?",
     "passwordCaptureSave": "Kaydet",
     "passwordCaptureDontSave": "Kaydetme",
-    "passwordCaptureNeverSave": null, //missing translation
+    "passwordCaptureNeverSave": "% için parolaları asla kaydetme",
```
PalmerAL, line 270: The English version of this is just "never save": https://github.com/minbrowser/min/blob/master/localization/languages/en-US.json#L270

eneskutlay: Yes, this is locatilazed version. It is being used like that in Turkish.

PalmerAL: I just want to make sure the Turkish version has the same meaning as the English one. The Turkish one here has a % sign in it, and Google Translate says it means "Never save passwords for %". the original English is just "Never save", without "passwords" or a site name.

It's used here:
Screen Shot 2021-09-11 at 6 10 45 PM

So it should be fairly short.

eneskutlay: I'm sorry, I didn't know I shouldn't have used the site name. I'll shorten it, edit it, and update it again

localization/languages/tr.json
```diff
@@ -267,11 +267,11 @@
     "passwordCaptureSavePassword": "% için parola kaydedilsin mi?",
     "passwordCaptureSave": "Kaydet",
     "passwordCaptureDontSave": "Kaydetme",
-    "passwordCaptureNeverSave": null, //missing translation
+    "passwordCaptureNeverSave": "Asla kaydetme", 
     /* Password viewer */
     "savedPasswordsHeading": "Kayıtlı Parolalar",
     "savedPasswordsEmpty": "Kaydedilmiş parola yok.",
-    "savedPasswordsNeverSavedLabel": null, //missing translation
+    "savedPasswordsNeverSavedLabel": "Asla kayıt edilmeyecek parolalar",
```
PalmerAL, line 274: I think this might also need to be shorter - the English is just "never saved". It's used in settings > view saved passwords:

Screen Shot 2021-09-12 at 10 38 23 PM

Sorry for not catching that earlier.

eneskutlay: Sentences can get longer while placing them. I'll update it with a better alternative. A direct translation of this sentence would not make sense. I have to do some localization.

PalmerAL:
I think everything looks good, thank you!

localization/languages/tr.json
```diff
@@ -281,8 +281,7 @@
     "email": "E-Posta",
     "password": "Parola",
     "secretKey": "Gizli anahtar",
-    "openExternalApp": null, //missing translation
-    "clickToCopy": null, //missing translation
-    "copied": null //missing translation
-  }
```
yurikhan, line 287: Looks like you lost this brace.

PR #1745, opened 2021-09-20
Fix tr.json

Hakerh400:
PR https://github.com/minbrowser/min/pull/1736 accidentally deleted closed brace.


PalmerAL:
Thanks, should've noticed that.

PR #1749, opened 2021-09-20
Improve error handling for invalid JSON

Hakerh400:
This PR fixes several issues:

* `decomment(data)` used to be called twice in case of an error. Since it is called at least once, we assign the result to a variable
* `String.prototype.match` is ~15 times faster than `RegExp.prototype.exec` (and is generally recommended to use `match` if only one match is needed, like in this case)
* There is no need for the `g` flag on the regex
* In case of missing closed brace (like it was the case with `tr.json`), the entire build would fail, because `exec(e)[1]` was not checking for `null` and the actual error message was `Unexpected end of JSON input`
* `exec(e)` was converting the entire error (including the stack trace) to a string. It might happen that somehow `at position` appears somewhere in the stack trace, while the error message is unrelated, leading to misleading error display info. Now we match only on the error message

### DIFF
--- a/scripts/buildLocalization.js
+++ b/scripts/buildLocalization.js
@@ -22,13 +22,25 @@ function buildLocalization () {
     const data = fs.readFileSync(path.join(languageFileDir, file), 'utf-8')
 
     let obj
+    let decommented = null
+
     try {
-      obj = JSON.parse(decomment(data))
+      decommented = decomment(data)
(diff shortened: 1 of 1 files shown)


PalmerAL:
LGTM, thanks.

I'm not too worried about performance here, because this only runs at build time (and most of it only runs when there's an error). But there's no harm in improving it either. The bug fixes look good.

PR #1758, opened 2021-10-01
Add search input to task overlay

PalmerAL:
This PR adds a search box to the top of the task overlay to search through open tabs. Pressing enter from within the search switches to the first result.

I'm not totally happy with the way the search input looks; if anyone has ideas for how to make it better, let me know. The first search result should also probably be highlighted so that it's clear you can select it with enter.

### DIFF
--- a/js/taskOverlay/taskOverlay.js
+++ b/js/taskOverlay/taskOverlay.js
@@ -119,6 +119,20 @@ var taskOverlay = {
     this.isShown = true
     taskSwitcherButton.classList.add('active')
 
+    taskOverlay.render()
+
+    // un-hide the overlay
+    this.overlayElement.hidden = false
+
+    // scroll to the selected element and focus it
+    var currentTabElement = document.querySelector('.task-tab-item[data-tab="{id}"]'.replace('{id}', tasks.getSelected().tabs.getSelected()))
+
+    if (currentTabElement) {
+      currentTabElement.classList.add('fakefocus')
+      currentTabElement.focus()
+    }
+  },
+  render: function () {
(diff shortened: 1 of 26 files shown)


js/taskOverlay/taskOverlay.js
```diff
@@ -216,7 +218,60 @@ var taskOverlay = {
...
+      }
+
+      tasks.forEach(function (task) {
+        var taskContainer = document.querySelector(`.task-container[data-task="${task.id}"]`)
+
+        var tabMatches = 0
+        task.tabs.forEach(function (tab) {
+          var tabContainer = document.querySelector(`.task-tab-item[data-tab="${tab.id}"]`)
+
+          var searchText = (tab.title + ' ' + tab.url).toLowerCase()
```
lubiedo, line 244: There could be also a search by task name:
```js
          const taskName = taskContainer.querySelector('input.task-name').getAttribute('placeholder')
          var tabContainer = document.querySelector(`.task-tab-item[data-tab="${tab.id}"]`)

          var searchText = (`${taskName} ${tab.title} ${tab.url}`).toLowerCase()
```
What do you think?

lubiedo (approved):

js/taskOverlay/taskOverlay.js
PalmerAL: Yeah good point, I did something like this in cbddc1fcaf6abb6feefef458593c3c10d855cf56. Compared to your solution, it won't match the default task names (like "task 1"), but I think it's unlikely that anyone would search for that.

I also changed it so that for a multi-word search, the words are allowed to be in any order within the tab / task name.

PR #1764, opened 2021-10-14
Add option to follow system darkmode

rost:
Adds the option to let the darkmode setting rely on the underlying OS darkmode.

Only tested on Mac.

---

I'm also not familiar with Electron and how these apps are usually structured. I read through the Min architecture document on the wiki but could only get this to work by grabbing the OS setting from the main thread.

### DIFF
--- a/js/util/theme.js
+++ b/js/util/theme.js
@@ -2,6 +2,10 @@ if (typeof require !== 'undefined') {
   var settings = require('util/settings/settings.js')
 }
 
+function systemShouldEnableDarkMode () {
+  return settings.list.systemShouldUseDarkColors
+}
+
 function shouldEnableDarkMode () {
   var hours = new Date().getHours()
   return (hours > 21 || hours < 6)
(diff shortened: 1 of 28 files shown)


PalmerAL:
I haven't had a chance to test this yet, but it looks pretty good. I think the architecture is fine - since you can only get the theme info from the main process, a setting is probably the easiest way to sync that everywhere across the app.

There's a nativeTheme `updated` event that we can use instead of setInterval: https://www.electronjs.org/docs/latest/api/native-theme#event-updated.

rost:
> nativeTheme `updated` event

Switched to listen to that instead. Much cleaner.

rost:
In general, I'm fairly certain the logic was crappy. I cargo culted my way through the switching logic until I saw the behaviour I thought was correct. 

As for 1 and 3, those are the only things I tested, and I never saw an issue.
As for 2 and 4, I was not aware of the possibility of using a settings.json file, I'll definitely try that.

What specific issues did you see that I might look for?

Issue #948, opened 2020-03-29
Rearrange tabs idea


956MB:
Referencing #684. 
I don't know if this sort of thing has been suggested or considered before, but I thought Id put an idea for this out there. Being that the task rearranging thing is less than perfect, maybe this could be one possible solution to tab rearranging.

![8dd61e748c85d6db780170202a0cffe2 (1)](https://user-images.githubusercontent.com/39519792/77843771-00aa8700-7166-11ea-8a61-21d31119baec.png)

With the red in this screenshot being the draggable area (on macOS at least), the small blue dots in the left of each tab could be the spot to drag tabs around. With them only being visible in the tab on hover of course. They could be small dots like in this, a thin line, or like a small right angle arrow facing the top left of the tab. Really anything that fits the minimal aesthetic of Min.

Syndamia:
I'm not sure how well this will work. The thing is that when clicking on a tab you open the `search`. If it's a small thing on the side, accidental openings of `search` and of moving tabs will inveitably happen and if it's big, then it will clutter the UI.

Also the drag spots being visibe only on hover will be less than ideal, because every time you hover over a tab it will add the button and move the text to the right (*I can see people, myself included, being annoyed by that*). Another options is for it to be just a blank space that fills in with the icon, but then you have a (*relatively*) big blank space most of the time.

Another thing to consider is such a feature will probably require a fair amount of rework of tab functionality, maybe even needing a rewrite of most of it (*I don't know how it works exactly, @PalmerAL should chime in*) which might not be worth it.

Bloumbs:
@Syndamia  Addressing the points you gave:

Opening search:
I think a resonable work around for this would be just a double click action to trigger search. So you can navigate in and out of tabs with the normal one click, drag around around your window on top of any of your tabs, not worrying about opening search, then a real quick double tap to open search for a tab. Or also a keybind to open search as well, like "_key_+mouseClick" on the tab.

The drag spots:
I dont think what you said about text being moved would really be an issue. The tab already has padding on the left for the text, so theres already room for it I think. I think it could be like the screenshot in the initial post above, the blue dot being the drag point doesnt need to be that large, just enough to be noticable, and the encapsulating div or whatever would be placed just to the left of the tabs text, filing vertically to the top/bottom of the tab. So there wouldnt be any reason for the drag area/dot to move the tabs text in any way, the dots color or whatever would just become visible on hover of the entire tab. And of course it could be a thin line or something instead if that looked better or took up even less space.

[...]

PR #1765, opened 2021-10-14
Move tabs in tabbar

rost:
Adds functionality and keybindings to move a tab in the tab bar.

I see this as a proposal for a temporary solution until tab rearrangement in the task view is thought through and somehow fixed on Mac as per described in https://github.com/minbrowser/min/issues/948

It is propsed there to perhaps hide the tab movement functionality behind a modifier key and mouseclick.

I thought that since we already have `option+mod+left/right` to move *between* tabs, perhaps adding `shift` for `shift+option+mod+left/right` could be used for moving the actual tab.

---

Only tested on Mac.


PalmerAL:
I removed the last of the tabBar methods in 77264c3761f3626f6f8c79ed68d24b2329b777ea. Other than that, LGTM, thanks!

PR #1781, opened 2021-10-30
Registry patch

nohopestage:
Fixed Min not showing in the Default Programs list, added support for .pdf and .url file extensions.

### DIFF
--- a/main/registryConfig.js
+++ b/main/registryConfig.js
@@ -3,32 +3,32 @@ var regedit = require('regedit')
 var installPath = process.execPath
 
 var keysToCreate = [
-  'HKCU\\Software\\Classes\\Min',
-  'HKCU\\Software\\Classes\\Min\\Application',
-  'HKCU\\Software\\Classes\\Min\\DefaulIcon',
-  'HKCU\\Software\\Classes\\Min\\shell\\open\\command',
-  'HKCU\\Software\\Clients\\StartMenuInternet\\Min\\Capabilities\\FileAssociations',
-  'HKCU\\Software\\Clients\\StartMenuInternet\\Min\\Capabilities\\StartMenu',
-  'HKCU\\Software\\Clients\\StartMenuInternet\\Min\\Capabilities\\URLAssociations',
-  'HKCU\\Software\\Clients\\StartMenuInternet\\Min\\DefaultIcon',
-  'HKCU\\Software\\Clients\\StartMenuInternet\\Min\\InstallInfo',
-  'HKCU\\Software\\Clients\\StartMenuInternet\\Min\\shell\\open\\command'
+  'HKCR\\Min',
+  'HKCR\\Min\\Application',
+  'HKCR\\Min\\DefaulIcon',
+  'HKCR\\Min\\shell\\open\\command',
+  'HKLM\\SOFTWARE\\Clients\\StartMenuInternet\\Min\\Capabilities\\FileAssociations',
+  'HKLM\\SOFTWARE\\Clients\\StartMenuInternet\\Min\\Capabilities\\StartMenu',
+  'HKLM\\SOFTWARE\\Clients\\StartMenuInternet\\Min\\Capabilities\\URLAssociations',
+  'HKLM\\SOFTWARE\\Clients\\StartMenuInternet\\Min\\DefaultIcon',
+  'HKLM\\SOFTWARE\\Clients\\StartMenuInternet\\Min\\InstallInfo',
+  'HKLM\\SOFTWARE\\Clients\\StartMenuInternet\\Min\\shell\\open\\command'
 ]
 
 var registryConfig = {
-  'HKCU\\Software\\RegisteredApplications': {
+  'HKLM\\SOFTWARE\\RegisteredApplications': {
     Min: {
-      value: 'Software\\Clients\\StartMenuInternet\\Min\\Capabilities',
+      value: 'SOFTWARE\\Clients\\StartMenuInternet\\Min\\Capabilities',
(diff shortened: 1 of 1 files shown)


PalmerAL:
This seems fine; I'll try to test it in a few days. Since I don't know much about the registry - what's the difference between HKCU / HKCR / HKLM?

Syndamia:
Correct me if I'm wrong, but I believe:
- HKEY_CURRENT_USER (HKCU) is for user-specific configurations
- HKEY_LOCAL_MACHINE (HKLM) is for global configurations
- HKEY_CLASSES_ROOT (HKCR) is for filename extension information (and some other stuff)

nohopestage:
> Correct me if I'm wrong, but I believe:
> 
> * HKEY_CURRENT_USER (HKCU) is for user-specific configurations
> * HKEY_LOCAL_MACHINE (HKLM) is for global configurations
> * HKEY_CLASSES_ROOT (HKCR) is for filename extension information (and some other stuff)

You are right. I don't really know why it didn't work with HKCU though, I just noticed all the other browsers have their keys in HKLM.

nohopestage:
Apparently, the original problem has nothing to do with HKCU/HKLM. For some reason the installer creates empty values under HKCU/SOFTWARE/Classes/Min in its subkeys. Specifically, an empty value in ../Min/shell/open/command causes the problem. At least setting it manually fixed the issue. Not sure if that's the case for Windows 7 though, as I remember setting the value manually with no result. I'll test it once again, please don't merge for now.

nohopestage:
Seems like Windows 7 requires the key to be under HKLM:
> The default installation of Windows does not register a per-user default Internet or email program, only a system-wide default. This provides a smooth upgrade path from previous versions of the operating system, in which only the HKEY_LOCAL_MACHINE subtree is supported for client registrations.
Regarding the empty values issue: some values in the array were named without single quotes, which caused the problem. Everything should be fine now.

PalmerAL:
This isn't working for me so far (Windows 10 20H2). It looks like the `HKCR\Min` keys are being set, but none of the `HKLM\SOFTWARE\Clients\StartMenuInternet` keys are showing up in the registry editor. I can still set Min as the default browser (although that may be from registry keys left over from an earlier installation), but I can't set it as the default for PDF.

Will try debugging in a bit, but let me know if you have any ideas.

nohopestage:
That is weird, I've tried running the script from node command line and it was working fine. Is the value under RegisteredApplications being created? Does the Min key have all its subkeys and values? Is the Min value under .pdf/OpenWithProgIds there? You may need to manually set Min as the default for .pdf in Default apps > Set defaults by app

PalmerAL:
I tried running the script directly from the node CLI (good idea!), and if I modify it to print out errors from the regedit functions, I get this:
Screen Shot 2021-11-14 at 7 26 19 PM
If I rerun as an administrator, I don't get any errors, and the keys show up in the registry editor. So I think the issue is that writing to HKCR / HKLM requires administrator access, but when the installer launches an instance of Min to write to the registry, it's not running as administrator. I'm not exactly sure how to fix this, but [this issue](https://github.com/electron/electron-packager/issues/647) suggests using https://www.npmjs.com/package/sudo-prompt, which looks like a decent solution.

Although when you were running it with Node, were you able to make it run successfully without administrator access?

nohopestage:
Ah, that makes sense. Sorry, writing to HKCR / HKLM does require administrator privileges, which I forgot to mention. There seems to be an easier [solution](https://stackoverflow.com/a/47654414) though.

PalmerAL:
I haven't tried it, but that looks like it would make Min run as administrator all the time, rather than just during the installation (which is what we want).

nohopestage:
Oh, actually yes, but isn't it also the case for the solution you mentioned? Apparently, there's no way to require a Squirrel installer to run with admin rights.

PalmerAL:
`sudo-prompt` executes a terminal command as root; I was thinking that we could run `registryConfig.js` in its own subprocess (using the electron binary and [electron_run_as_node](https://www.electronjs.org/docs/latest/api/environment-variables#electron_run_as_node)), and then we would only need to prompt when we want to update the registry during install.

The installer isn't running the script - it's launching an instance of Min, which then checks for a flag and runs the registry config if it's present: https://github.com/minbrowser/min/blob/master/main/main.js#L40-L42. So even if the installer ran as admin it wouldn't help.

nohopestage:
That makes sense. `sudo-prompt` is the way to go.

PalmerAL:
I think 588a0cf7a3baf629b2912f72bf33d7584f9f6502 is close to the right way to do it. If I run it directly with `node_modules\.bin\electron . --squirrel-install` it works, but if I build the installer and run that, the browser hangs on startup. I'm not sure why yet; maybe you'll see the issue.

nohopestage:
Oh, I wish I could. I'm not familiar with JS, so I see nothing wrong there. You can send me the build to see if I can reproduce.

PalmerAL:
I finally came back to this PR; I think the hang I was seeing before was actually a graphics issue with my VM. If I change the shortcut to Min to run with `--disable-gpu` it seems to be working now. Or maybe something else changed in the meantime 🤷 

I can set Min as the default handler for PDF files now, but if I try to open one, I get stuck in this loop:

https://user-images.githubusercontent.com/10314059/149641889-56dea9c9-a9b9-479a-a558-d4333dff5bd4.mov

nohopestage:
I was not able to reproduce this bug, it might have something to do with your VM. Could you please send me the installer, so I could try to reproduce it that way? I've only tried to run registryConfig.js in node CLI.

PR #1799, opened 2021-11-18
Update ar.json

hatembk:


PalmerAL:
Thanks! There's some other translations that are missing still - if you want to translate those, you can find them by searching for "null" in the file.

Issue #1394, opened 2020-12-20
Bitwarden Integration not Working


gpioneer90:
- **Min Version: 1.17.3**
- **Operating system:**
Windows 10

## Expected Behavior
When clicking the key icon and inputting BW password, to login to website/open interface
## Actual Behavior
Asks for password over and over, as if it won't accept. No error message. Just click OK after inputting on the password window, and a new one pops up. Only way to now close this window is to leave password blank and click OK. And then BW doesn't pass login info to website.
## To Reproduce
Try to log in to a website using the BW module, input password->login loop

more details: when I tried installing the CLI tool by dragging it into the prompt, it was stuck at "installing" for a very long time for my system (30-45 seconds, nvme ssd, nothing really every takes longer than a second on the system). I clicked outside the UI and it closed the drag and drop install window and Bitwarden was selected. When I went to try to log in to a website the first time, it would accept my password but keep crashing the browser immediately afterwards. I then chose the no password manager option and then Bitwarden again in the settings. Afterwards, when trying to log in to websites, it gets stuck in login loop as mentioned above. I don't know how to "uninstall" the BW module now to try to reinstall it, which may or may not fix the issue.

gpioneer90:
I have 2FA enabled on Bitwarden. Maybe that's why it doesn't work in Min? On Firefox on initial setup of BW it will request I enter a code sent to my email

PalmerAL:
2FA isn't supported for now, which is #929.

My guess is that trying to sign in the first time left Bitwarden in an unexpected state; if you're comfortable disabling 2FA, you can do this:
1. Quit Min.
2. Go to `\Users\Username\AppData\Roaming\Min\tools`
2. Delete "bw.exe"
3. Restart Min.

And install it again.

JackGFuller:
I have encountered a similar problem. And I did the process suggested by PalmerAL. When I log in to a website, the 'key' icon is shown in the UserName field, but it isn't filled in by Bitwarden. 

![image](https://user-images.githubusercontent.com/74563788/109093872-f1152480-76cd-11eb-9025-6a1f6c57246f.png)

Clicking on the 'key' icon results in no action.

PanoptesDreams:
Also experiencing problem, unable to resolve.

Did a full reset of the browser and tried to re-add BW CLI, now I can't even add the CLI back to Min

Min 1.22.0
Win 10 1703 & 21H1

aiyagari:
[...]

Issue #929, opened 2020-03-09
Bitwarden sign in fails if 2FA is enabled


anonymous1184:
For the miriad of sites in the password manager, I only use a few with more than one account.

Is there any plans for selecting the desired credentials? I belive every internet user has the need to use at very least a service with more than one set of credentials (e.g. work/personal email).

PalmerAL:
You should see a list that allows you to pick which credentials you want to use, like this: https://user-images.githubusercontent.com/10314059/73512567-b8622900-43af-11ea-9880-5f72efdc7b33.png. Are you not seeing that? Or do you mean something else?

anonymous1184:
I was not seeing that.

There's no way of setting up 2FA on the initial configuration of Bitwarden in Min. The issue arised after disabling 2FA, setting up Min and subsecuently enabling 2FA. It was solved by manually re-logging while Min was closed (re-syncing by restarting Min didn't work):

```
bw logout
bw login email password --method 3 --code <OTP>
```
(Refer to [online documentation](https://help.bitwarden.com/article/cli/#enums) for `--method` values, since CLI help doesn't cover them.)

On a side note, its worthy to remember that [match detection](https://help.bitwarden.com/article/uri-match-detection) rules applies, because even after being able to choose credentials one of them didn't appear (this was totally a layer 8 issue).

PalmerAL:
Thanks for investigating more; I guess it would make sense to add some options to the sign in form for that.

gpioneer90:
> 
> 
> I was not seeing that.
> 
> There's no way of setting up 2FA on the initial configuration of Bitwarden in Min. The issue arised after disabling 2FA, setting up Min and subsecuently enabling 2FA. It was solved by manually re-logging while Min was closed (re-syncing by restarting Min didn't work):
> 
> ```
> bw logout
> bw login email password --method 3 --code <OTP>
> ```
> 
> (Refer to [online documentation](https://help.bitwarden.com/article/cli/#enums) for `--method` values, since CLI help doesn't cover them.)
> 
> On a side note, its worthy to remember that [match detection](https://help.bitwarden.com/article/uri-match-detection) rules applies, because even after being able to choose credentials one of them didn't appear (this was totally a layer 8 issue).

Hi anon,
[...]

PR #1809, opened 2021-11-27
Improve Bitwarden integration

### LINKED ISSUE #1689: Password Manager Issue
Hello...Hope all is well in your world..Using Fedora 34...I just reinstalled Min after not having used it in a while...I use the Bitwarden  password manager and after  installing it  have found it is not working...I can install Bitwarden and sign in, but the keyboard shortcut or the right click in the password area does nothing...Properly signed in and having used it all before, something seems to be amiss...Thanks...Peace

### LINKED ISSUE #1796: Bitdefender
I selected Bitwarden for the password app, I followed the install instructions, it seemed to hang, and now whenever I try to use Bitwarden, it repeatedly askes for my password. Worse, even though I uninstall and reinstall Min, it somehow keeps my Bitwarden selection, so I can't try to reinstall Bitwarden.

Bob

PalmerAL:
This PR attempts to fix a few of the most common issues with the Bitwarden integration:

* Bitwarden recently started requiring an API key instead of a password in some cases: https://bitwarden.com/help/article/cli-auth-challenges/, which I believe is the cause of the recent issues about Bitwarden setup failing. This PR updates the installation flow to ask for a client ID and key instead of a username and password. This also makes it possible to connect Bitwarden accounts that have 2FA enabled.

Fixes #1796, #1394, #929

* Added a re-prompt for authentication if the user is signed out and tries to autofill a password. Previously, the authentication prompt would never show after setup was complete, which meant it was possible to end up in a state where you weren't signed in and there was no way to sign in again in the UI. This should hopefully fix #1689.

### DIFF
--- a/js/passwordManager/bitwarden.js
+++ b/js/passwordManager/bitwarden.js
@@ -1,6 +1,7 @@
 const ProcessSpawner = require('util/process.js')
 const path = require('path')
 const fs = require('fs')
+var { ipcRenderer } = require('electron')
 
 // Bitwarden password manager. Requires session key to unlock the vault.
 class Bitwarden {
@@ -144,15 +145,7 @@ class Bitwarden {
     }
   }
 
-  getSignInRequirements () {
-    return ['clientID', 'clientSecret']
-  }
-
-  getSignInInstructions () {
-    return l('passwordManagerBitwardenSignIn')
-  }
-
-  async signInAndSave (credentials, path = this.path) {
+  async signInAndSave (path = this.path) {
     // It's possible to be already logged in
     const logoutProcess = new ProcessSpawner(path, ['logout'])
     try {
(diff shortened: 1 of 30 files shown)


aiyagari:
@PalmerAL I have revisited this with the latest release 1.24 to see if it is fixed.  I entered the API key in the new dialog box as requested, and found the following:
- On nytimes.com it works
- On schwab.com it does not work
On both sites, it does put the "key" icon on the field, so it is recognized as a login field by the browser.  Somehow on one of the sites it is just not getting filled in.
This is actually just an example - I find it working on 50% of sites overall.

PalmerAL:
@aiyagari Thanks for letting me know - fixed in 2d1a3d78abf7417fb0951ef88b7dccbdf01b94b9. If you see other sites where it's not working, please file issues for them!

Issue #1374, opened 2020-12-11
Doesn't work on armhf


BioHazard1282:
- **Min Version:** 1.17
- **Operating system:** Ubuntu MATE 20.04 Focal Fossa

## Expected Behavior
Browse, change preferences

## Actual Behavior
Starts up, but doesn't do anything. No page loads.
## To Reproduce
Install the .deb package and run.

kodumbeats:
Looks like the 1.17 release brings some new challenges to compile for 32 bit Windows: 
https://github.com/minbrowser/min/issues/1369#issuecomment-740539113

Might be the same issue for compiling to `armhf`, but I don't have any lying around that are running 32 bit to test.

BioHazard1282:
> Looks like the 1.17 release brings some new challenges to compile for 32 bit Windows:
> [#1369 (comment)](https://github.com/minbrowser/min/issues/1369#issuecomment-740539113)
> 
> Might be the same issue for compiling to `armhf`, but I don't have any lying around that are running 32 bit to test.

Let me know what you get. I am hoping this will be a good and fast browser if things will be fixed, all other browsers are resource hogs and can't do anything with them

BioHazard1282:
I can try using 1.16.1 and see if it works.

BioHazard1282:
Nope. 1.16.1 fails too.

PalmerAL:
Does it work if you start it with --disable-gpu?

I think I have a pi zero lying around, I can test it in the next week or so and see what's happening.

BioHazard1282:
I also have some output for running 1.17.1:
![2020-12-11-130822_1600x900_scrot](https://user-images.githubusercontent.com/44819642/101953100-fbd41900-3bb6-11eb-8b08-79aae63d09e7.png)

[errors.txt](https://github.com/minbrowser/min/files/5681531/errors.txt)

BioHazard1282:
It says "GPU process cannot be used. Goodbye" and exits.

BioHazard1282:
No, it doesn't. It says "GPU process cannot be used. Goodbye."




*----------------------------------------------*
*Arbiter (BioHazard)*
*Senior at Cedar Valley High School*
*Computer and Calculator programmer*

*biohazard1282@gmail.com <biohazard1282@gmail.com>*

*dahlty000@alpinesd.org <dahlty000@alpinesd.org>*
*c3-productions.github.io <http://c3-productions.github.io/>*

PalmerAL:
[...]

PR #1813, opened 2021-12-02
Add buildRaspi to build-packages.yml

ryanfortner:


PalmerAL:
The reason the raspberry pi build was removed originally was that the library used for the password manager wouldn't compile: https://github.com/minbrowser/min/issues/1374. We recently removed the usage of that library: https://github.com/minbrowser/min/pull/1801, but it needs to stay in the release build for the next few releases for migration purposes.

Is compilation for raspberry pi working now? (meaning when you run buildRaspi and install the resulting package, it starts without issues and the password manager works?)

ryanfortner:
Seems to work fine for me, as well as for all the users in [Pi Apps](https://github.com/Botspot/pi-apps)

PalmerAL:
OK thanks, I'll start uploading those with the next release.

PR #1826, opened 2021-12-14
Suggestion for new app icon

Tuba2:
I've seen the work and effort put in the creation of the new app icon.

And although I admire the result of that collaboration, I believe the final product wasn't particularly _iconic_.

With all due respect, it ended up feeling a bit generic, and didn't stand out among other icons.
It was also quite hard to figure out what the icon referred to. It didn't send a clear message.
It wasn't very minimalistic, considering all the small details that it relied upon.
And, finally, it had a scaling problem: once shrunk to a certain size, all the detail was gone and the icon/logo looked even plainer.

Thus I decided to contribute the way I could: by proposing a new logo/icon for the app.
I got together with another designer friend and we worked on a solution.

We decided to try and stylize the app name in a way it could become an iconic figure.
We tried to reference some elements of web browsing, like buttons, banners, and columns, and took as our main criteria both minimalism and fluidity.

That's how we arrived at this result:
icon

The symbol contains both a condensed version of the name (Min) and an abstract representation of a webpage.

I used the colors of the current symbol and followed the design guidelines for macOS Big Sur and forward.

I've created all the necessary variations, and I've included the source vector file, so you can edit and use the logo/icon ad you please.

I hope you like it. It is my way of thanking you for all the effort you put in the creation of this awesome browser that helps me stay focused every day.

I'm open to your feedback — including criticism, of course.


PalmerAL:
Hey, thanks for the effort! I do see your points regarding the current logo, particularly the loss of detail at small sizes. The current logo has only been around for a year, so I'm a bit hesitant to replace it with something new again, but if the new logo is sufficiently better I think it'd be fine to do.

I like the idea behind your proposal, but I think it's trying to fit too many concepts in. If I look at, I mainly see a stylized "M"; I had to look pretty closely to see the full "Min", and I don't see the "webpage" part at all. It also seems a bit crowded to me, which I don't think is good (if anything, I think the goal should be for the icon to look more calming and focused than other browsers).

I could be wrong about this, but I think part of the problem is that there isn't a clear focal point. If I look at it alongside some other browser icons:
Screen Shot 2021-12-19 at 11 16 33 PM
With the Firefox and Chrome icons, I look at around the center of it (the bottom purple area for Firefox, the white-on-green border for Chrome), and my eyes stay there. With the Min icon, I first look near where the path starts in the corner, but then I start following the line up and to the right.

But, not a designer 🤷

Tuba2:
Hi, @PalmerAL ! Thanks a lot for taking such a careful look.

I absolutely see your points, especially the one about it being _crowded_.
Just so I can propose a properly reworked version, I'd like to know the following things:
- Do you like this approach with square-ish yet rounded shapes and simple gradients?
- Do you believe a focal point following this style, instead of a round shape, would work?
- Would it be okay if I brought you more than 1 option on this next round, like 3 ou 4 (max)? I believe it would be worth it, so we can check more boxes at once.

Let me know and I'll work on it.
I'm glad you like the idea, that's enough for me to want to make it happen ;)

Best regards and Happy Holidays!

PalmerAL:
Hey, sorry for taking so long to respond to this. Hope you had a good holiday!

* I don't have much of an opinion on the shapes. I'd definitely keep the blue gradient background so that it's recognizable for people used to the old logo.
* I'm not sure what you mean here.
* Definitely!

Tuba2:
Hey, @PalmerAL !
I had a great time, hope you did as well.

Awesome, now I have mor info to work with.

On my 2nd question I wanted to know if you would like a central focal point in the shape of a circle (as seen in Firefox and Chrome) or if a rounded square-ish shape would also be a valid option :)

I'd like to know that just so I can decide wether I should redo the whole set of shapes or if I can keep the current style.

JoaoNM:
Hey guys! Just found out about min and I think it's an awesome project! I saw the current icon and I think it could do with some development for sure. I like that new option although I agree it lacks a focal point. In my opinion a circle or square-ish focal point works too. 

Also, maybe worth exploring, but how about you try inverting the colours? i.e. make the background white and the foreground in blue, a little bit like the safari logo. Could be interesting! :)

PalmerAL:
> On my 2nd question I wanted to know if you would like a central focal point in the shape of a circle (as seen in Firefox and Chrome) or if a rounded square-ish shape would also be a valid option :)

To be honest, it's pretty hard for me to say without seeing it. Having a circle in the center might help make it more recognizable compared to the old logo, but I suppose it doesn't have to be a circle.

With the current style of a long squiggly line, it seems to me like it would be hard to fit a focal point in or make it less crowded, but I think I'm probably picturing it differently than you are.

PR #1838, opened 2021-12-27
Allow custom image backgrounds on new tab page

### LINKED ISSUE #1173: Re-enable dark new tab page on windows and linux
...which was disabled in 33c7876005c8ba589a81003a45a2d27e387ec8f1. It should be possible to add a class to the body element for tabs with no URL, and only apply the dark background in those cases.

PalmerAL:
Screenshot:
Screen Shot 2021-12-26 at 6 00 07 PM

My intent is to also fix #1173 as part of this, but I haven't done so yet.

### DIFF
--- a/css/modal.css
+++ b/css/modal.css
@@ -15,6 +15,10 @@ body.is-modal-mode #overlay {
   z-index: 1;
   pointer-events: none;
 }
+body.is-edit-mode.is-ntp #overlay {
+  opacity: 0.33;
+}
+
 body.is-modal-mode #overlay {
   /* In modal mode, the tab bar should be disabled */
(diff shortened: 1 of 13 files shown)


cfuendesign:
This is perfect. After I'm done with the multiple site !bangs feature I want to make a userscript that sets the background as Bing's image of the day (Since it's obtainable through an API call). This will make my job much easier.

PalmerAL:
Yeah I think it might be nice to have some kind of auto-rotating background (was thinking about using the Unsplash API). I haven't thought too much about how that should work though.

1173 should be fixed now; I think this is ready to merge if anyone wants to review it.

Issue #1633, opened 2021-06-03
Apple Silicon version doesn't work


iynere:
- 1.19.2
- macOS 11.3.1

## Expected Behavior
I can open Min

## Actual Behavior
"Min is damaged, you should move it to the trash"

## To Reproduce
1. On an M1 mac, download the Apple Silicon version of Min
2. Unzip it and try to open it

balupton:
Encountered this as well. Going to see if the intel version works.

Chamiu:
> Encountered this as well. Going to see if the intel version works.

like "Let them eat cake"
This issue is only not work unziped "Min-v1.19.2-darwin-arm64.zip".

PalmerAL:
In the past, that has gone away if you right click > "open" a second time, and then there should be an "open anyway" option. Does that work for any of you?

Chamiu:
Thanks suggest.
I try this.

> In the past, that has gone away if you right click > "open" a second time, and then there should be an "open anyway" option. Does that work for any of you?

But macOS Big Sur.4.0 say "Min is damaged, you should move it to the trash".

And I try unzip to use The Unarchiver 4.4.3, But this issue didn't change.

I set PreferencePane > Security & Privacy > General > Allow apps downloaded from: as "App Store and identified developers".

> [Safely open apps on your Mac - Apple Support](https://support.apple.com/en-us/HT202491)

I don't think the cause is Gatekeeper.
Perhaps compression as failing. please publish in dmg file.

PalmerAL:
I don't have an Apple developer account (previously because of cost, although I can probably afford it now), so you would need to follow the instructions under "If you want to open an app that hasn’t been notarized or is from an unidentified developer". Does that work?

I'll try to borrow an m1 mac in a few days and see if I can figure this out.

Chamiu:
> I don't have an Apple developer account (previously because of cost, although I can probably afford it now),

Ok, no problem. 

> "If you want to open an app that hasn’t been notarized or is from an unidentified developer". Does that work?

I hope to open dialog.
![macos-big-sur-alert-unverified-developer](https://user-images.githubusercontent.com/2077600/120877741-39a87c00-c5f3-11eb-9ef6-06fa769c141c.png)

But not work.

PalmerAL:
I tried this out on an m1 machine, and it is indeed broken. This worked before, I'm not sure what Apple changed...

[...]

PR #1963, opened 2022-03-26
Update build system to use electron-builder, add AppImage target

cog1to:
#### Disclaimer

I started this branch just to add AppImage target, but then realized that it might be beneficial to switch from the legacy `electron-installer-*` packages to the `electron-builder` system.

#### What's done

- Removed `electron-installer-debian`, `electron-installer-redhat` packages, added `electron-builder`
- Updated `createPackage` script to use electron-builder for creating a packaged app directory
- Updated all `build*` scripts to use electron-builder with `prepackaged` parameter pointing to the directory created in `createPackage` step.
- Added AppImage target

#### What's changed

I was trying to do the update with as little changes as possible to not disrupt the current build process, in case there's any automation/publish hooks dependent on it, but some changes were inevitable. The most noticeable one is that the new build system has a different install path on Linux: it puts the app in /opt/ instead of /usr/ directory. I've updated install scripts accordingly to point to new executable location. Also, the folder structure generated in `dist/app` during the build is somewhat different.

#### What else can be done

- There seems to be an option to switch to proper `dmg`/`pkg` packages for MacOS instead of manually archiving the app folder.
- Windows installers can be updated to use builder's NSIS target.
- For Linux it seems there's an option to build Snap and Pacman packages.
- If I understand the `electron-builder` system, the first two changes will allow to get rid of `createPackage` step, and use a single builder script for all targets.

#### Testing required

I was able to test AppImage, Mac-x64, and Debian-x64 packages. I need some help to verify the correctness of Windows, Mac-arm64, and Debian-arm64 artifacts.


PalmerAL:
Thanks for doing this, the general idea seems pretty good. I'll try to check the windows and debian builds later this week. I don't have an arm64 mac, and it isn't possible to test that in a VM, but we could post in this issue: https://github.com/minbrowser/min/issues/1633 and see if anyone would be willing to test it.

The only automation that depends on these is https://github.com/minbrowser/min/blob/master/.github/workflows/build-packages.yml. It doesn't depend on any internal details, although changing the output folders might break it (I haven't checked yet).

cog1to:
The final build artifacts are still in `dist/app`, so it should be ok, I think

cog1to:
Update: the Apple Silicon build is still broken, it seems. I'll try to investigate further

cog1to:
Was able to verify the Mac Arm64 build on a borrowed laptop. Works fine, it seems

PalmerAL:
Sorry for the late review. I read over the code, and it looks pretty good. Unfortunately, my Virtualbox install has stopped working, so I wasn't able to test the Windows build; if no one is available to test beforehand we can make a beta release (but we can probably post on Discord asking for testers first).

Two minor things:
* Copyright info is getting added to the Mac metadata somewhere - can that be removed?
Screen Shot 2022-04-22 at 10 12 37 PM

* Can you format everything with standard-format?

cog1to:
Thanks, fixed. I didn't realise that "extendInfo" option for mac targets now accepts plain dictionary instead of a plist file. Updated accordingly with explicitly settings copyright string to null.

PalmerAL:
Thanks!

This looks good to me. Let's go ahead and merge it; I'll make a beta release soonish and ask people to test it on Discord, and if there's any issues on platforms that I haven't been able to test, I'll let you know.

PR #1973, opened 2022-04-12
Add missing translations

xephtar:
There is no specific translated word for bang. I think it is global term. So, it shouldn't be translated.


PalmerAL:
Looks good, thanks!

PR #1980, opened 2022-04-24
Task bangs with list

flightmansam:
### Discussed in https://github.com/minbrowser/min/discussions/1978

<div type='discussions-op-text'>

<sup>Originally posted by **flightmansam** April 21, 2022</sup>
I think the reorganisation of tasks is very powerful. But I have found if I have a lot of tasks opened (not collapsed), it becomes cumbersome to drag tabs into the right task using the task overlay. Good thing there is the **!movetotask** bang to help me!

However I am a forgetful person and often forget the exact names/indexs of my tasks. So I have quickly tied together some search result functionality to **!movetotask** (and also **!task** task switcher, as they have the same call signature/functional behaviour).

I also disabled the transition through the task overlay ([L41-L46](https://github.com/flightmansam/min/blob/d73067696c70533db2d1bca5d6990afdd6bb5776/js/searchbar/tabMoverPlugin.js#L41-L46) in my branch), because I find it cumbersome, but could understand if that was put back if this makes it to the master branch. Sidenote: maybe we could have a preference in the settings "Reduce UI animations". This was also the basis for my [discussion here](https://github.com/minbrowser/min/discussions/1975#discussion-4021908).

https://user-images.githubusercontent.com/34013162/164276030-49188b76-9479-4165-addf-334d79e71d51.mov

You can check out the functionality in my [feature branch fork](https://github.com/flightmansam/min/tree/switch-and-movetotask-with-task-list) of min (beware it is a branch based off many of my own custom commits).

Or the specific files are in this commit below:

https://github.com/flightmansam/min/commit/d73067696c70533db2d1bca5d6990afdd6bb5776

Todo items:
- refactor out the getTaskByNameOrNumber(text) function as it is now duplicated in multiple places (sorry, that's just me being agile!)
- I have found that on a bunch of features I have worked on or looked at, we need a way of getting a formatted task name from the task index (i.e. "Task 1" from the 0th element in the tasks TaskList), once again there is a bunch of code duplication around the place (including here). We might want to refactor this into a task (or TaskList) specific functionality. </div>


flightmansam:
I think having !switchtotask in a tabMoverPlugin.js file makes it confusing. Maybe I put it all back into the customBangs.js?

flightmansam:
My only other thoughts are about how the list is sorted. Should it be sorted in task order (of the TaskList array) or should it be sorted by recently used tasks, or alphabetically... Let me know if you have a preference. I am leaning towards most recently used tab, but this would require implementing a lastActivity attribute and updating it. It's probably good as is now unless someone else pipes up about it.

flightmansam:
I reckon that it is probably best not to include **!closetask** with this feature. As it would be too easy to accidentally close your other task without warning if you used the dropdown list. What do you think @PalmerAL?

PalmerAL:
1. Don't have a strong preference which file it goes in.
2. New tasks are added to the end of the task array IIRC, so the ones at the beginning are going to be the oldest/least-useful. I'd vote for sorting by recently-used; there's a `tasks.getLastActivity` method that will do this for you by looking at the most-recently-used tab. (Although it's a bit slow to call repeatedly, so you'll want to do something like tasks.map(t => {t, time: t.getLastActivity()}) and then sort based on that).
3. Not really sure; I guess it's fine to leave it out. Although, just typing `!closetask ` and pressing enter closes the current task, and I don't recall any issues reported from people running it accidentally.

One other thought: maybe the list should have the lastActivity date next to each task? Otherwise, if you have a list of unnamed tasks, it's hard to figure out what the right task is:
Screen Shot 2022-04-28 at 3 32 44 PM

flightmansam:
I pinched the history viewer code to add a secondary text with the relative date formatter. I kinda want the text aligned though... You'll note here that "Just now" is a little further to the right than "In the past hour". Not sure how to do this, is it a css thing or do I need to fiddle with like hbox parents or something?

image

Alternatively we could group them and add headings like in **!history**, might be more consistent...

flightmansam:
This is kinda obvious but... Probably shouldn't include the current task in the selection list due to redundancy!

flightmansam:
With commit [e966158](https://github.com/minbrowser/min/pull/1980/commits/e9661585da991d6b93ad45d0d58e6e7974a885d0) I added some fuzzy searching to the bang. This way you don't have to do as much arrow key madness. You can still create new tasks from the move by typing any text and hitting enter (without focusing any of the search results).

https://user-images.githubusercontent.com/34013162/167147750-c40977c9-4214-4e03-a7cf-c9c9c85ade59.mov

js/searchbar/customBangs.js
```diff
@@ -12,8 +12,126 @@ var places = require('places/places.js')
...
+function searchAndSortTasks(text) {
+  var sortLastActivity = tasks.map(t => Object.assign({}, { task: t }, { lastActivity: tasks.getLastActivity(t.id) }))
+
+  sortLastActivity = sortLastActivity.sort(function (a, b) {
+    return b.lastActivity - a.lastActivity
+  })
+
+  var isSingleNumber = /^\d+$/.test(text)
+
+  if (text !== '' ? !isSingleNumber : !isSingleNumber) { //lXOR
```
PalmerAL, line 101: What's the reason for disallowing fuzzy search for names that are numbers? You could have task names that are long numbers, and it seems useful to support fuzzy search in that case.

(Also, this can be simplified to `!isSingleNumber` I think?)

js/searchbar/customBangs.js
```diff
@@ -12,8 +12,126 @@ var places = require('places/places.js')
...
+        matches.push({
+          task: t,
+          score: fuzzyTitleScore + exactMatch
+        })
+      }
+    })
+
+    matches = matches.sort(function (a, b) {
+      return b.score - a.score
+    })
```
PalmerAL, line 123: I don't think we should reorder matches based on score - it seems confusing to have tasks be ordered by most-recent most of the time, but occasionally not be. Filtering to include things above the threshold and then sorting by most-recent should be enough.

js/searchbar/customBangs.js
```diff
@@ -12,8 +12,126 @@ var places = require('places/places.js')
...
+
+    sortLastActivity = matches.map(t => t.task)
+
+    selected = sortLastActivity[0]
```
PalmerAL, line 127: This isn't used anywhere.

js/searchbar/customBangs.js
```diff
@@ -82,47 +200,84 @@ function initialize () {
...
+        if (task.id != tasks.getSelected().id) {
+          var taskName = (task.name ? task.name : l('defaultTaskName').replace('%n', tasks.getIndex(task.id) + 1))
+
+          var data = {
+            title: taskName,
+            secondaryText: formatRelativeDate(lastActivity),
+            fakeFocus: false,
+            click: function () {
+              tabEditor.hide()
+              moveToTask('%n'.replace('%n', tasks.getIndex(task.id) + 1))
```
PalmerAL, line 226: This is confusing - we have a task ID already, we're converting it to a string, and then we're searching through all the tasks to convert that back into an ID. Also, it's possible to have multiple tasks with the same name, and this might cause us to pick a task that's different than what the user clicked on.

How about:
* Move the get/create task logic in `moveToTask` into its own function.
* Then moveToTask / switchToTask can take a task ID, and the callers can do `moveToTask(getTaskByNameAndNumber(text))` or `moveToTask(addOrCreateTask(text))` where necessary.

PalmerAL (commented):
This works really well! Left a few small comments.

I can't think of a good way to align the text. The issue is that "2" is slightly wider than "1", but we don't have the option to use a monospace font here, and you can't set the task name container to be a fixed width either. It's probably not worth worrying about - if any of the tasks have custom names, they're not going to be aligned anyway.

js/searchbar/customBangs.js
flightmansam: This is trying to remain consistent with previous functionality, e.g. typing in 1 (and only that) should move to task 1. 

L101 is a logical XOR (I'm pretty sure). SO the idea is if the bang has no text OR is only numbers draw the list sorted by just recent activity. Now you bring it up, I suppose if the task is JUST a number, it should probably display the list in task index order (rather than recent activity).

However tasks with numbers in the name e.g. My Bean Recipes 22/05 should still fuzzy search.

I agree it is confusing. I never switch task by index anyway....

js/searchbar/customBangs.js
flightmansam: Oh yeah L127 was when I was experimenting with (fake)focusing . That needs to go.

js/searchbar/customBangs.js
flightmansam: Oh yes of course. It's a bit clunky. I had it in my head that by using the index (found by id and not name) would protect against duplicate names. But it is very clearly messy.

PalmerAL:
@flightmansam Do you have time to finish this up soon? I might be able to in a few days if you don't.

flightmansam:
> @flightmansam Do you have time to finish this up soon? I might be able to in a few days if you don't.

I don't have much time this weekend. All the functionality is there (I think?) just about changing the structure as per you recommendations. Happy for you to do that, if you have the time.

Fun fact... on my fork I added the ability to just search for the task you want in the normal searchbar. I use it all the damn time.

PalmerAL:
Sounds good, we can see who gets to it first.

> Fun fact... on my fork I added the ability to just search for the task you want in the normal searchbar. I use it all the damn time.

Good idea!

PalmerAL:
I've made some changes in 55f4de85d4798948f487addf2ae26eaff3dedfee that I think address everything I mentioned (allowing fuzzy search with numbers, using task IDs instead of strings), and also made some UI changes:

* Tasks are always sorted by most-recent now instead of being reordered.
* When you search for a task, the first result is highlighted (like how !bookmarks works), and pressing enter selects it.

Let me know what you think!

flightmansam:
Looks good to me. I noticed focus mode warning will be shown twice. Once in the movetotask bang and once in the moveToTaskCommand. I have deleted the one in the bang code.

PalmerAL:
Oh, good point. The focus mode check needs to happen before the new task is created, so I've changed it in the last commit.

Is there anything else you see that needs to be done here, or is this good to merge?

flightmansam:
Happy as larry!

PalmerAL:
Great, thanks for working on this!

PR #1992, opened 2022-05-12
Updated iconfont files

flightmansam:
I wanted to add the `carbon:autoscale` icon in a feature on my personal fork. Found that the iconfonts file we are referencing does not include all of the current carbon library. I see the repo doesn't package the iconfont files, but I used the following method to update what we have. Note, I haven't done a check against the current https://github.com/carbon-design-system/carbon-icons library for completeness. Maybe in the future there will be a script as part of the npm install or npm build that checks for latest icons?

Used this place: https://icones.js.org/collection/carbon
->click 3 dots -> "Downloads" -> iconfonts
- reformatted download to suit min's format in ext/icons


PalmerAL:
Looks good! A script would be nice, but updates should be infrequent enough that I don't think its necessary to add now.

PR #2004, opened 2022-06-06
Implement macOS handoff support

PalmerAL:
@flightmansam I implemented support for macOS handoff based on your suggestion in https://github.com/minbrowser/min/discussions/1988#discussioncomment-2741427 . The actual handoff implementation is really simple, since Electron has built-in APIs for it. I ran into some issues with the timing of receiving the handoff request while Min was being launched (if you tried to use handoff while Min wasn't already running), so I had to restructure some of the window-creation code to make it work.

If you have time, would you be willing to review this?

### DIFF
--- a/js/default.js
+++ b/js/default.js
@@ -153,6 +153,7 @@ require('pageTranslations.js').initialize()
 require('sessionRestore.js').initialize()
 require('bookmarkConverter.js').initialize()
 require('newTabPage.js').initialize()
+require('macHandoff.js').initialize()
 
 // default searchbar plugins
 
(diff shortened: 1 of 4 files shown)


PalmerAL:
Yup, I broke it, thanks for catching that. Should be fixed in ee73e1f930bd69eee42c4af8142bbe30f3e2b067?

flightmansam:
Yep looking real good on my end when testing features in the [command-line args discussion](https://github.com/minbrowser/min/discussions/2008). This is very nice feature to have in min.

PalmerAL:
Great!

PR #2010, opened 2022-06-10
Improve czech localization

BlueManCZ:
Reviewed and improved Czech localization. Added missing translations.

### DIFF
--- a/localization/languages/cs.json
+++ b/localization/languages/cs.json
@@ -81,7 +81,7 @@
     "findMatchesPlural": "%i z %t výsledků",
     /* Focus mode */
     "isFocusMode": "Jste v režimu zaměření.",
-    "closeDialog": "Rozumím", //used as a label for the button that closes the dialog
+    "closeDialog": "Zavřít", //used as a label for the button that closes the dialog
     "focusModeExplanation1": "V režimu zaměření nemůžete vytvářet nové karty nebo přepínat dimenze.",
     "focusModeExplanation2": "Režim zaměření můžete opustit odškrtnutím \"režim zaměření\" v menu zobrazení.",
     /* relative dates */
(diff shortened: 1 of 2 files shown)


localization/languages/cs.json
```diff
@@ -81,9 +81,9 @@
     "findMatchesPlural": "%i z %t výsledků",
     /* Focus mode */
     "isFocusMode": "Jste v režimu zaměření.",
-    "closeDialog": "OK", //used as a label for the button that closes the dialog
-    "focusModeExplanation1": "V režimu zaměření nemůžete přepínat záložky nebo měnit dimenze.",
-    "focusModeExplanation2": "Můžete opustit režim zaměření odškrtnutím \"režim zaměření\" v menu zobrazení.",
+    "closeDialog": "Rozumím", //used as a label for the button that closes the dialog
```
PalmerAL, line 84: If I understand correctly, this means "I understand", which makes sense for the focus mode dialog. However, this also gets used for the [about dialog](https://github.com/minbrowser/min/blob/d8ed7acad9230e7fdba4e801560ad9b8221d66e8/main/menu.js#L409) and potentially other dialogs in the future. Does this still make sense in that case?

localization/languages/cs.json
```diff
@@ -117,7 +117,7 @@
     "phishingErrorVisitAnyway": "Přesto stránky navštívit",
     "phishingErrorLeave": "Opustit tyto stránky",
     /* multiple instances alert */
-    "multipleInstancesErrorMessage": "Vyskytla se chyba. Prosím zavřete všechny ostatní spuštěné instance a restartujte Min.",
```
PalmerAL, line 120: What's the reason for this change?

PalmerAL:
Thanks! Looks good aside from comments.

localization/languages/cs.json
BlueManCZ: You are right. In other dialogs it doesn't make much sense. Let's use universal "close" instead.

localization/languages/cs.json
BlueManCZ: It sounds more civilized to reference Min as a browser. But since I used "browser Min" phrase in other translations, let's use it here as well.

PalmerAL:
Looks good, thanks!

PR #2013, opened 2022-06-14
Updates to command-line arguments 

flightmansam:
Discussed in https://github.com/minbrowser/min/discussions/2008.

I would like to add a `-t` argument to request specific tasks from the command-line. This may be an opportunity to look at the code structure of how min handles command-line arguments to avoid confusion, inadvertent bugs, and easier extension later on. One opportunity could be to use a specific library like [commander.js](https://github.com/tj/commander.js).

This branch has some uglies at the moment due to my agile desire to just have the feature I want:
- I use the searchAndSortTasks and moveToTaskCommand from customBangs.js, so this should be separated out (probably to task.js @PalmerAL?)
- BrowserUI ipc 'switchToTask' and BrowserUI.addTab() share commonality so it should be separated out?
- Currently the -t argument tries to handle both a task name or a task id by doing an "is all digits?" comparison. Probably should eliminate this ambiguity by having a switch for a task id (maybe --task-id)
- Would like to have a command line way of switching to a particular tab (maybe -tab, and changing -t to be -task).

### DIFF
--- a/js/menuRenderer.js
+++ b/js/menuRenderer.js
@@ -10,6 +10,7 @@ var PDFViewer = require('pdfViewer.js')
 var tabEditor = require('navbar/tabEditor.js')
 var readerView = require('readerView.js')
 var taskOverlay = require('taskOverlay/taskOverlay.js')
+var { searchAndSortTasks, moveToTaskCommand } = require('searchbar/customBangs.js')
 
 module.exports = {
   initialize: function () {
@@ -94,6 +95,18 @@ module.exports = {
       browserUI.addTab(newTab, {
         enterEditMode: !data.url // only enter edit mode if the new tab is empty
       })
+
+      if (data.taskQuery) {
+      // use the first search result
+      // if there is no result, need to create a new task
+      let task = searchAndSortTasks(data.taskQuery, excludeSelected=false)[0]?.task
+      if (!task) {
+        task = tasks.get(tasks.add(undefined, tasks.getIndex(tasks.getSelected().id) + 1))
+        task.name = data.taskQuery
+      }
+
+      moveToTaskCommand(task.id)
+      }
     })
 
     ipc.on('saveCurrentPage', async function () {
--- a/js/menuRenderer.js
+++ b/js/menuRenderer.js
@@ -97,15 +97,55 @@ module.exports = {
       })
 
       if (data.taskQuery) {
-      // use the first search result
-      // if there is no result, need to create a new task
-      let task = searchAndSortTasks(data.taskQuery, excludeSelected=false)[0]?.task
-      if (!task) {
-        task = tasks.get(tasks.add(undefined, tasks.getIndex(tasks.getSelected().id) + 1))
-        task.name = data.taskQuery
+        // use the first search result
+        // if there is no result, need to create a new task
+        let task
+
+        if (/^\d+$/.test(data.taskQuery)) {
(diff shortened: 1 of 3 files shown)


main/main.js
```diff
@@ -116,23 +116,38 @@ function openTabInWindow (url) {
 
 function handleCommandLineArguments (argv) {
   // the "ready" event must occur before this function can be used
+  var initTaskQuery = undefined
   if (argv) {
+
+    // check for -t task query
+    if (argv.includes('-t') && argv.indexOf('-t') > 0){
+      // query for specific task to add search to
+      initTaskQuery = argv[argv.indexOf('-t') + 1]
```
PalmerAL, line 125: I think we can assume that IPC messages will be received in the order they're sent, in which case we can just send a `switchToTask` IPC here, and then leave the code below as `switchToTab` without any task information. That should simplify this a fair amount.

main/main.js
```diff
@@ -367,14 +382,14 @@ app.on('continue-activity', function(e, type, userInfo, details) {
   }
 })
 
-app.on('second-instance', function (e, argv, workingDir) {
+app.on('second-instance', function (e, argv, workingDir, additionalData) {
```
PalmerAL, line 385: `argv` here should be the arguments of the second instance already - why do they need to be passed in with `additionalData`?

js/menuRenderer.js
```diff
@@ -94,6 +95,58 @@ module.exports = {
...
+        focusMode.warn()
+        return
+      }
+
+      if (data.taskQuery) {
+        // use the first search result
+        // if there is no result, need to create a new task
+        let task
+
+        if (/^\d+$/.test(data.taskQuery)) {
```
PalmerAL, line 137: Maybe this would be simpler / better as:
```
if (tasks.get(data.taskQuery)) {
    task = tasks.get...
} else {
    task = searchAndSortTasks...
}

if (!task) {
    task = tasks.get(tasks.add(...
}
```
So task IDs would work, but any number that's not a task ID would be used for search or to create a new task.

js/menuRenderer.js
```diff
@@ -94,6 +95,58 @@ module.exports = {
...
+        } else {
+          task = searchAndSortTasks(data.taskQuery, excludeSelected=false)[0]?.task
+        }
+
+        if (!task) {
+          task = tasks.get(tasks.add(undefined, tasks.getIndex(tasks.getSelected().id) + 1))
+          task.name = data.taskQuery
+        }
+
+        moveToTaskCommand(task.id)
```
PalmerAL, line 115: IIUC, this is creating a tab in the current task, then using `moveToTaskCommand` to move it to the new task. This should just be a switchToTask call before the addTab call instead.

(However, if we send a switchToTask IPC from the main process, as I suggested in my other comment, this whole block of code should go away anyway).

PalmerAL (commented):
Good idea!

> I use the searchAndSortTasks and moveToTaskCommand from customBangs.js, so this should be separated out (probably to task.js)

I think the moveToTaskCommand call can be removed (explained in comments). Separating searchAndSortTasks does seem like a good idea, it could either go in a file in `js/util` or maybe a method on `tasks`?

flightmansam:
Sorry I have done nothing more on this topic! Not a very high priority for me at the moment ahah.

PalmerAL:
No worries!

PR #2015, opened 2022-06-15
improved Hungarian localization

sndrkrly:
Reviewed, and change some for better understanding, and added new translates.

### DIFF
--- a/localization/languages/hu.json
+++ b/localization/languages/hu.json
@@ -6,28 +6,28 @@
         these are the items displayed in the menu when you right-click on a page
         */
     "addToDictionary": null, //missing translation
-    "pictureInPicture": null, //missing translation
+    "pictureInPicture": "Kép a képben", //missing translation
     "openInNewTab": "Megnyitás új fülön",
     "openInNewPrivateTab": "Megnyitás új privát fülön",
-    "saveLinkAs": null, //missing translation
+    "saveLinkAs": "Link mentése mint", //missing translation
     "viewImage": "Kép megtekintése",
     "openImageInNewTab": "Kép megnyitása új fülön",
     "openImageInNewPrivateTab": "Kép megnyitása új privát fülön",
     "saveImageAs": "Kép mentése másként",
     "searchWith": "Keresés %s keresővel", //%s will be replaced with the name of the current search engine
     "copyLink": "Hivatkozás másolás",
-    "copyEmailAddress": null, //missing translation
-    "selectAll": null, // Missing Translation
+    "copyEmailAddress": "E-mail cím másolása", //missing translation
+    "selectAll": "Mind kijelölés", // Missing Translation
     "undo": null, // Missing Translation
     "redo": null, // Missing Translation
-    "cut": null, // Missing Translation
+    "cut": "Vágás", // Missing Translation
     "copy": "Másolás", //this is a verb (as in "copy the currently-selected text")
     "paste": "Beillesztés",
     "goBack": "Vissza lépés",
     "goForward": "Előre lépés",
     "inspectElement": "Részletek megtekintése",
     /* searchbar */
-    "pasteAndGo": null, //missing translation
+    "pasteAndGo": "Beillesztés és megnyitása", //missing translation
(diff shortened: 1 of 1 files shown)


localization/languages/hu.json
```diff
@@ -6,28 +6,28 @@
         these are the items displayed in the menu when you right-click on a page
         */
     "addToDictionary": null, //missing translation
-    "pictureInPicture": null, //missing translation
+    "pictureInPicture": "Kép a képben", //missing translation
```
PalmerAL, line 9: Can you remove the `//missing translation` comments where you've added translations?

localization/languages/hu.json
```diff
@@ -39,21 +39,21 @@
     "viewSettings": "Beállítások megtekintése",
     "takeScreenshot": "Képernyőkép készitése",
     "clearHistory": "Összes történet törlés",
-    "enableBlocking": null, //missing translation
-    "disableBlocking": null, //missing translation
-    "clearHistoryConfirmation": null, //missing translation
+    "enableBlocking": "Blokkolás engedélyezése", //missing translation
+    "disableBlocking": "Blokkolás letiltása", //missing translation
+    "clearHistoryConfirmation": "Böngészési előzmények törlése", //missing translation
```
PalmerAL, line 44: The original English version is a question - should this have a question mark?
```
    "clearHistoryConfirmation": "Clear all history and browsing data?",
```

localization/languages/hu.json
```diff
@@ -69,10 +69,10 @@
     "newTask": "Új feladat", //"new" is a verb - it is used for a button at the bottom of the task overlay
     "defaultTaskName": "Feladat %n", //this is the name used for newly-created tasks; %n is replaced with a number ("task 1", "task 2", etc)
     "taskDeleteWarning": {
-      "unsafeHTML": null //missing translation
+      "unsafeHTML": "Nem biztonságos HTML fájl" //missing translation
```
PalmerAL, line 72: This should be translated based on the English version:
```
    "taskDeleteWarning": {
      "unsafeHTML": "Task deleted. <a>Undo?</a>"
    },
```

(`unsafeHTML` is just a marker that this translation string has HTML inside it; it isn't displayed to the user)

localization/languages/hu.json
```diff
@@ -69,10 +69,10 @@
...
     },
     "tasksSearchTabs": null, //missing translation
-    "returnToTask": null, //missing translation
+    "returnToTask": "Vissza a feladathoz", //missing translation
```
PalmerAL, line 75: Can you confirm this is correct? I would expect it to specify "previous task"

English is:
```
    "returnToTask": "Return to your previous task",
```

localization/languages/hu.json
```diff
@@ -88,8 +88,8 @@
     "timeRangeJustNow": "Idő intervalum, éppen most",
     "timeRangeMinutes": "Idő intervalum, perc",
     "timeRangeHour": "Idő intervalum, óra",
-    "timeRangeToday": null, //missing translation
-    "timeRangeYesterday": null, //missing translation
+    "timeRangeToday": "Idő intervalum, ma", //missing translation
+    "timeRangeYesterday": "Idő intervalum, tegnap", //missing translation
```
PalmerAL, line 92: I think these should just be "ma" / "tegnap"?

(If so, the existing translations are wrong also - would you mind fixing those?)

PalmerAL (commented):
Thanks! I left a few comments.

The names of the strings in the translation file are just identifiers that are used in the code - they don't always match the actual text. The actual original text of the strings is listed in the English file: https://github.com/minbrowser/min/blob/master/localization/languages/en-US.json

localization/languages/hu.json
sndrkrly: Of course.

localization/languages/hu.json
sndrkrly: It is correct, I just missed the "previous" (elöző) word, so I'll fix it along with the others.
And thanks for the comments, It helps a lot.

PR #2037, opened 2022-07-02
v8-compile-cache for (slightly) faster load times

djtech-dev:
This pull requests introduces [`v8-compile-cache`](https://github.com/zertosh/v8-compile-cache), a small library that speed up instantiation time using V8's code cache system.
At the moment, `v8-compile-cache` is integrated in `js/default.js` and `main/main.js`.
If needed, `require('v8-compile-cache);` can be used in other modules if they need a (little) speed up.

### DIFF
--- a/js/default.js
+++ b/js/default.js
@@ -1,3 +1,5 @@
+require("v8-compile-cache");
+
 window.globalArgs = {}
 
 process.argv.forEach(function (arg) {
--- a/main/main.js
+++ b/main/main.js
@@ -1,3 +1,5 @@
+require("v8-compile-cache");
+
 const electron = require('electron')
 const fs = require('fs')
 const path = require('path')
--- a/package.json
+++ b/package.json
@@ -39,7 +39,8 @@
     "pdfjs-dist": "2.12.313",
     "regedit": "^3.0.3",
     "stemmer": "^1.0.5",
-    "string_score": "^0.1.22"
+    "string_score": "^0.1.22",
+    "v8-compile-cache": "^2.3.0"
   },
   "devDependencies": {
     "archiver": "^4.0.1",


djtech-dev:
This pull request got expanded, beacuse now it also adds (experimental) Widevine support for websites like Netflix or Spotify.

## How does it work?
If possible, Min will load the Widevine libraries from an installed instance of Google Chrome and it will load it with Electron.

## Why does it use this method?
The main three reasons are:
- This is the easiest way without a license
- This is used by the [Electron.js doc](https://www.electronjs.org/docs/v14-x-y/tutorial/testing-widevine-cdm) itself 
- This is also used by some Kodi extensions for playing multimedia content protected by Widevine

## On which platform does it work?
The code support Linux, Windows and MacOS, for x86 both for 32 and 64bits, but this support for the DRM requires testing on all of the platforms.
In the feature, I hope that support for MacOS on Apple Silicon could be implemented (I use Linux, so I can't test it).

## Why is this one big pull request and not two?
Beacuse with this pull requests all the dependencies for the project got updated to the latest version, resolving many security and compatibility issues, so both the changes are tested with this new project update.

PalmerAL:
Have you measured this before/after to see if it's faster? The reason I ask is that for `main.js`, the amount of code loaded is pretty small, so I wouldn't expect caching to make much difference. For `default.js` (which is fairly large), Electron should have code caching enabled already: https://www.electronjs.org/docs/latest/api/browser-window#new-browserwindowoptions (see "v8CacheOptions").

For Widevine - I don't understand the licensing well enough to know what's permissible or not at the moment. In order to avoid potential issues, I think it's best not to support that. Sorry.

djtech-dev:
Hi.

I've checked, and Electron.js does (by default) use code caching.
For Widevine, I understand the possibility for legal issue.

I'll close this PR, but I will chec the Widevine License/ToS for checking if is possible to embed the DRM (legally).

djtech-dev:
In the PR I also update the dependencies for the project; if you wish, I can make another PR with just the updates. (see `package.json` from [this commit](https://github.com/minbrowser/min/pull/2037/commits/636f633a202091571f569be38b90928c8a118daa))

I also fixed a bug that blocked the use of SharedArrayBuffers (without this API sites like Google Earth can't run) in [this pull request](https://github.com/minbrowser/min/pull/2038)

PR #2063, opened 2022-08-01
Enable pinch zoom for webpages

PalmerAL:


GGIEnrike:
I'm genuinely impressed by the user experience of this browser after FIVE YEARS of development. And how two lines of code aren't getting merged for so long already. Definitely not using it anytime soon! It seems that my small efforts in creating my own browser also with electron are already much farther progressed in some of the most essential aspects. (I started the project before finding Min, so don't think that I have somehow based my idea on this project. I looked for some sort of a browser base to base my "dream browser" off of, probably thinking of using the chromium directly, with C++ at that time. And it actually was the [Wexond Browser Base](https://github.com/wexond/browser-base) that I found and got inspired/guided.)

PR #2075, opened 2022-08-22
Initial functionality google-search-suggestions

flightmansam:
Added search url (for english only).
Tentatively disabled search engine check in instantAnswerPlugin

### DIFF
--- a/js/searchbar/instantAnswerPlugin.js
+++ b/js/searchbar/instantAnswerPlugin.js
@@ -61,9 +61,9 @@ var instantAnswers = {
 
 function showSearchbarInstantAnswers (text, input, event) {
   // only make requests to the DDG api if DDG is set as the search engine
-  if (searchEngine.getCurrent().name !== 'DuckDuckGo') {
-    return
-  }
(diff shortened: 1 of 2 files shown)


flightmansam:
TODO:
Decide whether instantAnswerPlugin needs to fetch a preferences option to always provide instant answers using DDG even if the default search engine isn't DDG.

flightmansam:
@PalmerAL Is this okay to merge?

PalmerAL:
I don't want to use the Google URL unless they have an official API we can use that they won't try to block. We could make this an option when you set a custom search engine though.

gribanoveu:
Is any telemetry being sent to google at this point?

flightmansam:
Not any more than manually putting in the URL in any browser.

PalmerAL:
Oh, I'm realizing I never followed up on the DDG part of this PR, sorry. I believe the original reason for that restriction was due to a restriction in DDGs API terms, although I can't find anything supporting that anymore. They've since deprecated the API entirely and deleted all the documentation, so we really should figure out an alternative to use; I'm not sure at the moment what that would be though.

PR #2076, opened 2022-08-22
fix: change libsecret-1-dev to libsecret-1-0 for build debian

lsnepomuceno:
While trying to install the browser on Debian based distributions like Ubuntu and Zorin OS, I came across the dependency error stating that the libsecret-1-dev package was not found.

In other projects I participate in, by default we don't send dev packages to build in production.

This PR fixes the catch below.

min-browser-install-error

Congratulations on the project, it is very useful.


lsnepomuceno:
reference issue: https://github.com/minbrowser/min/issues/1753

PalmerAL:
This seems reasonable, thanks.

PR #2112, opened 2022-10-14
Add Indonesian localization

sglkc:
Truly a minimal browser and the performance is good too, I don't regret knowing this...

In this pull request, I'd like to add a localization for Indonesian language. I used Firefox's for references so it better be well understood for others too!


PalmerAL:
Looks good, thanks! Sorry for the slow response.

PR #2115, opened 2022-10-16
Update de translations

weases:
- Add missing german translations for tag management
- Fix translation for key 'tasksSearchTabs'


PalmerAL:
Thanks!

PR #2122, opened 2022-10-30
add missing german translation

titusio:


PalmerAL:
Somebody else beat you to it #2115. Thanks for the contribution though!

PR #2158, opened 2022-12-29
deps: add a dependabot to the repo ?

Sanix-Darker:
## WHAT

Just setting up a dependabot yaml file for depencies in the package.json


PalmerAL:
Dependency updates usually require manual testing, so I think it's best to just do this when I (or someone else) has time to do the testing rather than every week. Thanks for the suggestion though!

PR #2182, opened 2023-02-07
Do not change the tab activity colour (fade) when quickly skipping past tabs.

flightmansam:
If you are switching between tabs and you only stay in the tab for <.5s the activity (fade) colour will remain whatever it was previously.

Before:
EE039E0A-9756-41DE-8507-6F427DB59FE3

After:
5DAF20F1-482D-448B-AACB-ACE7288E8B06

### DIFF
--- a/js/navbar/tabActivity.js
+++ b/js/navbar/tabActivity.js
@@ -4,6 +4,28 @@ var tabBar = require('navbar/tabBar.js')
 
 var tabActivity = {
   minFadeAge: 330000,
+  refreshInterval: 7500,
+  tabSelected: function() {
+      var selected = tabs.getSelected()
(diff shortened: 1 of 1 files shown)


flightmansam:
@PalmerAL Any thoughts on this?

PalmerAL:
Sorry for the delay. I like the idea! Regarding the implementation, it doesn't make much sense to me for lastActivity to go backwards. What about implementing this in the tabs model code? https://github.com/minbrowser/min/blob/master/js/tabState/tab.js#L140

Also, there's a small bug - if you opened the browser with an old set of tabs previously, they would be faded out immediately as a side effect of `tab-selected` triggering a refresh. Now they start un-faded and fade out 7.5s later.

PR #2188, opened 2023-02-14
remove dangerous pacman -Sy command for arch linux install

pswilde:
Using `pacman -Sy <package>` in Arch Linux distributions is dangerous as it can lead to dependency mismatches and a "partial upgrade".   
using `pacman -S <package>` is the correct process, or if you want to update to the latest versions, using the full upgrade command, `pacman -Syu <package>` will update all packages and install the latest version of your package    
https://wiki.archlinux.org/title/System_maintenance#Partial_upgrades_are_unsupported


PalmerAL:
Good to know, thanks!

Issue #2149, opened 2022-12-20
'Reader View' zoom in / out via ctrl + mouse wheel or ctrl++ / ctrl+- not working


usiname:
- **Min Version:** v1.26.0
- **Operating system:** raspian os

## Expected Behavior
zoom in / out

## Actual Behavior
nothing happens

## To Reproduce
1. change to 'reader view'
2. zoom in / out by 
 a) ctrl + mouse wheel 
 or
 b) ctrl++ / ctrl+-

PR #2189, opened 2023-02-16
Fixed ctrl+wheel zooming in reader mode

Nassim-Saboundji:
A fix for : ['Reader View' zoom in / out via ctrl + mouse wheel or ctrl++ / ctrl+- not working](https://github.com/minbrowser/min/issues/2149). Though addresses only ctrl + mouse wheel zooming.

How it looks like in practice 

### Caveat
Not sure how this will behave in MacOS which uses the Cmd key instead of Ctrl. Do MacOS browsers make the substitution automatically or this case needs to be handled in code as well?

I could use e.metaKey but it will display true if the windows key is pressed.

https://user-images.githubusercontent.com/70408318/219492984-132f39e2-bb1e-4818-944a-18b070abe4ee.mp4

### DIFF
--- a/reader/reader.js
+++ b/reader/reader.js
@@ -265,7 +265,31 @@ function startReaderView (article, date) {
 
     window.addEventListener('resize', setReaderFrameSize)
   }
-  document.body.appendChild(rframe)
+
+  const overlayForZooming = document.createElement('div')
+  overlayForZooming.id = 'overlay'
+
+  rframe.addEventListener('load', function () {
+    rframe.contentWindow.document.addEventListener('keydown', function (e) {
+      if (e.getModifierState('Control')) {
+        overlayForZooming.style.pointerEvents = 'auto'
+      }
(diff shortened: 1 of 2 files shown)


PalmerAL:
This is a neat approach, but I think we should investigate first why ctrl+scroll inside an iframe doesn't work - I would expect that to zoom the whole page in or out (on any site, not just reader mode). The way that's implemented is that there's a [preload script](https://www.electronjs.org/docs/latest/tutorial/tutorial-preload) that runs inside the webpage that sends mouse events to the UI process:
https://github.com/minbrowser/min/blob/master/js/preload/default.js#L19-L22

And then in the parent process, it listens for those events and zooms the page:
https://github.com/minbrowser/min/blob/master/js/webviewGestures.js#L167-L176

My guess is that the mouse events aren't being forwarded from the iframe correctly somehow, but I haven't looked into it.

PR #2194, opened 2023-02-21
Beautify the input box of the settings page

XYZscratcher:


manoger:
I would like to see some "before-after" printscreens 😊

PR #2222, opened 2023-04-20
Fix/tab item alignment

RubenSmn:
# Description

Improves the item alignment in a tab.

Changes made:

- Vertical align items in a tab, introduced `display: flex`
- Fix misplaced hover state on buttons
- Fix permission buttons spacing
- Buttons now have a consistent height / width

|Before|
|:-:|
|_home_ruben_Development_forks_min_index html|
|After|
|_home_ruben_Development_forks_min_index html

*I forced some active states to show the icons.*

### DIFF
--- a/css/tabBar.css
+++ b/css/tabBar.css
@@ -50,7 +50,8 @@ body.is-edit-mode #toolbar-navigation-buttons {
   display: none;
 }
 
-#back-button, #forward-button {
+#back-button,
+#forward-button {
   opacity: 1;
   transition: none;
   padding: 0 0.6rem;
@@ -61,7 +62,8 @@ body.is-edit-mode #toolbar-navigation-buttons {
 #forward-button {
   margin-left: -0.75em;
 }
-#back-button:disabled, #forward-button:disabled {
+#back-button:disabled,
+#forward-button:disabled {
   opacity: 0.2;
 }
 
@@ -108,6 +110,10 @@ body.linux #menu-button {
   word-break: break-all;
   position: relative;
   padding: 0 0.8em;
+  display: flex;
+  justify-content: space-between;
+  align-items: stretch;
+  gap: 0.25rem;
 }
 .tab-item.gu-mirror {
   top: var(--control-space-top) !important;
@@ -153,14 +159,19 @@ body.linux #menu-button {
 /* icons */
 
 .tab-item .tab-icon-area {
-  float: right;
   margin-right: -0.7em;
+  display: flex;
+  align-items: center;
 }
 .tab-item .tab-icon {
   position: relative;
   font-size: 0.875em;
   padding: 0.7125em 0.5em;
-  box-sizing: content-box;
+  box-sizing: border-box;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100%;
 }
 .tab-item .tab-icon:not(.permission-request-icon):hover:after {
   content: "";
@@ -169,8 +180,6 @@ body.linux #menu-button {
   border-radius: 50%;
   background: rgba(0, 0, 0, 0.075);
   position: absolute;
-  top: -3px;
-  left: -3px;
   z-index: -1;
 }
 .dark-theme .tab-item .tab-icon:hover:after {
@@ -209,6 +218,7 @@ body.linux #menu-button {
 }
 .tab-item .title {
   font-size: 15px;
+  order: -1;
 }
 
 /* buttons */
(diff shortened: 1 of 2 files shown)


PalmerAL:
Thanks for the PR! It's odd things are so misaligned in your original screenshot - those buttons are mostly centered on macOS, and I recall them looking much better than that on Windows and Linux also. Switching to flexbox is a good solution though.

Two bugs I've noticed so far:

The hover effect on the tab close button is broken:

Screenshot 2023-04-25 at 11 58 50 PM

There's something weird going on with the layout of tabs with permission icons when selected:


https://user-images.githubusercontent.com/10314059/234474109-6e8cded2-62e4-41df-ac2c-d1f409db8fb2.mov

RubenSmn:
Hey thanks for the feedback! I'll get that fixed. 

Might be worth mentioning I'm using Linux: ZorinOS

RubenSmn:
@PalmerAL as you might be able to tell from the screenshots the permission icon changed a bit, its not fully round anymore. Should it still be round?

RubenSmn:
I've fixed the issues @PalmerAL.

- used `gap` for consistent spacing between items inside a `tab-item`
- title has 50% more spacing and does not 'touch' the right side
- used dynamic width for the back/forward button container as it was occupying unnecessary space

Let me know if there are anymore improvements I can make!

PalmerAL:
Sorry, two more things:

1. The spacing seems broken when the tab is wider than the title and there are icons:

Screenshot 2023-05-12 at 9 13 02 PM

Removing the `justify-content: space-between` on the tab item sort of fixes this, but then the close button is in the wrong place:
Screenshot 2023-05-12 at 9 18 11 PM


2. Previously, the forward button would appear with an animation - how difficult would it be to add this back?


https://github.com/minbrowser/min/assets/10314059/f0b3208c-2199-4c2a-baca-55473614be22

RubenSmn:
No need for any apology, thank you for finding these edge cases. I found 2 others myself

- spacing between tab close button and the the unsecure webpage indicator
- permissions indicator was not responsible for more than 1 permission e.g. camera and microphone

I've fixed the title spacing and the placement of the close button. Changed the fixed width of the navigation buttons to a more appropriate size (was occupying unnecessary space) which brings back the transition / animation.

PalmerAL:
This looks really good! I only see one minor issue left: the spacing on the sides of the tab item is a bit uneven, because the `gap` applies to the `tab-icon-area` even when no icons are showing:

Screenshot 2023-05-16 at 9 42 19 PM

This means an extra character or two of the tab title is hidden when it doesn't need to be. It's not a huge issue though, if you don't see a simple way of fixing it (I don't at the moment) I'm fine leaving it as-is.

RubenSmn:
Great! I was going to solve this by checking if the `tab-icon-area` had no children and then apply  the gap width as negative margin. But I found that it always has a `i` element as a secure indicator, so now wer'e checking if the area only contains the `i` element and then applying the style.

PalmerAL:
I don't think that works, because the icon area also contains the tab close button, but what do you think of the change in 1b679548bd5e3bd1b2136b843e2b98e3bd44ba57 ?

RubenSmn:
Ah okay, this works better but I'm having a issue with hovering. Might be something just on linux as I'm not seeing the close icon on your tabs and I assume they probably appear when hovering a tab.

[Kooha-2023-05-18-11:58:23.webm](https://github.com/minbrowser/min/assets/31519867/40d7602b-a8b5-4c01-9958-fd2fde0d40fd)

css/tabBar.css
```diff
@@ -153,14 +159,23 @@ body.linux #menu-button {
...
 
 .tab-item .tab-icon-area {
-  float: right;
-  margin-right: -0.7em;
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
+  order: 1;
+}
+.tab-item:not(:hover) .tab-icon-area:has(.tab-close-button:only-child) {
```
RubenSmn, line 167: I came up with this

```suggestion
.linux .tab-item .tab-icon-area:has(.tab-close-button:only-child),
:not(.linux)
  .tab-item:not(:hover)
  .tab-icon-area:has(.tab-close-button:only-child) {
```

css/tabBar.css
```diff
@@ -305,12 +323,9 @@ body.windows.dark-theme .navbar-action-button:not(:disabled):hover {
 /* in Carbon, close icon is a bit smaller than other icons, enlarge it in order to match with other icons */
 .tab-item .tab-close-button {
   font-size: 1.25em;
-  padding: 0.3em 0.45em;
-  margin-top: 0.15em;
+  box-sizing: border-box;
```
RubenSmn, line 326: Should the close icon not be the icon most to the right side?

image

```suggestion
  box-sizing: border-box;
  order: 1;
```

PalmerAL: Yes it should.

RubenSmn:
Yes I have a touch screen, the rule you've provided does the trick @PalmerAL!

I couldn't find anymore issues, let me know if you need any adjustments.

PalmerAL:
It all looks good to me! Thanks for all the revisions on this!

PR #2237, opened 2023-05-13
Multi-window support

PalmerAL:
This PR re-architects parts of the browser to support ([long-requested](https://github.com/minbrowser/min/issues/169)) multi-window support. In some places, this means modifying the code to pick the right window instead of assuming there's a single primary window, or modifying things like the history service to have a single global instance that can be shared across windows.

Most of the actual tab management is handled in [windowSync.js](https://github.com/minbrowser/min/blob/multi-window-2/js/tabState/windowSync.js); it listens for changes on the task/tab objects, and synchronizes those changes with the other windows using IPC. The result is that each window gets the same view into the tab state.

The task list is shared across windows, so each window acts as a different "view" of the task list. Opening a task in a window "steals" the task from any other window where it's currently open. Tabs stay in memory until the last window is closed (there should probably be a way to manually destroy them sooner; I plan to add this to the task overlay UI).

There's still a number of bugs with this that I'm working on solving, but I believe all of the core functionality should be working, and it's mostly usable without running into too many issues.


PalmerAL:
@RubenSmn If you have a chance to test this for a bit and see if you notice any obvious bugs before I merge it, that would be great. No worries if not though.

RubenSmn:
Switching to a task that is opened on another window you get a white screen (browser view) until you click with the mouse.

To reproduce:

- create 2 tasks with some tabs
- open each task on a different window
- `ctrl+[` to switch task on one of the windows

flightmansam:
This is so amazing! The day has come, I'll be using this over the next few days and see what I can see.

PalmerAL:
@RubenSmn Both should be fixed now. The switching task case is still a little weird - when you switch in the first window, the second window creates a blank third task and switches to that. In the simple case, it seems like the windows should just swap tasks, but in general I'm not sure how the second window should pick a new task from the task list.

PalmerAL:
I'm going to merge this so that I can make a beta release with it. However, if you see any other issues, please let me know!

PR #2239, opened 2023-05-17
Update to electron 22.3.9 & node 16

Alex313031:
- Update from Electron 22.0.0 to the latest 22.3.9. Why were we on old original 22.0.0 anyway?
 - Update nodejs from 15 to 16
 - Ensure devtools is always available by specifying in webpreferences

### DIFF
--- a/main/main.js
+++ b/main/main.js
@@ -313,15 +313,6 @@ function createWindowWithBounds (bounds) {
   return newWin
 }
 
-// Quit when all windows are closed.
-app.on('window-all-closed', function () {
-  // On OS X it is common for applications and their menu bar
-  // to stay active until the user quits explicitly with Cmd + Q
-  if (process.platform !== 'darwin') {
-    app.quit()
-  }
-})
-
 // This method will be called when Electron has finished
 // initialization and is ready to create browser windows.
 app.on('ready', function () {
--- a/main/windowManagement.js
+++ b/main/windowManagement.js
@@ -22,6 +22,11 @@ const windows = {
 
     window.on('closed', function() {
       windows.removeWindow(window)
+
+      // Quit on last window closed (ignoring secondary and hidden windows)
+      if (windows.openWindows.length === 0 && process.platform !== 'darwin') {
+        app.quit()
+      }
     })
   
     windows.nextId++


UnKnoWn-Consortium:
I think it does not hurt even to push it to the latest stable Electron 24. Yes the support for Windows 7, Windows 8, and Windows 8.1 will get dropped along the way as the underlying Chromium dropped them too but Min will get a long list of fixes and new features. Personally it is totally justified to have support dropped for unmaintained OSes as they are unmaintained after all.

Alex313031:
@UnKnoWn-Consortium I disagree, but I guess I am biased because I still use windows 7 alongside windows 10 and linux. 

Electron 13 and above breaks win7/8/8.1, as well as ubuntu 16.04, debian 8, and fedora <30. Because of this, I have been taking the approach of only updating my electron projects (or forks of other's projects) to versions above 22.x.x when absolutely necessary.

Also, updating electron versions is often not a trivial task of just updating the package.json.

You can see breaking changes here > https://www.electronjs.org/docs/latest/breaking-changes

Min might be using some of these API's.

UnKnoWn-Consortium:
@Alex313031 I can totally understand the urge to support older OSes. They are still in use and sometime they are more appropriate for some hardware. Just you can call me paranoid but I am not sure if it is a best practice for a browser that surfs the open web.

PalmerAL:
Because Min displays external web content, we need the Chromium feature additions and security fixes in newer versions. Staying on an older version isn't an option.

I typically update to the latest version before each new Min release; there hasn't been a new release in a while, but I'll make one soon.

It is unfortunate that Chromium is dropping older OSs. For that case, I would recommend just using an older version of Min; you won't get the latest UI changes, but that's a better option than keeping everyone on an old Chromium version.

PalmerAL:
Can you explain what you mean by "Ensure devtools is always available by specifying in webpreferences"? It should already be available from the developer menu.

PalmerAL:
Although in this case, it looks like we do have an issue upgrading past 22, potentially requiring a new tab bar design on macOS: https://github.com/electron/electron/issues/37789

PalmerAL:
Alternative upgrade PR, let me know what you think: https://github.com/minbrowser/min/pull/2250

PR #2245, opened 2023-05-25
Fix #2244

### LINKED ISSUE #2244: if i try to open a link from my history by pressing ctrl , it doesn't open in new tab

Sestowner:


js/searchbar/searchbar.js
```diff
@@ -52,7 +52,8 @@ var searchbar = {
       return
     }
 
-    if (event && event.metaKey) {
+    var platformModKey = (window.platformType === 'mac') ? event.metaKey : event.ctrlKey
```
PalmerAL, line 55: This will fail if `event` is undefined (it's an optional argument).

PalmerAL:
Looks good.

PR #2246, opened 2023-05-25
feat(bitwarden): added field for server url during setup for self-hosted servers and others

UnKnoWn-Consortium:
feat(bitwarden): added field for server url during setup for self-hosted servers
fix(bitwarden): fixed confusing cache logic that does not cache credential request responses properly
refactor(prompt): retrofitted the `createPrompt` function with promise
refactor(prompt): updated `createPrompt` usage for the promise retrofit
refactor: unwrapped AutofillSetup as individual functions and exported the `initialize` function that is used externally
refactor: unwrapped PasswordManagers as individual functions and exported functions that are used externally
refactor: replaced some usages of `var` with `let` and `const`


PalmerAL:
This seems reasonable from reading over it; can you explain "fixed confusing cache logic that does not cache credential request responses properly"?

UnKnoWn-Consortium:
@PalmerAL In the original code, the cache list `lastCallList` for a domain is set to a promise instead of the suggestion returned from the password manager. https://github.com/minbrowser/min/blob/5bd1544eb1bbd4ea7913d25da26f4395bc177d85/js/passwordManager/bitwarden.js#L87

UnKnoWn-Consortium:
@PalmerAL Thanks for bringing us multi-window support. I will resolve the conflict asap.

UnKnoWn-Consortium:
@PalmerAL Sorry to ping. Conflict resolved now.

PalmerAL:
Sorry, I haven't had time to test this. Will do in the next few days.

UnKnoWn-Consortium:
@PalmerAL Sorry my fault. Please take your time.

js/passwordManager/bitwarden.js
```diff
@@ -161,41 +153,43 @@ class Bitwarden {
       console.warn(e)
     }
 
-    // show credentials dialog
-
-    var signInFields = [
+    // show ask-for-credential dialog
+    const signInFields = [
+      { placeholder: 'Server URL (Leave blank for the default Bitwarden server)', id: 'url', type: 'text' },
```
PalmerAL, line 158: This should probably be a translated string - add it to `en-US.json`, then do `l(stringName)` like the examples below.
Actually, maybe "client ID" and "Client Secret" should be translated also, although I think the reason not to translate them was so that it would be clear in Bitwarden what we're referring to.

js/passwordManager/bitwarden.js
```diff
@@ -161,41 +153,43 @@ class Bitwarden {
...
-        throw new Error('no credentials entered')
+    const credentials = ipcRenderer.sendSync(
+      'prompt',
+      {
+        text: l('passwordManagerBitwardenSignIn'),
+        values: signInFields,
+        ok: l('dialogConfirmButton'),
+        cancel: l('dialogSkipButton'),
+        width: 500,
+        height: 260
```
PalmerAL, line 171: The dialog is a little bit too short now, at least on macOS:

js/passwordManager/passwordManager.js
```diff
@@ -9,126 +9,122 @@ const Bitwarden = require('js/passwordManager/bitwarden.js')
...
-          return
-        }
-
-        if (!manager.isUnlocked()) {
-          await PasswordManagers.unlock(manager)
-        }
-
-        var formattedHostname = hostname
-        if (formattedHostname.startsWith('www.')) {
-          formattedHostname = formattedHostname.slice(4)
```
PalmerAL, line 102: This breaks autofill on sites starting with `www.`, because we pass `hostname` in the autofill response, and it then makes this check fail:

https://github.com/minbrowser/min/blob/323e5dff0185be09a53dd558f891a5bdb373cfbe/js/preload/passwordFill.js#L273-L275

(We should keep the check, it's guarding against us accidentally sending IPC messages to the wrong tab)

js/passwordManager/bitwarden.js
```diff
@@ -161,41 +153,43 @@ class Bitwarden {
...
+
+    if (credentials.clientID === '' || credentials.clientSecret === '') {
+      throw new Error('no credentials entered')
     }
 
-    const process = new ProcessSpawner(path, ['login', '--apikey'], {
-      BW_CLIENTID: credentials.clientID.trim(),
-      BW_CLIENTSECRET: credentials.clientSecret.trim()
-    })
+    credentials.url = credentials.url || 'bitwarden.com'
```
PalmerAL, line 179: It wasn't clear to me whether I was supposed to type "bitwarden.com" or "vault.bitwarden.com" - do they both work? If not, maybe we should special-case this so if the user enters one we convert to the other.

PalmerAL (changes requested):
Looks good other than comments! I couldn't find a custom server to test this with (and didn't want to spend the time setting one up), but everything seems to work with the default server.

Issue #630, opened 2018-09-29
[Windows] - Circa 10px strip above tabs


Anavnaj:
I am absolutely in love with this browser - yet, I can't use it. The unnecessary strip of empty pixels in the upper section makes the user interface extremely hard to use. Using any of the user elements in the upper part (which is all of them) now requires quite a precise mouse work, which really breaks ones workflow.

If I could truly pin the windows to the top of screen so that I could just move to mouse to the upper edge and still be able to click that, that would be all I need.

In other words - please, get rid of the strip of empty pixels above the buttons, tabs and other user elements.
Thank you and keep up the great work.

Anavnaj:
I'm adding an image just to make myself perfectly clear about the issue.

![pixelstrip](https://user-images.githubusercontent.com/43697644/46244252-79e3e500-c3dc-11e8-83b7-404329f2aa06.png)

PalmerAL:
> A) Make the space to the right of the tabs (or, rather, to the left of the "+" button, or perhaps even of the "minimize window" button, as seen in the linked image) really small - I don't see a reason why it should be wider that 3-4 times the width of the cursor. This will not limit the space to see the open tabs significantly...

You mean something like this (where the red is the draggable area)?

<img width="777" alt="screen shot 2018-10-02" src="https://user-images.githubusercontent.com/10314059/46330020-6dc37780-c5d6-11e8-80a9-dc3c445eabd6.png">

That actually works better than I would have expected, although it does look a little bit odd to have a space between the two buttons. 

> B) Have the upper strip of pixels appear only when the browser window is not maximized. When the browser window is maximized, there's no real need for dragging it, right?* I mean - doing so "unmaximizes" it anyway.

This seems pretty reasonable to me, although the snapping thing you mentioned may be an issue. I also wonder if it would be confusing whether the window is actually maximized (in which case you would need to click the button first), or whether it was almost-but-not-quite maximized (in which case you shouldn't click the button and need to drag from the top instead).

Another thing I wonder about is whether not having the tabs at the top is actually the issue, or whether it's just the position of the draggable area and the height of the tabs. Currently, the draggable area extends a little bit lower than you would guess:

<img width="805" alt="screen shot 2018-10-02 2" src="https://user-images.githubusercontent.com/10314059/46330119-e7f3fc00-c5d6-11e8-9a41-31e00391c8c7.png">

(In particular, the bottom of the drag region extends a little bit past the top of the window buttons).
[...]

PR #2250, opened 2023-05-31
Modify Mac draggable region for Electron 24

PalmerAL:
In Electron 24, we can't have a region that is both draggable and clickable: https://github.com/electron/electron/issues/37789. This makes it impossible to drag the window on macOS with the existing titlebar design.

Windows and Linux currently have the same limitation; we solve it there by shifting the titlebar down and adding empty space above (see previous discussion: https://github.com/minbrowser/min/issues/630). But no other apps do the same thing on macOS, so it would look odd to do it there.

My current idea (which this PR implements) is to do this, where the draggable regions are in red:

Screenshot 2023-05-30 at 10 19 03 PM

Although the space to the right of the window controls looks slightly odd, and it's not obvious what's draggable:

Screenshot 2023-05-30 at 10 22 19 PM

If anyone has feedback or other ideas, please let me know!

### DIFF
--- a/css/windowControls.css
+++ b/css/windowControls.css
@@ -27,18 +27,19 @@ body.linux:not(.maximized):not(.fullscreen):not(.separate-titlebar):not(.disable
   -webkit-app-region: drag;
 }
 
-body.mac:not(.separate-titlebar) .window-drag-area {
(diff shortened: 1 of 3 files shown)


PalmerAL:
I'm going to merge this and make a beta release to get more feedback, and then we can adjust it again before the final release.

PR #2285, opened 2023-07-21
Remove "#pdfjs.action=download"

### LINKED ISSUE #2208: Trouble with "Save Link As..."

Sestowner:
URLs that contain "#pdfjs.action=download" was opening Electron's built-in pdf viewer.

To reproduce, open https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf#pdfjs.action=download

This also fixes #2208.

### DIFF
--- a/main/download.js
+++ b/main/download.js
@@ -11,8 +11,6 @@ function isAttachment (header) {
 }
 
 function downloadHandler (event, item, webContents) {
-  var itemURL = item.getURL()
-  var attachment = isAttachment(item.getContentDisposition())
   const sourceView = Object.values(viewMap).find(view => view.webContents.id === webContents.id)
   let sourceWindow
   if (sourceView) {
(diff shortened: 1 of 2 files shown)


PalmerAL:
Looks good, thanks!

(I was wondering why the download button in the viewer still works with this change. The reason is that the request made when clicking that has `details.resourceType: 'other'`).

Issue #2278, opened 2023-07-08
New window option when right clicking on dock icon


devgocri:
Hey there, window support is working great so far! Just a suggestion, it would be nice to have a "New Window" option in the menu when you right click on dock icon, because currently I have to switch to the new virtual desktop (where I want the second window), right click on the icon, click "New Tab", then switch to the right task, delete the old task that the new window created...

A menu when the new window option is clicked would also be great when opening from dock, maybe a "Which task would you like to open" or something that can be toggled off in settings if need be.

<img width="177" alt="image" src="https://github.com/minbrowser/min/assets/70841501/d21e95f9-0f58-40b2-b11d-85c69c458677">

PalmerAL:
Thanks! Good point; I honestly just forgot to update the dock menu since I don't use it very often.

I wonder if we should just list the tasks directly in the dock menu? I'm not sure how I would expect that to work though - whether it would switch to the selected task in the current window or open a new one. Thoughts?

devgocri:
Hm... if the tasks would be listed in the dock menu I'd usually assume it would open in a new window, as switching tasks in current window is quite easy to do/faster through the browser window, and there would be no point in someone going to the dock to do it. 

Listing the tasks in the dock might lead to an excessively large dock menu though if the user has a lot of tasks - it might be a pain to find what you want/navigating through a huge list...

Maybe a better idea would be to have a "New Window" option in the dock, and then automatically opening the task list in the new window that you would usually access using the icon on the top right. That would then allow the user to select which task they want or they can create a new task without having to clutter their dock menu.

devgocri:
Here's what this might look like: (I use the menu bar here but you should be able to do this through the dock)

https://github.com/minbrowser/min/assets/70841501/9ca9d661-4acb-47a2-b420-d7ee710a0b37

MohitBansal321:
@PalmerAL 
I guess its already added now time for close this issue?

Kepler-69c:
Could someone confirm that the feature also works on Windows and Linux?

Sadi58:
Unfortunately, it doesn't work under **Debian 13 Trixie/testing** with **KDE Plasma 5.27**.
AFAIK, for this, the file `min.desktop` should contain a line like `Actions=new-window;` followed by a  section like this:
```
[Desktop Action new-window]
Name=Open New Window
Exec=/opt/Min/min --new-window %U
Icon=min
```
[...]

PR #2303, opened 2023-08-20
Add option to show task overlay when opening a window

PalmerAL:
@devgocri This implements your suggestion in #2278 (I've added the dock menu option in f7535a6f307fe05c4486098faa75572e4c17d444). 

This will still create a new blank task when the window opens; the reason is that the rest of the code assumes that the window always has some task selected in it, so we have to make a new task to give it something to select. This blank task will get reused across multiple window opens though, so the list won't get too cluttered.


devgocri:
Hey @PalmerAL!

Just tested it - it works great! One issue though, if you close all open windows of Min, right click on the dock icon and select "New Window", it creates a new window but does not open task overlay as I've selected in settings. If I've already got a window of Min open however, it opens the task overlay as intended.

I'm really sorry for the late reply :C

PalmerAL:
Thanks!

> One issue though, if you close all open windows of Min, right click on the dock icon and select "New Window", it creates a new window but does not open task overlay

Fixed.

> Also, spam opening multiple new windows create a whole bunch of new tasks instead of re-using existing tasks.

If you have multiple new windows, they all need their own task, since you can only have each task open in a single window. But if you make a window, select some existing task, then make a new window, it should reuse the blank task that was created for the first one.

devgocri:
@PalmerAL Tested, works perfectly :D

PalmerAL:
Thanks!

Issue #2283, opened 2023-07-20
Paste and match style


devgocri:
Hey I would really love it if you could add the "Paste and match style" which the other browsers have :D

<img width="242" alt="image" src="https://github.com/minbrowser/min/assets/70841501/a5cb607e-b867-4720-8dc7-b882d16935b7">

sndrkrly:
What is this for?

devgocri:
@sndrkrly It's basically for pasting clipboard content without formatting.

For example, if you're copying and pasting a ChatGPT response into Google Docs, if you paste normally, it will paste with the grey background, a different font, font size and color. If you use paste and match style instead, it will use the font, color and size the text in the document already has.

PR #2315, opened 2023-09-09
added "Paste and Match Style"

sndrkrly:
### **Issue #2283**

I can't find where is the "right-click" menu, so 
it's not listed there yet, only on the "topbar" (or what is it called).


main/menu.js
```diff
@@ -212,6 +212,11 @@ function buildAppMenu (options = {}) {
           accelerator: 'CmdOrCtrl+V',
           role: 'paste'
         },
+        {
+          label: l("pasteAndMatchStyle"),
+          accelerator: "", /* change this to some accelerator */
+          role: "pasteAndMatchStyle"
```
sndrkrly, line 218: Issue #2283

localization/languages/en-US.json
```diff
@@ -23,6 +23,7 @@
     "cut": "Cut",
     "copy": "Copy", //this is a verb (as in "copy the currently-selected text")
     "paste": "Paste",
+    "pasteAndMatchStyle": "Paste and Match Style",
```
PalmerAL, line 26: Can you also add this line to the other language files, so the translators know to add it?

```
    "pasteAndMatchStyle": null, //missing translation
```

main/menu.js
```diff
@@ -212,6 +212,11 @@ function buildAppMenu (options = {}) {
           accelerator: 'CmdOrCtrl+V',
           role: 'paste'
         },
+        {
+          label: l("pasteAndMatchStyle"),
+          accelerator: "", /* change this to some accelerator */
```
PalmerAL, line 217: This opt+shift+cmd+v on macOS, so `alt+shift+CmdOrCtrl+v"? Not sure what it is on Windows though.

PalmerAL (commented):
Thanks!

The right-click menu is `js/webviewMenu.js`. There's also `js/contextMenu.js`, but that's really only used in the address bar, which doesn't accept styled text anyway, so I don't think it makes sense to add there.

main/menu.js
sndrkrly: I also use Mac so I don't know, but I will look for the "alternative" for windows

PalmerAL:
This was completed in #2376.

PR #2319, opened 2023-09-14
fix 1Password autofill not working

nekhaevskiy:
in case if users do not have `op` in their $PATH


PalmerAL:
Looks good, thanks! Sorry for the slow response.

PR #2362, opened 2023-12-28
Improve the process of writing userscripts

PalmerAL:
* Enabling userscripts now creates the directory automatically, and I've added a button from the settings page to open it:
Screenshot 2023-12-27 at 11 12 50 PM

* Userscripts now reload when edited, so that only the webpage needs to be reloaded rather than restarting the browser.

If anyone has feedback on this, let me know! I'd like for userscripts to be more accessible and easier to set up, so I'd be interested in hearing further ideas to make that happen.

### DIFF
--- a/js/preload/default.js
+++ b/js/preload/default.js
@@ -50,7 +50,11 @@ window.addEventListener('message', function (e) {
     return
   }
 
-  if (e.data && e.data.message && e.data.message === 'showCredentialList') {
+  if (e.data?.message === 'showCredentialList') {
     ipc.send('showCredentialList')
+  }
+
+  if (e.data?.message === 'showUserscriptDirectory') {
+    ipc.send('showUserscriptDirectory')
   }
 })
--- a/js/userscripts.js
+++ b/js/userscripts.js
@@ -1,6 +1,7 @@
 /* implements userscript support */
 
 var path = require('path')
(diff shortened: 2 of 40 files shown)


wong-justin:
I like these changes. I think the only improvements would be in documentation. The ["learn more" wiki page](https://github.com/minbrowser/min/wiki/userscripts) is a nice entry-level summary, but it doesn't appear at the top of web search results, and it doesn't document the userscript header thoroughly (`@match` syntax, `@run-at` options, etc).

Some suggestions:

- change "learn more" to something descriptive like "see userscript documentation"
- link to the userscripts wiki page from a central location like the [readme](https://github.com/minbrowser/min#readme) or the [website](https://minbrowser.org/)
- in the userscripts wiki page, document the syntax more thoroughly, or link to a more comprehensive reference like [Violentmonkey API docs](https://violentmonkey.github.io/api/metadata-block/).

PalmerAL:
@wong-justin Good points!

1. I think I'm going to leave this as-is, only because this string is translated into a bunch of languages and it'd have to be re-translated.
2. Done: a95538bf730c29fb24c517ac3d44b4217f473d56
3. Unfortunately, I think our syntax isn't totally compliant with Violentmonkey or any other spec; we should probably address that first (always open to contributions on this :) )

PR #2370, opened 2024-01-19
Switch to quick-score for fuzzy search (leftover)

tobiabocchi:
I think you might have left this behind when you were working on [this PR](https://github.com/minbrowser/min/pull/2342) I noticed it because the fuzzy finding functionality using `!task` was not working anymore. I also noticed in that PR you changed the score target from .4 to .35, I was not sure wether it was necessary here so I did not do it for now. One last thing I am not sure that the `require` statement is in the right place but if it isn't I can move it where you prefer.
This is my first time contributing to Min I hope I did everything according to your requirements.

### DIFF
--- a/js/searchbar/customBangs.js
+++ b/js/searchbar/customBangs.js
@@ -15,6 +15,7 @@ const bookmarkConverter = require('bookmarkConverter.js')
 const searchbarPlugins = require('searchbar/searchbarPlugins.js')
 const tabEditor = require('navbar/tabEditor.js')
 const formatRelativeDate = require('util/relativeDate.js')
+const quickScore = require('quick-score').quickScore
 
 function moveToTaskCommand (taskId) {
   // remove the tab from the current task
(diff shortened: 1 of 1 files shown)


PalmerAL:
Thanks! You're right, I forgot to change this.

This file seems to be sorted with external dependencies followed by other files in the project, so I guess the import should go after `fs`. It doesn't matter too much though.

The score threshold is arbitrary; in that PR, I tested some different values with my own history database and `0.35` seemed to produce the best results. If you try changing it here, do the results seem better with one value or another?

tobiabocchi:
I moved the dependency up, I tried both `.35` and `.4` and it did not seem to make any difference to me, maybe you can try it out and see which one works best for you or give me some tasknames and search patterns to try out

PalmerAL:
I agree, it seems fine as is, let's go with it.

Thanks for fixing this!

PR #2413, opened 2024-03-28
Improved Brazilian Portuguese localization

GabrielDasNevesPinheiro:
added translations in null fields, translations based on the en-US.js file in the localization folder.


PalmerAL:
Thank you!

PR #2415, opened 2024-04-03
Improve Korean translation.

Edward-Lucas:
Improve Korean translation.
Corrected incorrect words and grammar and made it easier to understand.

### DIFF
--- a/localization/languages/ko.json
+++ b/localization/languages/ko.json
@@ -7,76 +7,76 @@
         */
     "addToDictionary": "사전에 추가",
     "pictureInPicture": "화면 속 화면(Picture in Picture)",
-    "openInNewTab": "새 작업",
-    "openInNewPrivateTab": "새 개인 작업",
-    "saveLinkAs": "다른 이름으로 저장",
-    "viewImage": "그림 보기",
-    "openImageInNewTab": "새 작업에서 그림 열기",
-    "openImageInNewPrivateTab": "새 개인 작업에서 그림 열기",
+    "openInNewTab": "새 탭",
+    "openInNewPrivateTab": "새 시크릿탭",
+    "saveLinkAs": "링크를 다른 이름으로 저장",
+    "viewImage": "이미지 보기",
+    "openImageInNewTab": "새 탭에서 이미지 열기",
+    "openImageInNewPrivateTab": "새 시크릿탭에서 이미지 열기",
     "saveImageAs": "그림을 다른 이름으로 저장",
     "searchWith": "%s에서 검색", //%s will be replaced with the name of the current search engine
-    "copyLink": "주소 복사",
-    "copyEmailAddress": "전자우편주소 복사",
-    "selectAll": "전체선택",
+    "copyLink": "링크 복사",
+    "copyEmailAddress": "이메일 복사",
+    "selectAll": "모두 선택",
     "undo": "되돌리기",
     "redo": "재실행",
-    "cut": "오리기",
-    "copy": "복사하기", //this is a verb (as in "copy the currently-selected text")
+    "cut": "자르기",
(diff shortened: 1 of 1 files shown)


localization/languages/ko.json
```diff
@@ -97,53 +97,53 @@
     "timeRangeWeek": "지난 주",
     "timeRangeMonth": "지난 달",
     "timeRangeYear": "작년",
-    "timeRangeLongerAgo": "오래전",
+    "timeRangeLongerAgo": "호랑이 담배 피던 시절",
```
PalmerAL, line 100: If I understand correctly, this is an idiomatic expression - would a more literal statement be better here?

localization/languages/ko.json
```diff
@@ -97,53 +97,53 @@
...
     "settingsBlockScriptsToggle": "명령(스크립트) 차단",
-    "settingsBlockImagesToggle": "그림 차단",
+    "settingsBlockImagesToggle": "이미지 차단",
     "settingsBlockedRequestCount": {
-      "unsafeHTML": "지금까지 민(Min)은 <strong></strong> 광고 및 추적기를 차단했습니다."
+      "unsafeHTML": "지금까지 차단된 광고 및 추적기: <strong></strong>"
     },
-    "settingsCustomBangs": "사용자 정의 연결 명령",
-    "settingsCustomBangsAdd": "사용자 정의 연결 명령 추가",
+    "settingsCustomBangs": "Custom Bang Search",
```
PalmerAL, line 145: If it's not possible to translate this, can you leave it as "custom bangs" as in the English version?

localization/languages/ko.json
```diff
@@ -234,48 +234,48 @@
...
+    "articleReaderView": "돌아가기",
     /* Download manager */
     "downloadCancel": "취소",
-    "downloadStateCompleted": "완료됨",
-    "downloadStateFailed": "실패함",
+    "downloadStateCompleted": "완료",
+    "downloadStateFailed": "실패",
     /* Update Notifications */
-    "updateNotificationTitle": "새롭게 판올림된 민(Min) 사용 가능",
+    "updateNotificationTitle": "업데이트가 필요합니다.",
```
PalmerAL, line 261: Google/deepl both translate this as "An update is required". Can you make sure it matches the original English ("A new version of Min is available")? The user isn't required to install the update.

PalmerAL (commented):
Thanks! Left a few comments.

localization/languages/ko.json
Edward-Lucas: Korea enjoys including funny phrases in its programs.
The phrase has exactly the same meaning as it did a longer ago. this will be an interesting point.
I prefer not to edit it, but if you want to edit it, I'll accept it.

localization/languages/ko.json
Edward-Lucas: OK, I will edit it.

localization/languages/ko.json
Edward-Lucas: "업데이트가 필요합니다." is close to meaning that it's absolutely necessary.
If an update is not essential, the "새 버전의 민(Min)이 존재합니다." is appropriate.

localization/languages/ko.json
PalmerAL: Let's go with the boring version, it's important to have consistency between the different languages the program can run in. Also, I asked a friend who speaks Korean for their opinion, and they agreed the boring version was better.

localization/languages/ko.json
PalmerAL: I think it's better as an informational statement. It's good if the user updates, but they don't have to immediately.

Edward-Lucas: I am new to GitHub.... I accidentally created a new pull request. 
I've edited out everything you said.

PR #2430, opened 2024-04-27
enable Github Actions build with os matrix

alswl:
This issue enables GitHub Actions to build with an OS matrix and automatically generate a release draft.
Upon completion of the pipeline, a release will be drafted automatically.

The workflow operates similarly to [this example](https://github.com/alswl/min/actions/runs/8861128186).
The release is similarly to [this](https://github.com/alswl/min/releases/tag/v1.32.0-fork).


PalmerAL:
Thanks! I think the matrix changes look good.

I'm a little bit skeptical about importing a third-party action that has permissions to directly upload releases. Is there a way to scope GITHUB_TOKEN so that it can only publish draft releases? If not, I'd probably prefer to keep uploading the artifacts manually.

alswl:
Hi @PalmerAL ,

For a better understanding of `GITHUB_TOKEN` and its permissions in GitHub Actions, you can refer to the [official documentation](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret). It outlines how the token is used and what permissions it has by default.

The grant feature seems to be tailored for workflows, allowing you to manually adjust the permissions for specific pipelines as detailed [here](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token).

Additionally, `svenstaro/upload-release-action` is a popular action that operates at the repository level only. Given this scope limitation, using the token with this action should be secure.

alswl:
@PalmerAL ping.

alswl:
@PalmerAL
Is this request applicable to this project. If it is feasible, I will resolve the conflict, otherwise I will close this request.

PalmerAL:
Sorry, I forgot about this. I don't think github_token supports a fine-grained enough scope to limit to only draft releases (assuming this is still true: https://stackoverflow.com/questions/67389957/what-permissions-does-github-token-require-for-releases-from-a-github-action). So I don't want to move forward with that part.

The reorganization of the action seems good, so up to you if you want to go forward with that.

alswl:
@PalmerAL OK, I removed the drafting release action. Please review my pull-request again.

.github/workflows/build-packages.yml
```diff
@@ -1,54 +1,71 @@
...
       with:
-        node-version: 20
-    - run: npm install --omit=optional
-    - name: build package
-      run: npm run buildDebian && npm run buildRedhat && npm run buildLinuxArm64 && npm run buildRaspi
-    - name: upload package
-      uses: actions/upload-artifact@v4
+          name: windows-package
+          path: dist/app
+    - uses: actions/download-artifact@v4
```
PalmerAL, line 64: I think this part can be removed too, right?

alswl: OK, the sections is removed.

PalmerAL:
Thanks!

Issue #2444, opened 2024-05-23
Import bookmarks file not working


Silver25:
Hi Palmer,

I've been using browser Min 1.32.0 for a few weeks now (both Win 10 and 11).
Till then I didn't even know about this browser.
My main use of a lightweight browser is to play music from 1000oldies.de, love that,
and simple surfing/browsing, opening pages with articles I need at the moment.
Was a really nice surprised when finding that Min can start temp WordPress setups
on "zipwp.com" and "wpsandbox.net".
I can say that I really love your browser.
As you find in Title of this msg I can't import bookmark files from Edge, Chrome or Brave.
The second issue is not possible to open a local file(drag and drop or with address c: c:/ c:\ or file:///C:)

Appreciate any help or information.
Greetings!
Silvio

goesbyabhi:
Same here. Currently on Windows 11. I've followed the instructions, but it just can't import the bookmarks.

PalmerAL:
@Silver25 Thanks for reporting! @goesbyabhi has fixed the bookmark issue.

>  The second issue is not possible to open a local file(drag and drop or with address c: c:/ c:\ or file:///C:)

Do you see an error message of any kind when you do this? The `file:///C:` variant works for me.

Silver25:
@goesbyabhi I don't get any info/error, the page is completely blank.

AstringentBeet:
Currently using version 1.32.1 on MacOS (intel), and am experiencing the same issue.

PalmerAL:
@AstringentBeet I just published a new release that includes the import bookmark fix, sorry for the delay.

@Silver25 I don't have any further ideas regarding the local file issue unfortunately.

PR #2445, opened 2024-05-25
Fix: bookmarks data not being updated after importing

goesbyabhi:
On Windows (not sure about other OS), users are not able to import their old bookmarks from other browsers like Edge (haven't tried on Firefox). 
Ref #2444 

This patch fixes that issue. Now I can import and search through my bookmarks

image


PalmerAL:
Ah, thank you!

(This was an oversight of mine from 592da4f).

PR #2461, opened 2024-07-07
Implement "generate password" option for signup forms

PalmerAL:
https://github.com/minbrowser/min/assets/10314059/c1bfe2ee-4ac1-4ed5-83ad-9596491c4b9f

PR #2462, opened 2024-07-07
Update uk.json & ru.json

Sestowner:


PalmerAL:
Thanks!

I'm about to add another string in this PR also: #2461

PR #2468, opened 2024-07-22
Added a report button for reporting bugs on the error page

Tedd-Codes:
Adds a report button for reporting bugs/errors on the error page. the button links to the issue page but it night be better to make a web-page for report bugs that reports it on GitHub. it would be more user friendly but not needed. But anyway i added a report button on the error page.

### DIFF
--- a/pages/error/index.html
+++ b/pages/error/index.html
@@ -10,13 +10,14 @@
     <div id="page-wrapper">
       <h1 id="error-name"></h1>
       <h2 id="error-desc"></h2>
+
       <button id="secondary-button" class="secondary-button" hidden></button>
       <button
         id="primary-button"
         class="primary-button"
         data-string="errorPagePrimaryAction"
       ></button>
-      <button id="secondary-button" class="secondary-button" onclick="window.location.href='https://github.com/minbrowser/min/issues';">Report Bug</button>
+       <button id="secondary-button" class="secondary-button" onclick="window.location.href='https://github.com/minbrowser/min/issues';">Report Bug</button>
     </div>
 
     <script src="../../js/util/settings/settingsContent.js"></script>
--- a/pages/error/index.html
+++ b/pages/error/index.html
@@ -10,13 +10,13 @@
     <div id="page-wrapper">
       <h1 id="error-name"></h1>
       <h2 id="error-desc"></h2>
-
       <button id="secondary-button" class="secondary-button" hidden></button>
       <button
         id="primary-button"
         class="primary-button"
         data-string="errorPagePrimaryAction"
       ></button>
+      <button id="secondary-button" class="secondary-button" onclick="window.location.href='https://github.com/minbrowser/min/issues';">Report Bug</button>
     </div>
 
     <script src="../../js/util/settings/settingsContent.js"></script>


PalmerAL:
This seems like a good idea generally, but most of the times that the error page appears, it's an issue with the user's network or the website they're trying to visit; we don't want bug reports for those. There's a full list of error types in the error.js file; maybe we could go through those and pick some that are more likely to be the browser's fault, and only show the report button for those?

Also, the button label needs to be added to the translation system so it can appear in the user's language; you would do a find-and-replace action to add it to the language files, like this PR: https://github.com/minbrowser/min/pull/2461/files, and then set a `data-string` attribute like the button above the one you inserted.

Tedd-Codes:
Is there a error page for browser errors? If not how can we add it? Where i can put the report button. I know that most of the error codes are internet/server related but i thought that some of them might be a error witch the browser. For example if someone couldn't connect to any websites but the internet is fine it might be a browser error . But maybe there should be a report button in the menu? or somewhere people can find it when there is a browser error. Im not super familiar with the way this browse is coded so if you could make a test page with some buttons and integrations from the browser (like showing verrsion # or a quit button) so i can understand how the html connects with the browser that would be great. if not could you explain how those elements of the browser connect with the pages then i can code in some of the features that people want

PalmerAL:
> I know that most of the error codes are internet/server related but i thought that some of them might be a error witch the browser

Yeah, I agree. The question is, which ones might be browser errors? For example, if the error is "no internet connection", there's like a 0.01% chance it's our fault, and a 99.99% chance it really is a problem with the user's internet connection. But if the error is "UNEXPECTED_SECURITY_LIBRARY_STATUS" (https://github.com/minbrowser/min/blob/master/pages/error/error.js#L190C12-L190C46) that's probably our fault. We have this object that maps error codes to the text that we show the user: https://github.com/minbrowser/min/blob/master/pages/error/error.js#L261. So I think we could add an additional property there that indicates for a particular error code, whether we should show the button to the user.

> But maybe there should be a report button in the menu?

There is :)

Screenshot 2024-07-23 at 7 45 23 PM

> Im not super familiar with the way this browse is coded so if you could make a test page with some buttons and integrations from the browser (like showing verrsion # or a quit button) so i can understand how the html connects with the browser that would be great. if not could you explain how those elements of the browser connect with the pages then i can code in some of the features that people want

Start with this document: https://github.com/minbrowser/min/wiki/Architecture. And if you run into questions reading it, let me know; I know it's a bit complicated and have been looking for feedback on improving it.

Tedd-Codes:
@PalmerAL So, For example for Spanish, Report Bug would be "Reportar un error" and in the code it would be
```
    /* Error page Report Button */
    "Report Bug": "Reportar un error"
```
Now in the html how would i do it??
I see this in the documentation but i dont know how to use it?
```
Localization system (Main, UI, content)
The strings for each language are located in the localization/languages/ directory. If you add a new string, you'll need to add it to each language file (find and replace is useful for this).
Running npm run buildLocalization generates a dist/localization.build.js file that can be loaded in any page.
Once this file is loaded, you can call l(stringName) anywhere to get a localized string in the user's current language.
There are also some special attributes you can use to automatically load strings inside HTML files (however, these will only evaluated once when the document is loaded);
data-string sets the element's text content
data-label sets the element's title attribute
data-value sets the element's value attribute
```

I think adding examples to the documentation would help alot, because instructions are good but instructions with examples are GOOD!!!!!

PalmerAL:
Take a look at the line in erorr/index.html above your change:

```
<button
        id="primary-button"
        class="primary-button"
        data-string="errorPagePrimaryAction"
      ></button>
```

You set `data-string` equal to the ID of the string in the JSON file, and the localization script fills in the content.

Good point regarding an example, I'll add one.

PR #2485, opened 2024-08-20
add Thai support translation

kingprapop:
Request to add Thai language in translation function.


PalmerAL:
Good point! I've updated the list to include all the languages that the API supports in 217b46560dc13e9509e6612b994fb76fdae80357 (including Thai).

PR #2490, opened 2024-09-05
Added a default zoom level setting for new tabs

romantelychko:
Added a default zoom level setting for new tabs.
Fixed some IDs on the settings.html page.

### DIFF
--- a/js/webviewGestures.js
+++ b/js/webviewGestures.js
@@ -1,4 +1,5 @@
 var webviews = require('webviews.js')
+var settings = require( "./util/settings/settings" );
 
 var webviewGestures = {
   showBackArrow: function () {
@@ -39,7 +40,8 @@ var webviewGestures = {
     return this.zoomWebviewBy(tabId, -0.2)
   },
   resetWebviewZoom: function (tabId) {
-    webviews.callAsync(tabId, 'zoomFactor', 1.0)
+    const defaultZoomLevel = parseFloat( settings.get('defaultZoomLevel') || '1' );
+    webviews.callAsync(tabId, 'zoomFactor', defaultZoomLevel)
   }
 }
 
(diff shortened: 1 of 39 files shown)


js/webviewGestures.js
```diff
@@ -39,7 +40,8 @@ var webviewGestures = {
     return this.zoomWebviewBy(tabId, -0.2)
   },
   resetWebviewZoom: function (tabId) {
-    webviews.callAsync(tabId, 'zoomFactor', 1.0)
+    const defaultZoomLevel = parseFloat( settings.get('defaultZoomLevel') || '1' );
```
PalmerAL, line 43: It should be possible to store the setting value directly as a number, so we wouldn't need to parse the string.

js/webviews.js
```diff
@@ -51,6 +51,16 @@ function onNavigate (tabId, url, isInPlace, isMainFrame, frameProcessId, frameRo
   }
 }
 
+// set default zoom level for tab
+function setTabDefaultZoomLevel (tabId) {
```
PalmerAL, line 55: This isn't working for me. I think the issue is that when it's called from `webviews.add`, the view object hasn't been created in the main process yet, so calling view.setZoomFactor isn't doing anything.

Another thing to consider is how this interacts with user-set zoom levels. For example, right now, if I open Github and zoom it to 150%, Electron will remember that and set the zoom to 150% on my next visit. It seems like this would reset that back to my default zoom level, which seems undesirable.

It looks like webPreferences has a zoomFactor option that can be used to do this instead: https://www.electronjs.org/docs/latest/api/structures/web-preferences. That might work better with custom zoom levels.

(the webPreferences object is created in viewManager.js)

localization/languages/en-US.json
```diff
@@ -159,6 +159,7 @@
     "settingsUserscriptsToggle": "Enable user scripts",
     "settingsShowDividerToggle": "Show divider between tabs",
     "settingsLanguageSelection": "Language: ",
+    "settingsDefaultZoomLevel": "Default zoom",
```
PalmerAL, line 162:
```suggestion
    "settingsDefaultZoomLevel": "Default zoom:",
```

Issue #2519, opened 2024-11-02
Min Browser Reset All My Custom Preferences. How Do I Restore?


DiscipleX:
I've been using Min Browser for years now, and it's been absolutely lovely.
Unfortunately, I've encountered a problem when I opened min browser like I usually do it starts, but ALL of my saved data, scripts, custom properties, tabs, and tasks are all gone.
I've been using the same set-up within min for over 2 years, and you can imagine how I felt seeing that all my custom preferences have disappeared.

Please assist me on how I can restore my Min Browser's settings back to it's previous state.

PalmerAL:
Hi, sorry to hear that that happened.

It's rather strange that all of those things would be gone, particularly as they aren't all saved in the same place. Have you recently updated Min, or is there anything else you can think of that might be related?

Anyway, I would recommend:
* Quit Min.
* Find you user data folder. This is going to be:

    ~/Library/Application Support/Min/on macOS
    %AppData%\Min\on Windows (which opens C:\Users\username\AppData\Roaming)
    /home/username/.config/Min/on Linux.

* Make sure the permissions on this folder are correct (ie your user account should be able to read/write to it).
* Make a copy of this folder as a backup.
* Userscripts should be in a `userscripts` folder inside of this. Are they there?
* Open tabs are normally saved in `sessionRestore.json`. If Min detects corruption in this during startup, it should create a `sessionRestoreBackup` file in this directory. Do you see a file starting with that name?
  * If so, try manually opening it in a text editor to get a list of URLs.
* There's also a `bookmarksBackup.html` that might be useful - do your history and bookmarks still appear inside the application?

DiscipleX:
Hey there!

It's okay my friend I definitely understand things happen.

Thank you so very much for responding and taking the time to walk me through these steps to check.

I recently did, move some files on my computer to, a different hard drive on my computer, but I did not touch, Min's folder at all. 

I'll do more research and keep this checklist in mind you sent, and will ensure that these files are in there correct location.
I definitely appreciate your response once again, I'll be in contact with you soon when I've gone through these steps!

Thank you for all the work you've put in, it's absolutely phenomenal!
Cheers/Blessings!

DiscipleX:
Hey there,

After reviewing the files in the folder you specified above: 
%AppData%\Min\on Windows (which opens C:\Users\username\AppData\Roaming)

Btw thank you for listed that exact directory, this helped alot!

I actually do have the Userscripts in this folder as you stated.

For the Tabs:
[...]

Issue #2503, opened 2024-10-02
can't see bookmarks & history once I exit min after logging in to windows


andyb9:
Min Version: v1.33.1

Operating system:
Windows 10

Expected Behavior

See bookmarks or history or both at all times.

Actual Behavior

I cannot see my see both bookmarks or history once I exit min after logging into windows and there is only a blank page. The only time I am subsequently able to see bookmarks or history is either 'hibernating' or signing out of windows and signing back into windows.

To Reproduce

I can reproduce it by looking back into to windows and starting Min and exit Min and then restarting Min and the above happen every time.

Conradus2:
I have a similar problem in the latest version of MIn.
The browsing history disappears and I have to close and then open the browser - then it's OK!!!?

Conradus2:
I have a similar problem in the latest version of MIn.
The browsing history disappears and I have to close and then open the browser - then it's OK!!!?

rduinkerken:
## Tech-information
Min version 1.35.1
Windows 11 Business, version 24H2 

## Expected behavior:
- Imported bookmarks are maintained after restarting laptop.

## Actual behavior:
- Imported bookmarks are gone after restarting laptop.

## To reproduce:

1. Import bookmarks
2. Look at Bookmarks overview. -> No bookmarks 💔
3. Close Min by pressing X
4. Startup Min
5. Look at bookmarks overview. -> Bookmarks! 🎉
6. Close Min by pressing X
7. Startup Min
9. Look at bookmarks overview. -> Bookmarks! 🎉
10. Close Min by pressing X
11. Restart laptop. 
12. Startup Min
13. Look at Bookmarks overview. -> No bookmarks 💔

**Note**
Custom key bindings are not removed.

---

Is this well-placed or should I create a separate issue for this?

PalmerAL:
I know it's been a very long time since this issue was opened, but I think I've finally identified a fix for this, which is included in the latest release (https://github.com/minbrowser/min/releases/tag/v1.35.7). If you were having this issue previously, could you try updating and let me know if it still happens?

PR #2529, opened 2024-11-13
Use atomic writes for session restore

PalmerAL:
Recently, there's been a couple reports of Min losing session restore data unexpectedly (#2519, #2527, also #2503 and https://discord.com/channels/764269005195968512/764544014259060797/1305916870473547776 which may or may not be related).

I'm not currently sure what the cause(s) of these are, as nothing has changed recently in this code, but this PR updates session restore to write the file atomically, which should mean that we can always re-launch with a valid restore file even if something goes wrong during the write. I also added some metrics collection for write failures and restoration failures to try to understand if this is a widespread issue.

It's possible that this is slower than a standard write, which mainly impacts shutdown times, as we write the file synchronously before exiting. In my very limited experimentation it doesn't seem to be too bad though.

If this resolves the issue, we could also extend this approach to the settings file, about which we've also had similar reports. That data is less critical though, so the performance vs reliability tradeoff may be different there.

### DIFF
--- a/js/sessionRestore.js
+++ b/js/sessionRestore.js
@@ -4,6 +4,8 @@ var tabEditor = require('navbar/tabEditor.js')
 var tabState = require('tabState.js')
 var settings = require('util/settings/settings.js')
 var taskOverlay = require('taskOverlay/taskOverlay.js')
+const writeFileAtomic = require('write-file-atomic')
+const statistics = require('js/statistics.js')
 
 const sessionRestore = {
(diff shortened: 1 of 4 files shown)


cristianomalben-dot:
https://support.google.com/chrome?p=chrome_uninstall_survey&crversion=142.0.7444.163&os=10.0.19045

cristianomalben-dot:
cristianomalben

Issue #2114, opened 2022-10-16
Export built in password


vudoanbinhminh:
Is there any way to export the password?  I need to import it to other browsers (or password managers)

delsallorenzo:
YES PLEASE. would have come in handy last week as I was setting up the new computer

PR #2548, opened 2025-01-11
Can import / export credentials from built-in password manager

RubenSmn:
Allows you to import / export credentials from the built-in password manager.

It will import / export using the same schema the built-in password manager uses to store the credentials which is the following:

```json
{
  "version": 1,
  "credentials": [
    {
      "domain": "google.com",
      "username": "Google User",
      "password": "123456"
    },
    {
      "domain": "github.com",
      "username": "Github User",
      "password": "123456"
    }
  ]
}
```

Currently it will only let you import if you have 0 passwords and only export if you have more than 1 password. We might want to do some merge thing but I didn't want to do that yet since that could go a couple ways, e.g. override, merge.

Also refactored some "render to dom" functionality into separate functions.

Let me know what you think.

### DIFF
--- a/index.html
+++ b/index.html
@@ -338,8 +338,8 @@
       ></div>
       <div id="password-viewer-list"></div>
       <div id="password-viewer-import-container">
-        <button id="password-viewer-import"><span data-string="importCredentials"></span><span class="i carbon:document-import"></span></button>
-        <button id="password-viewer-export"><span data-string="exportCredentials"></span><span class="i carbon:export"></span></button>
+        <button id="password-viewer-import"><span class="i carbon:document-import"></span><span data-string="importCredentials"></span></button>
+        <button id="password-viewer-export"><span class="i carbon:export"></span><span data-string="exportCredentials"></span></button>
 </div>
     </div>
 
--- a/index.html
+++ b/index.html
@@ -337,8 +337,10 @@
         hidden
       ></div>
       <div id="password-viewer-list"></div>
-      <button id="password-viewer-export"><span data-string="exportCredentials"></span><span class="i carbon:export"></span></button>
-      <button id="password-viewer-import"><span data-string="importCredentials"></span><span class="i carbon:document-import"></span></button>
+      <div id="password-viewer-import-container">
+        <button id="password-viewer-import"><span data-string="importCredentials"></span><span class="i carbon:document-import"></span></button>
+        <button id="password-viewer-export"><span data-string="exportCredentials"></span><span class="i carbon:export"></span></button>
+</div>
     </div>
 
     <!-- add scripts in Gruntfile.js -->
--- a/index.html
+++ b/index.html
@@ -337,6 +337,8 @@
         hidden
       ></div>
       <div id="password-viewer-list"></div>
+      <button id="password-viewer-export"><span data-string="exportCredentials"></span><span class="i carbon:export"></span></button>
+      <button id="password-viewer-import"><span data-string="importCredentials"></span><span class="i carbon:document-import"></span></button>
     </div>
 
     <!-- add scripts in Gruntfile.js -->
--- a/js/passwordManager/keychain.js
+++ b/js/passwordManager/keychain.js
@@ -61,7 +61,7 @@ class Keychain {
     try {
       const file = fs.readFileSync(filePaths[0], 'utf8')
       const lines = file.split('\n')
-      const credentials = lines.slice(1).map(line => {
+      const credentialsToImport = lines.slice(1).map(line => {
         const values = line.split(',')
         if (values.length !== 3 || values.some(value => value === '')) return null
         return {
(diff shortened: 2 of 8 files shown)


RubenSmn:
Might resolve #2114

PalmerAL:
Seems like a good idea!

I'm not sure whether there's a formal specification, but Safari/Chrome/Firefox all use similar CSV formats for this. What do you think about doing the same thing so that you can import/export between Min and other browsers?

For reference, the headers for each:

Safari:

```
Title,URL,Username,Password,Notes,OTPAuth
```

Chrome:
```
name,url,username,password,note
```

Firefox:
```
"url","username","password","httpRealm","formActionOrigin","guid","timeCreated","timeLastUsed","timePasswordChanged"
```

The important things seem to be "url", "username", and "password", so we would just need to do a conversion between domains and URLs.

Also, they all display some kind of security warning when you export; we should probably do something similar. This is from Firefox:
Screenshot 2025-01-11 at 3 32 04 PM

RubenSmn:
I agree that it should be compatible with the major browsers. I'll make the changes accordingly.

The security warning is a good feature to have indeed.

What about how to handle the import? Firefox gives the following success message:
image

What do you think about a warning before importing that states something like: "Existing credentials will be updated / overwritten, are you sure you want to continue?"

RubenSmn:
We can now import / export from csv files that contain the columns `url`, `username` and `password`.

After clicking import / export buttons there will first be a warning stating that the "credentials will be in readable text" (export) or "existing password will be overwritten" (import).

PalmerAL:
The import warning is good! We should probably only show the warning if you have > 0 saved passwords before the import though.

For the export warning, the Firefox one is good, but I don't want to copy it exactly :) How about "Your exported passwords will be saved in a file in plain text. When you're done using this file, you should delete it. Are you sure you want to continue?"

What do you think about using an existing CSV package, like [this](https://www.npmjs.com/package/csv)? I remember CSV having various quirks to parsing it; it might be more robust to do that rather than doing it ourselves.

index.html
```diff
@@ -337,6 +337,10 @@ <h2 class="modal-title" data-string="savedPasswordsHeading"></h2>
         hidden
       ></div>
       <div id="password-viewer-list"></div>
+      <div id="password-viewer-import-container">
+        <button id="password-viewer-import"><span data-string="importCredentials"></span><span class="i carbon:document-import"></span></button>
```
PalmerAL, line 341: Can you move these icons so that they're to the left of the text?

js/passwordManager/keychain.js
```diff
@@ -48,6 +48,53 @@ class Keychain {
     ipcRenderer.invoke('credentialStoreDeletePassword', { domain, username })
   }
 
+  async importCredentials () {
```
PalmerAL, line 51: I think it would be better if this took the CSV file contents as an argument, and the open dialog part was moved into passwordViewer. It's possible the CSV parsing would be better moved into there as well, but I'm not sure about that, so I'll leave that up to you.

RubenSmn: I was reading this comment and was like, "huh but this is already in the viewer right?" apparently not, thanks for catching :)

PalmerAL:
This seems pretty close now! At the moment I only see two more issues:

1. We need to convert between the `url` field that other browsers use and the `domain` field that we use - right now, if you import another browser's CSV, autofill doesn't work because of this mismatch. I think on import, we can do this:
```
const credentialsToImport = csvData.data.map((credential) => {
  try {
    return {
      domain: new URL(credential.url).hostname,
      username: credential.username,
      password: credential.password
    }
  } catch (e) {
    return null
  }
}).filter(credential => credential !== null)
```

For export, I don't know of a nice solution, but we can probably do `'https://' + hostname + '/'`?

2. Can you update the Dutch translation of the export message to match the new English version?

RubenSmn:
The other browsers mentioned export the domains with `www.` and Min removes the `www.` when asked to store the credentials for a domain. https://github.com/minbrowser/min/blob/master/js/passwordManager/passwordCapture.js#L50-L53

When using `new URL(...).hostname` it will contain the `www.`. We could do something like `new URL(...).hostname.replace(/www\./,'')` to make sure the urls/domains match.

For the export I think `'https://' + hostname` would work fine, I don't think there is a need for the trailing slash.

PalmerAL:
Oh, I see, removing the `www.` does make sense in that case. We need to make sure to only remove it from the start of the hostname though - otherwise we'd transform examplewww.com -> examplecom, which would be wrong.

PalmerAL:
Sorry I missed this earlier, but we need to escape characters correctly on export. For example, if your password is `abc,def`, we'll currently do this, which won't be parsed correctly later:
```
https://example.com,user,abc,def
```

Safari escapes it like this:
```
example.com (user),https://example.com/,user,"abc,def",,
```

I imagine it we use papaparse for the export it will handle this automatically?

Also, I noticed that we mostly use the word "passwords" in the existing UI, but the strings in this PR use "credentials". What do you think about using "passwords" instead?

RubenSmn:
I've made the changes.

One little thing, whenever we import for example `http://localhost:3000` into min, if we export it afterwards it gets a `https://` protocol, this means that the credentials won't work for that url anymore.

Do you think we should account for this? Maybe specifying a protocol field when saving to the build-in manager could be a solution?

PalmerAL:
Good point. I imagine the right thing to do is to save a full URL like other browsers do, although we don't have a good way to migrate passwords that are already saved to this format, other than just assuming `https:// + domain`. Perhaps this would be better in a follow-up PR, if it's something you want to do?

Other than the localhost example, I imagine there shouldn't be too many sites sending passwords over insecure HTTP, so it hopefully shouldn't cause issues very often.

RubenSmn:
Yes I think it would be a better fit for a separate PR to keep this one just related to the import / export functionality.

PalmerAL:
I think everything looks good here; we'll address the credentials issue later. Thank you!

PR #2555, opened 2025-01-25
feat: Add hide translate menu setting

5brian:
Changes:
- added option to hide `Translate Page(Beta)` from context menu
- located in additional features

### DIFF
--- a/js/webviewMenu.js
+++ b/js/webviewMenu.js
@@ -347,7 +347,9 @@ const webviewMenu = {
       }
     })
 
-    menuSections.push([translateMenu])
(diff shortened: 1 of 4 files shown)


PalmerAL:
Can you explain the reasoning for adding this option?

5brian:
I just thought it would clean up the context menu for those who do not use the feature, which would add to the minimal look of the browser

PalmerAL:
Sorry I didn't get a chance to reply to this earlier, but since the menu items are fairly unobtrusive and hidden most of the time, I don't think this option is necessary. Thanks for your effort though!

PR #2562, opened 2025-02-08
Chore/credential migration

RubenSmn:
New version of storing the credentials. Allows for storing the whole URL including, protocols and all subdomains.

The Migrator will run on startup check the version of the built-in store and updates the credentials to the new format if necessary.

For the migration it will
- check if a url matches any in the history, if yes, use that url, else
- check if the url already has a protocol if not, add https:// protocol
- finally store the new credentials and update the version

The same is done for the credentials that should never be saved


PalmerAL:
Thanks, I think this approach looks good generally. I haven't fully tested this yet, but a couple of initial comments:

* In the keychain import, the saved field is this: `domain: new URL(domainWithProtocol).origin,`, but in the migrator, the equivalent field is `url: historyEntry.url`. Is your intent to save an origin or a complete page URL? (I think origin should be fine)
* Doing this isn't secure: `matchHost.replace('www.', '') === domain || matchHost === domain` - imagine comparing "example.com" and "example.www.com". The www should be removed only if it appears at the start, immediately after a protocol.
* We'll need a change for Bitwarden as well (like I did [here](https://github.com/minbrowser/min/commit/03dfe91d3a59b84b86811988cdece84de4e0c6c5#diff-0c983941ba1ffaebf8d04ded3f8a869a58c378d701f3d8f5a6029839dd138f5e)).

RubenSmn:
Thanks for the feedback, I missed those first 2 points you mentioned :sweat_smile: 
Added the change for Bitwarden.

Let me know if I need to make more improvements!

main/keychainService.js
```diff
@@ -88,3 +88,13 @@ ipc.handle('credentialStoreDeletePassword', async function (event, account) {
 ipc.handle('credentialStoreGetCredentials', async function () {
   return readSavedPasswordFile().credentials
 })
+
+ipc.handle('credentialStoreGetVersion', async function () {
+  return readSavedPasswordFile().version
+})
+
+ipc.handl('credentialStoreSetVersion', async function (event, version) {
```
PalmerAL, line 96:
```suggestion
ipc.handle('credentialStoreSetVersion', async function (event, version) {
```

Typo :)

js/passwordManager/passwordCapture.js
```diff
@@ -48,11 +48,11 @@ const passwordCapture = {
   },
   handleRecieveCredentials: function (tab, args, frameId) {
     var domain = args[0][0]
```
PalmerAL, line 50: Shouldn't we be saving an origin or a URL into `currentDomain` (which later gets passed to `saveCredential`)? With this way, I think we're not recording the protocol.

In my branch I addressed this by passing the entire URL into the manager: https://github.com/minbrowser/min/compare/master...password-url-matching#diff-dfe869359e479c9fd76fbe8e10c5f46545b3ec71d7a029aafd096869489da6ed

js/passwordManager/keychain.js
PalmerAL, line 1: We need to change `saveCredential`, right? As I mentioned in the comment on `passwordCapture`, it's only recording a domain still.

I think we should pass the whole URL into the manager, and then let the manager parse and save the origin: https://github.com/minbrowser/min/compare/master...password-url-matching#diff-cead69e9db343b1d609e235d95b2610037795876bcd42860ebb620b78d9fa7ecR47

js/passwordManager/passwordMigrator.js
```diff
@@ -0,0 +1,94 @@
...
+    console.log('[PasswordMigrator]: Found', currentCredentials.length, 'credentials in the current password manager', currentCredentials)
+
+    function createNewCredential(credential) {
+      // 1) check if the saved url has been visited, if so use that url
+      const historyEntry = historyData.find(entry => new URL(entry.url).host.replace(/^(https?:\/\/)?(www\.)?/, '') === credential.domain.replace(/^(https?:\/\/)?(www\.)?/, ''))
+      if (historyEntry) {
+        return {
+          username: credential.username,
+          password: credential.password,
+          url: new URL(historyEntry.url).origin
```
PalmerAL, line 58: The name of this field is `domain` in the import in keychain.js, so it needs to use the same name here.

js/passwordManager/passwordMigrator.js
```diff
@@ -0,0 +1,94 @@
...
+
+      // 2) check if domain has protocol, if not, add 'https://'
+      if (!newUrl.startsWith('http://') && !newUrl.startsWith('https://')) {
+        newUrl = `https://${newUrl}`
+      }
+
+      return {
+        username: credential.username,
+        password: credential.password,
+        url: newUrl
```
PalmerAL, line 70: Same thing here: The name of this field is `domain` in the import in keychain.js, so it needs to use the same name here.

PalmerAL (commented):
I left some more comments. Let me know if I read something incorrectly!

Due to the typo in `ipc.handle`, this branch actually crashes on launch when I try to run it. In addition to the comments, could you please make sure to test this on some passwords/sites of your own and confirm that everything works?

js/passwordManager/passwordCapture.js
RubenSmn: In the [passwordFill.js](https://github.com/minbrowser/min/pull/2562/files#diff-690430546b74ba1ee9ff083e7ac36602f2d263fd486073ba41545ea7256bf5dbR347) we pass the `origin` part of the url instead of the `hostname` which will give us the whole base url `https://sub.example.com` and that should be sufficient for the `saveCredential` funtionality.

Maybe the `domain` variable name is a little confusing, perhaps it needs to be something like `urlBase` or `urlOrigin`

js/passwordManager/passwordMigrator.js
RubenSmn: Good catch, we could migrate it to be `url` so we align with the other browsers' default?

js/passwordManager/passwordCapture.js
PalmerAL: Got it, sorry for the delay. It sounds like something with "origin" would be a better name then.

js/passwordManager/passwordMigrator.js
PalmerAL: Sure, if you want to update the code to actually save the entire URL, that's fine too. If you want to just save the origin, then we should call the field `origin`.

PR #2577, opened 2025-03-01
feat: Prevents other programs from capturing

eric-gitta-moore:
link: https://github.com/minbrowser/min/discussions/2576


PalmerAL:
Can you explain what the use case for this is?

In desktop OSs, there's effectively no sandboxing between applications, so this isn't a security improvement - a malicious application could still access the contents of any of Min's memory while it's running.

It does prevent you from accidentally including a Min window in a recording of something else - for example, screen sharing on a video call. But it feels like that's something you'd want to enable only for specific tabs, because there are still a lot of times where you'll want to take a screenshot of something in the browser.

eric-gitta-moore:
@PalmerAL Yeah, what you said is okay. It is only able to prevent other applications from capturing, and it cannot do anything about memory scanning. However, in some cases, it may only be necessary for Google Meet to fail to record, and this setting can take effect immediately

PalmerAL:
What if we make the setting apply only to private tabs?

eric-gitta-moore:
> What if we make the setting apply only to private tabs?

In macOS, if regular tabs and private tabs are under the same window, it is not possible to hide a specific part individually by setting [NSWindow.SharingType.none](https://developer.apple.com/documentation/appkit/nswindow/sharingtype-swift.enum/none).

However, I think we could try separating the private tabs into a standalone window, which would allow us to set the `NSWindow.SharingType` for this individual window.

PalmerAL:
Are we able to repeatedly change the setting after the window is created? We only show one tab on screen per window at a time, and I don't think we have any plans to implement split view. So I was thinking we could enable the protection when a private tab is selected, and disable it when the current tab is switched to a non-private one.

eric-gitta-moore:
> Are we able to repeatedly change the setting after the window is created? We only show one tab on screen per window at a time, and I don't think we have any plans to implement split view. So I was thinking we could enable the protection when a private tab is selected, and disable it when the current tab is switched to a non-private one. Although I guess we would have to be careful about the ordering of those things, to prevent showing a flicker of the private tab when it's selected.

Yes, it is a feasible approach.

PalmerAL:
Do you think it's a good idea to do it that way?

Issue #1974, opened 2022-04-14
Tabs and bookmarks should have favicon


ductrantech:
I think you guys should add icon for each tabs and bookmarks for the ui, it helps searching tab more faster.

flightmansam:
Interesting idea... There are favicons already in the task overlay <img src="https://user-images.githubusercontent.com/34013162/164213545-06b5b464-fd0e-4b91-b670-3d2234045041.png" width="30">. 

I think including favicons EVERYWHERE goes against the design philosophy of min being a minimalist browser.

il1yaz:
@flightmansam,  In my daily use i have a lot of tabs in browser. And for me it's hard to find specific tab without some hints. We have title, but it's not helping much. Maybe you can add some setting that allows users to chose with or without favicons? I agree with you that minimalistic browser should use as minimal things as could, but it's need to be usable at the other side.

flightmansam:
> @flightmansam,  In my daily use i have a lot of tabs in browser. And for me it's hard to find specific tab without some hints. We have title, but it's not helping much. Maybe you can add some setting that allows users to chose with or without favicons? I agree with you that minimalistic browser should use as minimal things as could, but it's need to be usable at the other side.  

Okay. You know what? I'll have a crack.

sosise:
bump

Issue #1855, opened 2022-01-07
Favicon displays


JoaoNM:
Would really like to have favicon displays for better UX so users can quickly find tabs they're looking for

subodhpareek18:
Yes that's a good idea, but perhaps make it optional in the setting?

fdemir:
Any updates on this? I would like to contribute if you give me some detail on it.

sosise:
I also need this "feature"

Issue #1078, opened 2020-06-19
showing website icons in tabs


raj123:
Min shows website title in tabs. It does not show icons of website.

![image](https://user-images.githubusercontent.com/310979/85095694-c81fc400-b20f-11ea-87a2-5b4bd0d4f07b.png)

nanaian:
Possibly worth considering multiple options for favicons:
- No favicons (current behaviour)
- Standard favicons
- Mask favicons where supported (see Apple's [mask-icon meta tag](https://stackoverflow.com/questions/40844731/custom-safari-bookmark-icon-and-color-on-touch-bar#40960861))

PalmerAL:
I'm not convinced it makes sense to support this, because icons take up a lot of space, can be distracting, and probably aren't compatible with the current system of making the tab bar color match the site (since making the icon visible against arbitrary backgrounds is difficult).

The mask icon thing is interesting, although I don't have much of an idea of how widely supported it is.

Nesaijn:
I like the idea of having icons on the tabs. For me it makes it easier and faster to recognize what tab is for which site.
Relating the issue with space, instead of naming the domain on the tab, you could show the icon which would probably even lead to less space taken.

Issue #757, opened 2019-05-22
Favicons


Tomadelostacos:
Hello, is there something planned for the favicons in the tabs ?

PalmerAL:
Duplicate of #304.

In general, I think favicons would be kind of distracting, and it's also going to be difficult to find a way to do it that looks good, since tabs can have any possible color as their background (unless we wanted to remove that), but I can see how it would be nice to have a way to identify tabs more quickly. Maybe something where a portion of the tab uses the theme color for that page?

Tomadelostacos:
Yep, i personally identify my tabs with favicons in Opera, so that i instantly know in which website i'm gonna go when clicking, it's annoying (for me) if i have to read all labels beginnings when choosing a tab. 
Maybe having a little div at the left of every tab, a square with a dark grey background, hideable, could allow users to have favicons and to identify even more tabs ?

PalmerAL:
That's true, I guess if there was a background behind the favicon, it could be arranged in such a way that it never conflicts with the tab color.

This does seem like it would be useful in some situations, I guess I would be OK with adding an option to it.

PalmerAL:
Actually, we could probably detect what the colors in the favicon are, compare them to the background color, and only add the gray background if they're too similar to each other...

Tomadelostacos:
It could be a great idea, it's not "too heavy" if the user has a lot of opened tabs ?

PalmerAL:
It would probably make sense to hide the favicons if there are too many tabs open, although I'm not exactly sure what number that would be.

Tomadelostacos:
I personally don't fully agree on that. As a user, when i have a lot of tabs opened, i can find my tab faster with favicons + location in my tabs than with the first letters. (It's my personal opinion ! )

PalmerAL:
That's a good point, I guess an alternative would be to hide the title and just use the favicon when the tab reaches the minimum width.

PalmerAL:
Closing as duplicate of #304.

PR #2598, opened 2025-04-18
feat: Favicons in tabs

PennDraken:
# Description
Added commonly requested option to toggle favicons in tabs.
Open issues covered (note: many duplicates): 
[#304](https://github.com/minbrowser/min/issues/304)
[#757](https://github.com/minbrowser/min/issues/757)
[#304](https://github.com/minbrowser/min/issues/304)
[#1078](https://github.com/minbrowser/min/issues/1078)
[#1974](https://github.com/minbrowser/min/issues/1974)
[#1855](https://github.com/minbrowser/min/issues/1855)

# Images
No favicons (default):
image
Favicons:
image
Settings page:
image
# Known issues
- Icons might have low contrast depending on tab color
- Favicon only updates when page is fully loaded. This means it sometimes takes a little while for the old favicon to update.

### DIFF
--- a/css/tabBar.css
+++ b/css/tabBar.css
@@ -183,6 +183,19 @@ body:not(.touch)
   margin-left: -0.5rem;
 }
 
+.favicon {
+  display: none; /* Default state (not enbabled) */
+}
+
+#navbar.show-favicons .favicon {
+  display: revert;
+  width: 16px;
+  height: 16px;
+  margin: auto 0;
+  padding: 0;
+  font-size: 1.1em !important;
+}
+
 .tab-item .tab-icon {
   position: relative;
   font-size: 0.875em;
--- a/js/navbar/tabBar.js
+++ b/js/navbar/tabBar.js
@@ -52,6 +52,16 @@ const tabBar = {
     tabEl.appendChild(readerView.getButton(data.id))
     tabEl.appendChild(tabAudio.getButton(data.id))
     tabEl.appendChild(progressBar.create())
+    if (data.favicon) {
+      var favicon = document.createElement('img')
+      favicon.className = 'favicon'
+      if (data.favicon) {
+        favicon.src = data.favicon.url
+      } else {
+        favicon.src = data.favicon.url
+      }
+      tabEl.appendChild(favicon)
+    }
 
     // icons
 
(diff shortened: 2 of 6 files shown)


PalmerAL:
Thanks for opening a PR; there have been enough requests for this that it probably does make sense to do something.

The contrast issue has been one of the big reasons for not doing this until now; do you have any thoughts on how to address that? Some possibilities are:

* Filter the favicon to make it black-and-white
* Use mask-icons instead of favicons, where available: https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/pinnedTabs/pinnedTabs.html
* Only make the favicon option available when the option to change the tab bar color is turned off.

If the favicons are too slow to load, you could look into detecting them from the preload script, rather than relying on the page-favicon-updated event; that might be faster, but I'm not sure.

PennDraken:
Thanks for your response! <3

> Filter the favicon to make it black-and-white

image
Grayscale would probably not improve contrast, but perhaps forcing it to be either black or white might. This would not work very well for the Google icon in the screenshot however.
Might be solveable by mapping different values to either black or white.
image
_Brightness mapping_

Perhaps by calculating average color of icon, and checking contrast with tab color, we could then invert icons that have bad contrast.
image
I think this approach might work well with keeping to the minimal theme of the browser

> Use mask-icons instead of favicons, where available:

Would probably work, although many sites do not support them (such as apple.com and google.com). Are you just thinking we re-color it either black or white to match the tab text color?

> Only make the favicon option available when the option to change the tab bar color is turned off.

Could be a quick fix, although I think a sizeable amount of users might prefer tab theming with ugly low-contrast favicons.

> Using preload script, rather than page-favicon-updated event

I'll look into this

Another solution might be to use dark-mode and light-mode favicons when available.

PalmerAL:
> Brightness mapping

I like the screenshot a lot!

>the issue will be that this brightness mapping will be different for different favicons

This does seem tricky. For a "flat" style favicon with only a few colors, you could probably just sort all the pixels by brightness and pick the median as the threshold. This would totally stop working once you have a favicon with a gradient though. Maybe an alternative would be: put all the pixels in buckets by brightness: [0.00, 0.01, ... 0.99, 1.00]. Look in the range [0.3, 0.7] for a discontinuity, which will be represented as a series of buckets greater than some length with few or no pixels, and make that the white/black threshold.

> Are you just thinking we re-color it either black or white to match the tab text color?

Right; since they only have a single color they're easy to re-color. But I haven't researched how many sites support them; if usage is low that won't be very useful.

One downside to doing recoloring like this is that color is a big part of what makes the favicons quickly recognizable. If you recolor them to black and white, do they still feel useful for identifying tabs?

Maybe the favicon setting should be always available, but if tab color is turned on we convert the favicons to black/white, and if it's off we show them in original color. I'm not sure.

> I also have a question regarding the placement of favicons in combination with the tab-icons (reading, playing-audio). What do you think feels best, placing favicons left-most when other tab-icons are visible or right-most?

I'm not sure about this either :) I'll probably need to try it both ways and see. 

Also, we probably want to hide the favicons entirely when the tabs become very small.  There's an existing `compact-tabs` class that can probably be reused for this.

PennDraken:
Sorry I've been busy but am looking into this feature again.

Something I missed the last time: The background color of the top bar is calculated based on the favicon's prominent color (`getColorFromImage`). This of course often creates conflicts between the selected tabs favicon and the bg color.
I think another approach (to instead of changing a favicons color), we could instead **try to shift the background colors brightness** slightly to reduce visual conflicts between all favicons. I have not experimented with this approach, but it would allow icons to be unaltered.

# Monochrome favicon approach
I'm trying out some favicon rendering tests now, using a simple algorithm to turn favicons into a single color. And matching this color with the fg. It's not the smoothest right now, but I think it looks pretty promising!
favicon-demo

PR #2630, opened 2025-08-07
Favicons in tabs (In the latest Version)

zoieyyy:
-Added favicons in tabs (in the latest version)
-Added setting toggle for favicons and US localisation


PalmerAL:
This is from #2598, right? It'd be great if you want to continue working on this, since I don't think the original author is any more. Do you plan to address the issues that were brought up in that PR?

PR #2631, opened 2025-08-10
Improve Serbian language

rammba:
Hello @PalmerAL , I'm glad to have a chance to contribute to your project.
I'm currently working on translating the [official React documentation](https://github.com/reactjs/sr.react.dev) to Serbian, but I also contributed with translations to few other open-source projects.

My change here was mainly in using cyrillic instead of latin version of Serbian language introduced in #1914. Cyrillic is formal and it's used in all legal papers, so I thought it should be used here as well.
If I understood [Chromium](https://source.chromium.org/chromium/chromium/src/+/main:ui/base/l10n/l10n_util.cc) well, they don't differentiate Serbian cyrillic and latin versions, right? If I'm missing something, please tell me how to name latin file to have both versions available.

Hey @pvlvukovic, since I'm changing your previous work, please feel free to give your suggestions.

### DIFF
--- a/localization/languages/sr.json
+++ b/localization/languages/sr.json
@@ -1,315 +1,319 @@
 {
-  "name": "Serbian",
+  "name": "Српски",
   "identifier": "sr",
   "translations": {
     /* Context menu items
         these are the items displayed in the menu when you right-click on a page
         */
-    "addToDictionary": "Dodaj u rečnik",
-    "pictureInPicture": "Slika u slici",
-    "openInNewTab": "Otvori u novom tabu",
-    "openInNewPrivateTab": "Otvori u novom privatnom tabu",
-    "saveLinkAs": "Sačuvaj link kao...",
-    "viewImage": "Pogledaj sliku",
-    "openImageInNewTab": "Otvori sliku u novom tabu",
-    "openImageInNewPrivateTab": "Otvori sliku u novom privatnom tabu",
-    "saveImageAs": "Sačuvaj sliku kao",
-    "searchWith": "Pretraži na %s", //%s will be replaced with the name of the current search engine
-    "copyLink": "Kopiraj link",
-    "copyEmailAddress": "Kopiraj email adresu",
-    "selectAll": "Selektuj sve",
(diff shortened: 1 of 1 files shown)


PalmerAL:
I agree, it doesn't look like Chromium differentiates. I'm not familiar with the language at all, so sorry for the basic questions, but which form is more commonly used in general? What form do other browsers or operating systems use by default?

rammba:
Hey @PalmerAL, thanks for the quick response.

First of all, all Serbians understand both cylillic and latin versions. It's basically the same language just with a different set of letters.
As I mentioned, cyrillic is a more formal one which needs to be used in all official documents and papers. For software, I think it depends a lot. Sometimes it has both versions, sometimes only one can be supported and I guess it depends on the translator(s) preferences.
I think that people oftenly use latin for computer/digital communication because it's easier to use it without changing the keyboard.

I proposed similar change recently in the [darkreader](https://github.com/darkreader/darkreader/pull/14412) & [FluentValidation](https://github.com/FluentValidation/FluentValidation/pull/2283) applications. If you think changing the language version can cause problem to your app, I can use latin script then.

PalmerAL:
I don't think it would cause any problems to use either script, I just wonder what would be more consistent with other applications. For example - when you set Serbian as the language in Windows, which is used by default? Does it give you a choice between both?

What about Chrome? This would be particularly useful to know since they seem to only support one script - do they use Latin or Cyrillic?

rammba:
Hello @PalmerAL,
Chrome uses cyrillic version. Windows & Microsoft Edge supports both versions and also variants of the language used in Montenegro and Bosnia.

Chrome:
image
image

Windows:
image

Microsot Edge:
image

rammba:
Hello @PalmerAL, is it ok to use cyrillic version, or to switch back to latin?

PalmerAL:
OK, sure, I think it's fine to use Cyrillic. Sorry for the delay. I will try to review this again soon.

PalmerAL:
I think everything looks good, thanks.

Issue #2639, opened 2025-09-16
(Re-)Add setting to disable swipe gestures for navigation


butburg:
**Context:**
Swipe gestures (back/forward navigation) can be unintentionally triggered on trackpads and touch devices (testet on mac). It seems like it was a feature in past, but i couldn't find the comment again.

Several related issues discuss this behavior (in comments):
- [Add better page forward/back gestures](https://github.com/minbrowser/min/issues/743#top)
    - Collaborator PalmerAL [on Feb 29, 2020](https://github.com/minbrowser/min/issues/743#issuecomment-592990267):
    > Hmm, OK. I'm not seeing any way to detect the trackpad sensitivity to compensate for this; if it's a common issue, we could probably add a setting to manually adjust the amount of movement required, but it sounds like it would be fine to leave it as-is for now.

- [Trackpad gestures are too sensitive](https://github.com/minbrowser/min/issues/1138#top)
    - strindhaug [on May 22, 2024](https://github.com/minbrowser/min/issues/1138#issuecomment-2124191977):
    > Option to disable this misfeature is really needed! [...]

- [Indicator showing when page is about to go to previous page](https://github.com/minbrowser/min/issues/2241#top)
    - strindhaug [on May 22, 2024](https://github.com/minbrowser/min/issues/2241#issuecomment-2124249061):
    > Still ridiculously sensitive on Mac touchpad [...] 

- IanDLacy [on Feb 3, 2022](https://github.com/minbrowser/min/issues/27#issuecomment-1029471885):
    > Is there a way to disable this behavior entirely?

**Proposal:**
Introduce a user setting to disable swipe gestures. This gives users who don’t want gesture navigation a simple opt-out.

**Implementation sketch:**

* **webviewGestures.js** [line 91](https://github.com/minbrowser/min/blob/a8aa8ae271213597551bf83540b418851ace9033/js/webviewGestures.js#L91)
  Current guard:

  ```js
  if (isInFrame) { return }
  ```

  Suggested change:

  ```js
  if (isInFrame || isSwipeSettingDisabled) { return }
  ```

* **index.html** (Settings UI, maybe placed below the autoplay toggle):

  ```html
  <div class="setting-section" id="section-disable-swipe">
    <input type="checkbox" id="checkbox-disable-swipe" />
    <label
      for="checkbox-disable-swipe"
      data-string="settingsDisableSwipeGestures"
    >
      Disable swipe gestures for navigating backward and forward
    </label>
  </div>
  ```

[...]

PR #2642, opened 2025-10-07
Added setting to disable swipe gestures for navigation

sandeep-singhDEV:
What does this PR do ?
It adds setting to disable swipe gestures for navigation.

It fixes issue #2639 

Type of issue
Re-add of a feature


PalmerAL:
I suppose we could to this, but I think it would be better to work on improving the gesture detection system so that it triggers accidentally less often. Is that something you'd be interested in working on?

PR #2646, opened 2025-10-19
Add hidden numpad shortcuts for zoom actions

Chickaboo:
Introduces hidden menu items to enable CmdOrCtrl+numadd and CmdOrCtrl+numsub shortcuts for zooming in and out using the numpad. This improves accessibility for users who prefer numpad shortcuts.


Chickaboo:
I forgot GitHub keeps adding commits on a branch to a pull request after its made. I fixed the mess.

PalmerAL:
On my (US QWERTY) keyboard with a numpad, zooming with the +/- keys already works - it looks like these are mapped to the standard +/- keys that already have shortcuts registered. If this isn't already working for you, which keyboard layout do you have?

Otherwise; this looks fine; if it improves things on certain keyboard layouts it should be good to merge.

Chickaboo:
I am on the standard Windows keyboard layout, which I believe is the same as yours US QWERTY. Zooming with the main +/- keys works but the matching keys on the numberpad do not support the same functionality in my situation. I thought this would be a simple contribution which would bring more compatibility, allowing all users to utilize whatever they prefer: the main keyboard or the numberpad, as well as insuring harmony with chrome keybinds. (I am not sure why the numberpad did not work for me and Ariadesu, the person that originally made the corisponding issue to this pull request, but worked for you. This did fix it for me though.) Thanks for reviewing this.

PalmerAL:
OK, good to know; thanks!

PR #2648, opened 2025-10-20
Add customizable homepage

Chickaboo:
Introduces a new customizable new tab page with clock, greeting, and search bar. Adds settings for enabling/disabling the homepage, choosing between default or custom URL, toggling clock and greeting, and selecting time format. Implements new internal protocols for 'min://newtab' and 'min://blank', updates localization, and refines related CSS and JS for the new tab and settings pages.
Screenshot 2025-10-19 213511
Screenshot 2025-10-19 213254
Screenshot 2025-10-19 213217
Screenshot 2025-10-19 213230

### DIFF
--- a/css/newTabPage.css
+++ b/css/newTabPage.css
@@ -18,11 +18,11 @@ body:not(.is-ntp) #ntp-content {
     margin: auto;
     object-fit: cover;
     transition: 0.2s filter, 0.2s transform;
+    transform: scale(1.03);
 }
 
 .ntp-blur #ntp-background {
     filter: blur(12px);
-    transform: scale(1.1);
 }
 
 #ntp-background-controls {


PalmerAL:
Thanks for the PR!

I think it makes sense to have a setting to set a custom new tab URL. I'm less convinced that we should have a built-in page. My impression is that there are a lot of potential designs for a new tab page, and people tend to have strong opinions about which design is the best for them; designing something that makes most people happy would require a ton of different customization options and add a lot of complexity.

How would you feel about keeping just the setting to set a custom URL, and putting any specific page designs in an external repository?

Also, it would be good to figure out a better interaction between the NTP and the searchbar. When I open a new tab, I don't actually see the NTP, because the searchbar is covering it:

Screenshot 2025-11-09 at 2 04 40 PM

I think this should probably work like this (what do you think?):
1. When you open a new tab, the searchbar is active and focused, but if a custom NTP is configured, no suggestions are shown underneath.
2. The NTP is immediately visible.
3. If you start typing in the searchbar, we start showing the searchbar suggestions and blur the NTP.

PalmerAL:
Also, I should point out, we already have an option to set a custom image background on the new tab page:

Screenshot 2025-11-09 at 2 09 55 PM

We should ensure any changes are compatible with that.

Chickaboo:
That sounds good to me. I'll remove the new tab page and make it so that in the settings, you can choose either the NTP or a custom URL. A better implemetion of the NTP and the search bar interaction is something I will look into. I think the outline you gave will work well. I'll try to work on this sometime soon, but it might be a week or two.

Chickaboo:
Everything should be done. I moved the Custom URL into the Additional Settings. There is now a dropdown menu which has three options one for Blank, Custom URL, and NTP. I made sure that both the NTP widget and the preferences functionality are in sync. The changes to the new tab design have been made as well. I think the outline you gave worked great. Let me know if there are any changes or tweaks needed—its looking pretty good to me though. [9b59089](https://github.com/minbrowser/min/pull/2648/commits/9b59089262fd20acd5bce825dc581ac3dff5472b)

css/newTabPage.css
```diff
@@ -17,6 +17,12 @@ body:not(.is-ntp) #ntp-content {
     height: 100%;
     margin: auto;
     object-fit: cover;
+    transition: 0.2s filter, 0.2s transform;
+}
+
+.ntp-blur #ntp-background {
+    filter: blur(12px);
+    transform: scale(1.1);
```
PalmerAL: This makes sense to eliminate the white edges around the blur, but I think it's pretty distracting. What if we:
* Reduce the magnitude to `scale(1.03)` - it looks like this is still enough to eliminate the white edges
* Apply the scale to the image all the time, so that there's no motion when the blur is applied?

js/navbar/tabEditor.js
```diff
@@ -45,8 +47,18 @@ const tabEditor = {
...
         searchbar.showResults(editingValue, null)
+        searchbar.el.classList.remove('searchbar-hidden')
       } else {
         searchbar.showResults('', null)
+        const newTabUrlMode = settings.get('newTabUrl')
+        const isNTP = (newTabUrlMode === 'backgroundImage' && currentURL === '') ||
+                      (newTabUrlMode === 'customUrl' && currentURL === settings.get('newTabCustomUrl'))
+        
+        if (isNTP) {
+          searchbar.el.classList.add('searchbar-hidden')
```
PalmerAL: I think we could just skip calling `searchbar.showResults` in this case, and then we no longer need the hidden class.

js/newTabPage.js
```diff
@@ -9,6 +11,15 @@ const newTabPage = {
   imagePath: path.join(window.globalArgs['user-data-path'], 'newTabBackground'),
   blobInstance: null,
   reloadBackground: function () {
+    const newTabUrl = settings.get('newTabUrl')
+    if (newTabUrl === 'blank') {
```
PalmerAL: If the setting has never been modified, settings.get will return undefined, which this also needs to check for.

main/menu.js
```diff
@@ -257,24 +257,6 @@ function buildAppMenu (options = {}) {
             sendIPCToWindow(window, 'zoomOut')
           }
         },
-        // Hidden item to enable shortcut on numpad
```
PalmerAL: This seems like an accidental change, can you revert it?

PalmerAL (commented):
This looks pretty good! I left some comments on the code, but I think the functionality is working well.

PR #2655, opened 2025-11-01
Completed the localisation for German

PingPongBing:
I updated and filled the null statements (where translation is missing) with the respective German translation and thus updated the localization. The localization is complete for the German language.
I think its worth a review.


PalmerAL:
Looks good, thanks!

PR #2678, opened 2026-02-11
Add smooth width animations for tabs

m11tya:
- Tabs expand to 1.25x width on hover
- Active tabs expand to 1.6x width
- Smooth transitions using flexbox
- Other tabs compress when one expands

https://github.com/user-attachments/assets/8b666e1e-37a4-4190-9dfe-6010e0c25bcd

### DIFF
--- a/css/tabBar.css
+++ b/css/tabBar.css
@@ -100,7 +100,7 @@ body.linux #menu-button {
 .tab-item {
   flex: 1;
   min-width: 125px;
-  transition: 0.2s transform;
+  transition: flex 0.3s cubic-bezier(0.4, 0, 0.2, 1), min-width 0.3s cubic-bezier(0.4, 0, 0.2, 1), background-color 0.2s;
   line-height: 36px;
   height: 36px;
   overflow: hidden;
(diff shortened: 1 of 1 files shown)


css/tabBar.css
```diff
@@ -113,7 +113,8 @@ body.linux #menu-button {
   padding-right: 0.5rem;
 }
 .tab-item.active {
-  min-width: 168px;
+  flex: 1.6;
+  min-width: 268px;
```
PalmerAL: Obviously there's no objectively correct value, but this feels a bit too large to me. How about 220px?

css/tabBar.css
```diff
@@ -161,12 +162,18 @@ body.linux #menu-button {
...
 }
 .dark-theme .tab-item:not(.active):hover {
   background-color: rgba(255, 255, 255, 0.1);
+  flex: 1.25;
+  min-width: 156px;
 }
 .dark-theme.theme-background-low-contrast .tab-item:not(.active):hover {
   background-color: rgba(255, 255, 255, 0.15);
+  flex: 1.25;
+  min-width: 156px;
```
PalmerAL: I think these are already inherited from the rule above.

m1tyarch:
I will implement your comments as soon as possible.

@PalmerAL thank you

m1tyarch:
Done https://github.com/minbrowser/min/pull/2679

@PalmerAL

PalmerAL:
Thanks; I've added comments there. FYI, in the future, you can add commits to this PR rather than opening a new one.

PR #2679, opened 2026-02-15
feat: improve UI animations and tab behavior

m11tya:
- Fix scrollbar colors in dark mode (darker thumb)
- Improve tab sizing: 1.1x for 2-4 tabs, 1.6x for 5+ tabs
- Add trackpad scroll transform animation support

PR #2684, opened 2026-02-19
Use compact-tabs for active tab growth

m11tya:
Fixed everything that @PalmerAL asked for.


PalmerAL:
Can you modify your previous PR (#2679)?

This PR now has _only_ the most recent change, but it needs to be combined with the older ones from that PR.

PR #2687, opened 2026-03-06
Fix unable to open local pdfs

### LINKED ISSUE #2375: PDF files don't open correctly anymore
- **Min Version:** 1.31.1

## Expected Behavior

Opening a local pdf file with Min should let you view the file.

## Actual Behavior

Instead, the browser instantly prompts to save the file, upon canceling the download the file never loads.

## To Reproduce

Open a locally saved pdf file using Min.

pechenb:
FIxes #2375.

Tested only on MacOS 15.7.3

### Changes

- `main/download.js`
  Added IPC handler `readLocalPDF` that reads local PDF bytes in the main process.

- `main/localPDFAccess.js`
  Added centralized validation for local PDF access:
  - request must come from internal PDF viewer (`min://app/pages/pdfViewer/index.html...`)
  - requested URL must be `file://`
  - requested file must exactly match the PDF URL opened in that viewer

- `js/preload/default.js`
  Added bridge for `readLocalPDF` requests:
  - receives `window.postMessage` from viewer
  - calls `ipc.invoke('readLocalPDF')`
  - returns result/error back to viewer with `window.postMessage`

- `pages/pdfViewer/viewer.js`
  Updated loading flow:
  - `file://` PDFs: load via byte data (`getDocument({ data })`)
  - remote PDFs: keep existing URL flow (`getDocument({ url, withCredentials: true })`)

- `pages/pdfViewer/documentRequest.mjs`
  Added helper for selecting local-vs-remote PDF.js request mode.

### Security
- only internal PDF viewer can use local file read path
- only for the exact local file currently opened in that viewer


pages/pdfViewer/viewer.js
```diff
@@ -239,7 +240,61 @@ function setUpPageAnnotationLayer (pageView) {
   }
 }
 
-pdfjsLib.getDocument({ url: url, withCredentials: true }).promise.then(async function (_pdf) {
+function requestLocalPDFData (fileURL) {
+  return new Promise(function (resolve, reject) {
+    var requestId = ++localPDFRequestId
```
PalmerAL, line 245: `localPDFRequestId` is unique to each viewer instance, so it will be 0 in basically all cases. To create a unique request ID, you probably need to use `Math.random()` or similar.

pages/pdfViewer/viewer.js
```diff
@@ -239,7 +240,61 @@ function setUpPageAnnotationLayer (pageView) {
...
+
+      resolve(new Uint8Array(e.data.data))
+    }
+
+    window.addEventListener('message', onResponse)
+    window.postMessage({ message: 'readLocalPDF', requestId, url: fileURL }, 'min://app')
+  })
+}
+
+function getPDFDocumentRequest (targetURL, requestLocalPDFData) {
```
PalmerAL, line 276: I think this can just call the global requestLocalPDFData; it doesn't need to be an argument.

Issue #2602, opened 2025-04-24
localhost:5000


Dart-o-s:
Sends a search request to DuckDuckGo.

Certainly not what one would expect. Bu I know, Microsoft Old-Edge or however the previous browser was called, has the same misbehaviour.

Lets see if it at least recognizes an IP address: 127.0.0.1:5000

Yay! It does :D

PalmerAL:
This works for me. We're not related to Microsoft, can you confirm that you've opened this issue in the correct project?

However, it's possible that if you've edited your `/etc/hosts` file (C:\Windows\System32\drivers\etc\hosts  on Windows) to not include localhost, it won't work. Could you comment the contents of that file here, redacting any personal information?

Dart-o-s:
Well, the browser is not opening localhost.

It is searching for the term "localhost" on the default configured search
engine.

Has nothing to do with a host file.

But thanks for letting me check my hosts file. Strangely the docker install
changed my 127.0.0.1 to kubernetes.docker.internal.

Hm, I am in the process of migrating to Linux anyway.

Have a nice Weekend!

Angelo

OmegaGiven:
made a way to fix this:

https://github.com/minbrowser/min/pull/2700

Issue #2516, opened 2024-10-28
Not able to set as the default browser


merrilsamshiju:
The app doesn't seem to have been installed as an .exe file. On the Windows 10, when "min" is searched, only the min-setup.exe file is shown. So a separate app has not been installed (atleast in the sense of a normal user, it doesn't show up on the app list). Therefore, I am not able to set Min as the default browser
![Screenshot (4)](https://github.com/user-attachments/assets/05731576-8197-460f-91f3-3351c59600cc)

Conradus2:
In the default applications search for the browser.which is set as the default and click one by one on the list changing to Min,This is the only way in this failed version!
Sorry for the poor English!!!

OmegaGiven:
see this PR: https://github.com/minbrowser/min/pull/2700

good alternative if merged

Issue #482, opened 2018-01-20
localhost:3000 in the location bar calls the search engine, not the address


code-hunger:
It's very annoying, I need to write a valid file name to get it opened (e.g. localhost:3000/index.html)

PalmerAL:
Until this is fixed, you can work around it by adding localhost to your hosts file (in C:/Windows/System32/drivers/etc/hosts).

dbfreem:
This seems to only be a windows issue. I used localhost:3000 on xubuntu today and it worked. I went and tried it on windows and it went to DuckDuckGo.

code-hunger:
Haven't checked it on windows.  
What do you mean by "it went to DuckDuckGo"? It must open the address, not the search engine.

OmegaGiven:
2026 i have this same issue on arch linux... i made a PR to fix this giving you an option that will try as if its a domain basically appending the https to the beginning.

PR #2700, opened 2026-04-09
Added visit Site option to access custom domains (including localhost)

OmegaGiven:
This is a simple addition to always allow a bottom action to just try and search the url entered as if it was an actual domain. this is especially useful for homelab settups and custom domains in a network for big homelabers like me

main use case is go/ is a custom domain that redirects me to different websites on any device in my tailnet and in the vanilla min browser it just treats it as a web search.

its the bottom recommended option will now always be there for those of us that need it for first time min users so that we dont have to type https://go/ manually and such

this would also give those that have issues with localhost an option to use this to override any issues with the basic interpertation and not result in a search instead
image

Issues this would cover with this as an option and is how brave browser handles it in a round about way:

https://github.com/minbrowser/min/issues/482

https://github.com/minbrowser/min/issues/2602

https://github.com/minbrowser/min/issues/2516

### DIFF
--- a/js/default.js
+++ b/js/default.js
@@ -171,6 +171,7 @@ require('macHandoff.js').initialize()
 // default searchbar plugins
 
 require('searchbar/placesPlugin.js').initialize()
+require('searchbar/domainGuessPlugin.js').initialize()
 require('searchbar/instantAnswerPlugin.js').initialize()
 require('searchbar/bangsPlugin.js').initialize()
 require('searchbar/customBangs.js').initialize()
--- a/js/searchbar/domainGuessPlugin.js
+++ b/js/searchbar/domainGuessPlugin.js
@@ -9,11 +9,19 @@ function showDomainGuess (text) {
     return
   }
 
+  var url
+
+  if (urlParser.isHTTPSUpgreadable(text)) {
+    url = 'https://' + text
+  } else {
+    url = 'http://' + text
+  }
+
   searchbarPlugins.addResult('domainGuess', {
     icon: 'carbon:earth-filled',
     title: text,
     secondaryText: l('visitSite'),
-    url: urlParser.parseAsURL(text)
+    url: url
   }, { allowDuplicates: true })
 }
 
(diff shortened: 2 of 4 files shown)


PalmerAL:
This does seem like something that's worth fixing, but I wonder if this is the right solution. If we did this, every time you visited the go/ URL, you would have to tab down to the bottom to select this entry, right?

It looks like once you manually type "http://go" once, autocomplete is already able to autocomplete the "go" part if you start typing "g". The problem is that when you press enter, the original autocomplete entry is ignored, and the searchbar interprets "go" as a search instead. I'm thinking that we could update that so that we track the original URL that generated the autocompletion, and when enter is pressed, load that instead of trying to parse the text in the searchbar.

The effect would be that you would have to manually type "http://go" on the first visit, but on subsequent visits, just typing "go" would work and would be converted to "http://go" by autocomplete.

If we did that, do you think that would sufficiently address the issue, or would it still be better to have something like this to handle the case where the name has never been typed before?

OmegaGiven:
I see what you mean.
Here is what brave browser does:
image

It puts it as the first option but i think it tries if its a passing dns and if fails then it will fall back to search engine. Firefox has a hard time with this but chrome is ok but I find myself having to scroll down in chrome to the dns search. So just depends on where you want it. Look at my changes or if you want to test it and let me know what you think and if we want it to be the first or last option, I just know this is a showstopper for homelab people like me to daily drive this. Also know I dont understand the min bowser deeply so maybe changes for this should live somewhere else. I just know for people that do this they want to type a custom dns and it work first time else its a pain.

PalmerAL:
Electron has a method to look up whether a DNS entry exists: https://www.electronjs.org/docs/latest/api/session#sesresolvehosthost-options. Do you want to try using that?

Perhaps the plugin could work like this:
1. We add 50-100ms of debouncing to wait for the user to stop typing
2. If the searchbar text is a single word, we use resolveHost to check if it's a valid DNS entry.
3. If it is, we show the suggestion first in the list.
4. If it's not, we show nothing.

PalmerAL:
Also, I've published a branch with the autocomplete change I mentioned previously: https://github.com/minbrowser/min/tree/autocomplete-for-dns-names. Can you try it out and let me know what you think?

PR #2702, opened 2026-04-15
Add missing translations in zh-CN.json

FanXiangTM:
Supplementary translation

### DIFF
--- a/localization/languages/zh-CN.json
+++ b/localization/languages/zh-CN.json
@@ -23,15 +23,15 @@
     "cut": "剪切",
     "copy": "复制", //this is a verb (as in "copy the currently-selected text")
     "paste": "粘贴",
-    "pasteAndMatchStyle": null, //missing translation
+    "pasteAndMatchStyle": "粘贴并匹配样式",
     "goBack": "返回",
     "goForward": "前进",
     "inspectElement": "查看元素",
-    "translatePage": null, // missing translation
+    "translatePage": "翻译页面",
     /* searchbar */
-    "placesPluginOpenAgain": null, // missing translation
-    "placesPluginSwitchToTab": null, // missing translation
-    "placesPluginSwitchToTask": null, // missing translation
+    "placesPluginOpenAgain": "再次打开插件",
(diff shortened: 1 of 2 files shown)


localization/languages/zh-CN.json
```diff
@@ -23,15 +23,15 @@
...
     "goBack": "返回",
     "goForward": "前进",
     "inspectElement": "查看元素",
-    "translatePage": null, // missing translation
+    "translatePage": "翻译页面",
     /* searchbar */
-    "placesPluginOpenAgain": null, // missing translation
-    "placesPluginSwitchToTab": null, // missing translation
-    "placesPluginSwitchToTask": null, // missing translation
+    "placesPluginOpenAgain": "再次打开插件",
```
PalmerAL: Can you double-check that these are correct? It looks like they contain a translation for "plugin", which is incorrect. They should be translations of the [English source](https://github.com/minbrowser/min/blob/master/localization/languages/en-US.json#L32-L34):

```
    "placesPluginOpenAgain": "Open Again",
    "placesPluginSwitchToTab": "Switch to Tab",
    "placesPluginSwitchToTask": "Switch to \"%t\"", // %t is replaced with the name of a task
```

Please review this for all of the other translations also.

PalmerAL (changes requested):

localization/languages/zh-CN.json
```diff
@@ -210,7 +210,7 @@
     "appMenuCut": "剪切",
     "appMenuCopy": "复制",
     "appMenuPaste": "粘贴",
-    "appMenuPasteAndMatchStyle": null, //missing translation
+    "appMenuPasteAndMatchStyle": "应用程序菜单粘贴和匹配样式",
```
PalmerAL: Can you double-check this? It should not contain a translation for "application menu":
```
    "appMenuPasteAndMatchStyle": "Paste and Match Style",
```

PalmerAL (changes requested):

PalmerAL:
Thanks!

PR #2715, opened 2026-06-10
Add Focus Mode feature

smhvigorously:
## Summary
- Implements a Focus Mode for distraction-blocking
- Adds dedicated CSS (`css/focusMode.css`) and expanded JS logic (`js/focusMode.js`)
- Adds UI elements to the main window and settings page

## Test plan
- [ ] Toggle Focus Mode on/off from settings
- [ ] Verify blocked sites are inaccessible while focus mode is active
- [ ] Verify UI indicators appear correctly in the toolbar

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/css/focusMode.css
+++ b/css/focusMode.css
@@ -1,0 +1,266 @@
+/* ─── Toolbar indicator pill ─────────────────────────────────────────────── */
+
+#focus-mode-indicator {
+  position: absolute;
+  /* centered horizontally over the tab bar area */
+  left: 50%;
+  top: 50%;
+  transform: translate(-50%, -50%);
+  z-index: 10;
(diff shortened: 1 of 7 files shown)


PalmerAL:
This seems like a neat idea, but I think we'd need to come up with a more sophisticated/accurate way of tracking whether you're on goal than the relevance scoring function in this PR. I also expect we'd want to integrate this with the existing feature named "focus mode".

I cant find it right now, but a while ago I had a branch where I tested taking the text content of the page, mapping each word to a vector in a set of word embeddings, and summing them. It was pretty accurate at identifying if two pages were related; perhaps you could do something similar here.

